PR #14, opened 2020-09-23
Dashboard

akorchev:


mikecasas:
Great job on the dashboard! Will this be a component too? Meaning we can add the component and enter the Github repo name?

akorchev:
We don't plan to make a component out of it but you can easily use your own repo by editing the [GitHubService](https://github.com/akorchev/blazor.radzen.com/blob/master/Services/GitHubService.cs#L130).

PR #19, opened 2020-10-31
Update CodeViewer.razor

robertmclaws:
The height specified presently makes the code window unnecessarily small, with the result being the code samples are too difficult to read on larger screens. This change guarantees the layout you want while making it easier for people with 4k monitors to read more of the code.


akorchev:
Thanks!

PR #43, opened 2021-02-11
enable customization of RadzenFieldset display style

RyanMarcotte:
Enables customization of `RadzenFieldset` display style by defining a `DisplayStyle` enum and a `DisplayStyle` property for `RadzenFieldset`.  The default value `DisplayStyle.Block` preserves existing behavior.

Using `DisplayStyle.Block`:
image

Using `DisplayStyle.Contents`:
image

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -220,6 +220,12 @@ namespace Radzen
         Info
     }
 
+    public enum DisplayStyle
+    {
+        Block,
+        Contents
+    }
+
     public enum FilterMode
     {
         Simple,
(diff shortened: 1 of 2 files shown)


akorchev:
In general we don't like the idea of having properties that do nothing but apply certain CSS rule(s). The reason is there would be no end of them e.g. 'FontSize', 'Padding' etc.  This is why all Radzen Blazor components support custom css class applied via the `class` attribute. This would allow you to achieve your goal which is to have `display: contents`.

The only thing we need to do is implement the hiding in such a way so it doesn't use `display: block`. Something like 

```
contentStyle = collapsed ? "display: none;" : "";
```

Then one could use a CSS class to specify `display: content`

```
.content-display .rz-fieldset-content-wrapper {
   display: content;
}
```

```
<RadzenFieldset class="content-display" />
```

Would you be willing to update your pull request with the aforementioned change only?

RyanMarcotte:
That works, thanks for the suggestion!

PR #52, opened 2021-02-17
Add refresh to dialog service

vflame:
also, changed demo project dependency off nuget package


enchev:
Hey @vflame,

I'll merge your pull request however the reference to the NuGet should be restored. At the moment our demos are deployed to Docker and project reference will not work - we will check how we can enable it in the future.

enchev:
Implements #52

PR #90, opened 2021-03-29
Support Formatting in RadzenNumeric

### LINKED ISSUE #89: Format displayed number in RadzenNumeric
Numbers get quite confusing, especially if they are big and need to be handled by users (in a GUI :)).
Therefore most frontend frameworks support a formatting of numbers with decimal seperators, thousand separators etc.
C# offers the ToString Method on numeric values with a formatter.

Add a Format Property to the RadzenNumeric that (optionally) allows to format the Display according to the current UI Culture.

* Creating an own component --> the RadzenNumeric Component is the perfect place to use
* Hack it into RadzenInput --> Multiple String to Value Conversions are necessary to handle that

MPapst:
fixes #89 

This pullrequest adds the optional capability to format the value of a RadzenNumeric.

The current handling always converts the internal value to a decimal in order to handle it. It may be a good Idea to do this only when it is set from external (but maybe this is by design so I did not change it)

### DIFF
--- a/Radzen.Blazor.Tests/NumericTests.cs
+++ b/Radzen.Blazor.Tests/NumericTests.cs
@@ -331,5 +331,23 @@ namespace Radzen.Blazor.Tests
             Assert.DoesNotContain(@$"rz-spinner-up", component.Markup);
             Assert.DoesNotContain(@$"rz-spinner-down", component.Markup);
         }
+
+        [Fact]
+        public void Numeric_Formatted()
+        {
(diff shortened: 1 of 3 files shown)


Codingjames:
How can i use this commit for modify 1.0000 to 1.00

MPapst:
It supports all the format specifiers for the .ToString() method:
https://docs.microsoft.com/de-de/dotnet/api/system.double.tostring?view=net-5.0

PR #94, opened 2021-04-07
fix fail to input password properly while typing fast

kinit-ltd:
@bind="@Password" already bind the property "Password"
@oninput="@(args => Password = $"{args.Value}")" is duplicate and makes conflict with @bind


enchev:
Without `@oninput` code one will unable to login with Enter key press after typing password

kinit-ltd:
However, the @oninput makes the textbox even can't input a correct password into it. Please refer to the attached video. [https://drive.google.com/file/d/1pFonHmlKwtTzqfMEz72sD0ycyafT4MZW/view](url)
The Enter Key Press feature may better implement in client side code, i.e. javascript.

PR #107, opened 2021-04-28
Method is unused

chrgraefe:
This method is unused and can removed


chrgraefe:
Moved "Busy" method into DialogService and cleand up the example razor page

PR #108, opened 2021-04-28
Moved "Busy" method into DialogService and cleand up the example razor page

chrgraefe:
Moved "Busy" method into DialogService and cleand up the example razor page


enchev:
Hey @chrgraefe ,

We intentionally left both methods as examples:
- how to use with a markup directly
- how to create markup programmatically

PR #121, opened 2021-05-23
feature/busy indicator button

### LINKED ISSUE #120: RadzenButton to expose IsBusy Property
Most Buttons in Web Application trigger actions (duh!) 
As those actions are executed async, there is need to inform the user that there is progress, but not visible to him. There are several solution to create these progress informations, one of the is to use a "Busy Indicator" on the button where the user clicked.
Currently this is not that easy to achieve with Radzen, as there are multiple properties to adjust and animation of an icon is not possible throught the RadzenButton API.

Please see PR.

See above in Description.

**Additional context**
Various other Button implementations (especially Angular, where I worked with) offer similar APIs.

**Notes**
A.K.A.: There is stuff to discuss on my PR:
* I am unable to run the tests, as the compilation only works in the CLI not in VS 2019; and the CLI does not find the Tests
*  I added inline style items --> I am not 100% sure if this matches the correct style
* Maybe there is a better icon for the spinner (or even plain css)
* The solution is overridable by using ChildContent
* Not setting the BusyText Property removes eventual text and the button will resize!

For the Documentation:
* When setting IsBusy = true, the Disabled state is stored and restored when setting IsBusy to false
* Does not work in combination with ChildContent

MPapst:
fixes #120

### DIFF
--- a/Radzen.Blazor/RadzenButton.razor
+++ b/Radzen.Blazor/RadzenButton.razor
@@ -2,7 +2,7 @@
 
 @if (Visible)
 {
-    <button @ref="@Element" style="@Style" disabled="@Disabled"
+    <button @ref="@Element" style="@Style" disabled="@IsDisabled"
             type="@Enum.GetName(typeof(ButtonType), ButtonType).ToLower()"
             @attributes="Attributes" class="@GetCssClass()" id="@GetId()"
             @onclick="@((args) => OnClick(args))">
@@ -39,9 +39,6 @@
     </button>
 }
 @code {
-    private bool _isBusy;
-    private bool _originalDisabled;
-
     private string getButtonSize()
     {
         return Size == ButtonSize.Medium ? "md" : "sm";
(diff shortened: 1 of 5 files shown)


Radzen.Blazor/RadzenButton.razor
```diff
@@ -60,6 +74,27 @@
...
+            {
+                // store the original Value of disabled property
+                // when entering busy state
+                _originalDisabled = Disabled;
+            }
+            _isBusy = value;
+            // return to original disabled state, when leaving busy state
+            Disabled = value ? value : _originalDisabled;
+        }
+    }
```
akorchev, line 93: I think this can be simplified by using a private property which returns the actual disabled state:

```
public bool IsDisabled
{
    get
    {
         return Disabled || IsBusy;
    }
}
```

And then use it instead of Disabled e.g. `<button disabled=@IsDisabled />`

If this is done IsBusy can be an auto property (eliminate _isBusy) and there would be no need to (re)store Disabled (thus eliminate _originalDisabled)

MPapst: Done

PR #150, opened 2021-06-19
fix: RadzenNumeric unable to convert currency formatted values

### LINKED ISSUE #147: RadzenNumeric isn't updating value when format property is used
**Describe the bug**
RadzenNumeric isn't updating value when format property is used

**To Reproduce**
Repo: https://github.com/chrgraefe/RadzenNumeric

1. Start project
2. Navigate to "/radzennumericexample"
3. Click into numeric field
4. delete just the value (50) after the decimal point
5. type 25 as new value after the decimal point. So the new value is 1200.25 €
6. leave the form field
7. value is now 0.00 €

**Expected behavior**
Value is in the field as editted

MPapst:
fixes #147

### DIFF
--- a/Radzen.Blazor/RadzenNumeric.razor
+++ b/Radzen.Blazor/RadzenNumeric.razor
@@ -152,10 +152,20 @@
         await InternalValueChanged(args.Value);
     }
 
+    private string RemoveNonNumericCharacters(object value)
+    {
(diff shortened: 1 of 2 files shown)


MPapst:
I tried to add a unit test for regression, but I am failing to trigger a change in bUnit of the input's value

PR #171, opened 2021-07-13
Feature/badge component

### LINKED ISSUE #66: New "Badge" component
Could you add a Badge component?'

Small numerical value or status descriptor for UI elements.

like below shown to us.

Badge component

MPapst:
fixes #66 

Introduces a RadzenBadge component using the Bootstrap Badge classes.
Using a more generic case than only displaying numbers as shown in #66.


akorchev:
Hi @MPapst  and thank you for the pull request!

We would probably not rely on the Bootstrap built-in badge classes though. We have a -base version of the themes which does not include any Bootstrap CSS. Those would render the badge component style-less.

You can check what we currently do for the button styling - it still has bootstrap like styles (primary, secondary etc) but they do not depend on Bootstrap. Would you consider adding a new _badge.scss that uses custom styles? You can check [_button.scss](https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/themes/components/blazor/_button.scss) for inspiration.

MPapst:
Sure, will have a look and adapt to match the overall style . But that may be tomorrow

akorchev:
Forgot to mention - we also use a common prefix for CSS classes "rz-" to avoid conflicts with other style systems.

Thanks again for the contribution!

MPapst:
Tried it - currently stuck with: scss is not transpiled. Will have a look hopefully tomorrow on that.

MPapst:
found it :)

@akorchev could you have a look on the scss? I'm a first time user and especially the part when for removing the `text-decoration` on a `a.rz-badge:hover` may be better to solve?

There is still one thing bothering me:
When hovering a badge with Path, the font color changes (due to the a-tag). I think it may be more suitable to have it behave like a button.
Will have a look tomorrow on that.

akorchev:
Do we need the badge to behave like a link too? I know Bootstrap's supports it but I am not sure what the use case it. Right now it just adds complexity (the rendering needs to be duplicated and needs more CSS to counter the built-in `<a>` styling).

MPapst:
That depends on you :)
I am in need of the default badge - and I think if there is need to provide a link, I would go using a Tooltip to provide more information to the user what happens if he clicks on it. I just added that one because of the support in bootstrap.

akorchev:
In that case let's scrap the link behavior - would reduce the implementation quite a bit and remove complexity. One can still achieve something similar by adding an `@onclick` attribute and handling it from there.

MPapst:
removed the link

PR #185, opened 2021-07-27
Support custom filtering

viordash:
This PR adds following:
- ability to filter data with custom rules by assigning the RadzenDataGridColumn.CustomFiltering function in Client-side mode.
- disabling unnecessary filtering/sorting operations in RadzenDataGrid.View in Server-side mode.
- added related tests, 
   - DataGrid_View_Use_ClientSide_Filtering_When_LoadData_Empty
   - DataGrid_View_Use_ServerSide_Filtering_When_LoadData_Used
   - DataGrid_View_Use_Custom_Filtering 

 RadzenDataGridColumn custom filtering:
```cs
    [Parameter]
    public Func<CustomFilteringArgs<TItem>, bool> CustomFiltering { get; set; } = null;
```
```cs
    public class CustomFilteringArgs<T>
    {
        public Blazor.RadzenDataGridColumn<T> Column { get; internal set; }
        public T Data { get; internal set; }
    }
```

Trigger to this PR was the System.Linq.Dynamic.Core.Exceptions.ParseException that was thrown when trying to use RadzenDataGrid.View in event LoadData and active filter FilterOperator.Contains in a column with composite data, for example, an array of objects with fields [int Id], [string Name]

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -814,38 +814,42 @@
     {
         get
         {
-            var view = base.View.Where<TItem>(columns);
-            var orderBy = GetOrderBy();
+            bool serverSideDataOperations = LoadData.HasDelegate;
+            var view = base.View;
+            if (!serverSideDataOperations)
+            {            
+                var orderBy = GetOrderBy();
 
-            if (!string.IsNullOrEmpty(orderBy))
-            {
-                if (typeof(TItem) == typeof(object))
+                if (!string.IsNullOrEmpty(orderBy))
                 {
-                    var firstItem = view.FirstOrDefault();
-                    if (firstItem != null)
+                    if (typeof(TItem) == typeof(object))
                     {
-                        view = view.Cast(firstItem.GetType()).AsQueryable().OrderBy(orderBy).Cast<TItem>();
+                        var firstItem = view.FirstOrDefault();
+                        if (firstItem != null)
+                        {
+                            view = view.Cast(firstItem.GetType()).AsQueryable().OrderBy(orderBy).Cast<TItem>();
+                        }
+                    }
+                    else
+                    {
+                        view = view.OrderBy(orderBy);
                     }
                 }
-                else
-                {
-                    view = view.OrderBy(orderBy);
-                }
-            }
 
-            if (!LoadData.HasDelegate && (!IsVirtualizationAllowed() || AllowPaging))
-            {
-                var count = view.Count();
-                if (count != Count)
+                if (!IsVirtualizationAllowed() || AllowPaging)
                 {
-                    Count = count;
-
-                    if (skip >= Count && Count > PageSize)
+                    var count = view.Count();
+                    if (count != Count)
                     {
-                        skip = Count - PageSize;
-                    }
+                        Count = count;
 
-                    StateHasChanged();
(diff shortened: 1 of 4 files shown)


viordash:
Apologize, I posted this PR too early
In real use of CustomFiltering, two issues appeared, an InvalidOperationException thrown if data from EF, and the inaccessibility of filtering parameters, internal access modifier for some RadzenDataGridColumn fields.
I added the necessary changes ([85b75ed](https://github.com/radzenhq/radzen-blazor/pull/185/commits/85b75ede5536a606c54efc1df847f94c1fdb12a6), [963f7bb](https://github.com/radzenhq/radzen-blazor/pull/185/commits/963f7bbb1e285b9889b005b164f4c86380e54b45)) to resolve these issues, and now the CustomFilteringArgs is following:
```cs
    public class CustomFilteringArgs<T>
    {
        public T Data { get; internal set; }
        public string Property { get; set; }
        public FilterOperator FilterOperator { get; internal set; }
        public object FilterValue { get; internal set; }
        public FilterOperator SecondFilterOperator { get; internal set; }
        public object SecondFilterValue { get; internal set; }
        public LogicalFilterOperator LogicalFilterOperator { get; internal set; }
    }
```

Radzen.Blazor/RadzenDataGrid.razor
```diff
@@ -829,38 +829,51 @@
...
+            var view = base.View;
+            if (!serverSideDataOperations)
+            {
+                var columnsWithNoCustomFiltering = columns.Where(c => c.CustomFiltering == null);
+                view = view.Where<TItem>(columnsWithNoCustomFiltering);
+
+                var columnsWithCustomFiltering = columns.Where(c => c.CustomFiltering != null);
+                if (columnsWithCustomFiltering.Any())
+                {
+                    view = view.AsEnumerable<TItem>().Where<TItem>(x => ApplyDataCustomFiltering(x, columnsWithCustomFiltering)).AsQueryable();
```
enchev, line 842: AsEnumerable<TItem>() will get all data and will brake completely IQueryable support.

viordash: I assume that the Custom filter can be executed with only in-memory data, probably no ORM exist that can execute the requirements that may be required by the user in the code of CustomFilter.

A custom filter can be placed after OrderBy so that OrderBy is also quered on the database side. But it seems to me that in this case it will not be "real" filtering, since there is a possibility that the user in the custom code will add a limit on the amount of output data as one of the filtering conditions. And if the CustomFilter is located after the OrderBy, then the output dataset will not be equal to the expected one.

enchev: Hey @viordash,

I'm afraid that we cannot merge such custom filtering. Both in-memory (EnumerableQuery or so called LINQ to Objects) and database filtering can be done using the same approach - LINQ expressions, however this is not the case with this pull request. More info about Expression class can be found here:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression?view=net-5.0

I see what you are after, FilterExpression property for the DataGridColumn that can work both for in-memory and database filtering and we will do our best to provide such in some of our next versions.

viordash:
Ok, 
Can you fix the System.Linq.Dynamic.Core.Exceptions.ParseException error described at the end of the PR description when using RadzenDataGrid.View? Or make a separate PR for this?

PR #201, opened 2021-08-19
AllowInput Implementation

douglassimaodev:
Implementation of AllowInput for calendar


douglassimaodev:
Thanks,  will that be generating a new update for the  nuget package?

akorchev:
No. We will release a new version by the end of the week.

Issue #206, opened 2021-08-31
Add a `Culture` parameter on components for the default templates


javiercampos:
**Is your feature request related to a problem? Please describe.**
When changing the Cutlure dynamically on a page (without reloading, thus, without changing the renderer thread's `CurrentCulture`/`CurrentUICulture`), you cannot reliably change the renderer's synchronization context Culture, so async calls (or `InvokeAsync()`) may have a different Culture.

I've made a feature request for Blazor to include this capability here: https://github.com/dotnet/aspnetcore/issues/35965

When you show the value or make binding yourself in the template you can use `.ToString(MyCulture)` or `@bind-Value:culture="@MyCulture"` if needed, but components like the `RadzenDataGrid`, or `RadzenDropdown`, etc., where you don't bind directly (instead, you specify a type and a property name), always use the thread's culture on their default templates, so if you want a different culture, you need to specify the whole `Template` for every column, or dropdown template, etc., which is really inconvenient.

**Describe the solution you'd like**
Have a parameter on those components of type `CultureInfo` to specify the default culture on the component, e.g.: `<RadzenDataGrid Culture="@MyCulture">`, so that it takes it into account in the default templates (and it does `.ToString(Culture)` when rendering, instead of just `ToString()` -or nothing, which calls `ToString` with the `CurrentUICulture` format provider).

It could use `CultureInfo.CurrentUICulture` if not specified, so it would not break any existing code.

Additionally, default in-component templates for reflection-based rendered components (where you specify the type and the property name, instead of binding directly) would need to be revised to make use of that parameter.

akorchev:
I think we won't implement this as there are [supported ways to change the current culture](https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0&pivots=server#dynamically-set-the-culture-from-the-accept-language-header-1) in an ASP.NET Core application which work. Yes they require a reload but this is not such a big problem.

The Startup.cs file of the Sample application contains commented code which can be used to set the current application culture.

I have linked a screenshot showing this in action in this comment: https://github.com/radzenhq/radzen-blazor/pull/229#issuecomment-927694226

akorchev:
This turned out to be a problem when certain conditions are met (static classes and events - more [here](https://github.com/radzenhq/radzen-blazor/pull/229#issuecomment-927826151). We are reopening the issue and will accept pull requests.

javiercampos:
[...]

Issue #168, opened 2021-07-08
Feature Request - Scheduler Improvements


recomrad:
- [x] First minor - make hours headers customizable instead of strict "h tt" format.
- [ ] Second major - resize and drag & drop appointments ability.

Dazfl:
I would love to see the 'second major' implemented as well.

JosephMampaey:
@recomrad Are you still working on this? Can we help with anything?

KevinKrueger:
Resize for duration is important.

KevinKrueger:
So if the resize(start, end) with visual cursor feedback and the drag-drop functionality works now, then I'm the happiest developer in the whole world. Probably XD

akorchev:
Closing this one in favor of #1833. Drag and drop to move is already supported.

PR #229, opened 2021-09-27
RadzenScheduler Culture override support

ThnxDaniel:
Enables complete localization of the scheduler component.

Related issues:
- #168
- #206

### DIFF
--- a/Radzen.Blazor/RadzenMonthView.razor
+++ b/Radzen.Blazor/RadzenMonthView.razor
@@ -6,12 +6,7 @@
 @code {    
     public override string Title
     {
-        get
-        {
-            var sb = new System.Text.StringBuilder(Scheduler.CurrentDate.ToString("MMMM yyyy", Scheduler.Culture));
-            sb[0] = char.ToUpper(sb[0], Scheduler.Culture);
-            return sb.ToString();
-        }
+        get => Scheduler.CurrentDate.ToString("MMMM yyyy", Scheduler.Culture);
     }
 
     [Parameter]
--- a/Radzen.Blazor/RadzenMonthView.razor
+++ b/Radzen.Blazor/RadzenMonthView.razor
@@ -8,7 +8,9 @@
     {
         get
         {
-            return Scheduler.CurrentDate.ToString("MMMM yyyy");
+            var sb = new System.Text.StringBuilder(Scheduler.CurrentDate.ToString("MMMM yyyy", Scheduler.Culture));
+            sb[0] = char.ToUpper(sb[0], Scheduler.Culture);
+            return sb.ToString();
         }
     }
 
@@ -45,12 +47,12 @@
 
         if (MaxAppointmentsInSlot != null)
         {
-          maxAppointmentsInSlot = MaxAppointmentsInSlot.Value;
+            maxAppointmentsInSlot = MaxAppointmentsInSlot.Value;
         }
         else
         {
-          var slotHeight = (Scheduler.Height - 60) / 6;
-          maxAppointmentsInSlot = Convert.ToInt32(Math.Floor(slotHeight / 24)) - 1;
+            var slotHeight = (Scheduler.Height - 60) / 6;
+            maxAppointmentsInSlot = Convert.ToInt32(Math.Floor(slotHeight / 24)) - 1;
         }
 
         return @<CascadingValue Value=@Scheduler>
--- a/Radzen.Blazor/IScheduler.cs
+++ b/Radzen.Blazor/IScheduler.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
 
@@ -19,5 +20,7 @@ namespace Radzen.Blazor
         RenderFragment RenderAppointment(AppointmentData item);
         Task Reload();
         double Height { get; }
+
+        CultureInfo Culture { get; set; }
     }
 }
--- a/Radzen.Blazor/RadzenDayView.razor
+++ b/Radzen.Blazor/RadzenDayView.razor
@@ -8,6 +8,9 @@
     public override string Text { get; set; } = "Day";
 
     [Parameter]
+    public string TimeFormat { get; set; } = "h tt";
+
+    [Parameter]
     public TimeSpan StartTime { get; set; } = TimeSpan.FromHours(8);
 
     [Parameter]
@@ -17,7 +20,7 @@
     {
         get
         {
-            return Scheduler.CurrentDate.ToShortDateString();
+            return Scheduler.CurrentDate.ToString(Scheduler.Culture.DateTimeFormat.ShortDatePattern);
         }
     }
 
@@ -44,7 +47,7 @@
         var appointments = Scheduler.GetAppointmentsInRange(date, date.AddDays(1)).ToList();
 
         return @<CascadingValue Value=@Scheduler>
-                    <DayView StartDate=@StartDate EndDate=@EndDate StartTime=@StartTime EndTime=@EndTime Appointments=@appointments />
+                    <DayView StartDate=@StartDate EndDate=@EndDate StartTime=@StartTime EndTime=@EndTime Appointments=@appointments TimeFormat="@TimeFormat" />
                 </CascadingValue>;
     }
 
--- a/Radzen.Blazor/RadzenScheduler.razor
+++ b/Radzen.Blazor/RadzenScheduler.razor
@@ -1,4 +1,5 @@
 @using Radzen
+@using System.Globalization;
 @using System.Linq.Dynamic.Core;
 @using Microsoft.JSInterop
 @inherits RadzenComponent
@@ -71,6 +72,9 @@
 
     [Parameter]
     public EventCallback<SchedulerLoadDataEventArgs> LoadData { get; set; }
+
+    [Parameter]
+    public CultureInfo Culture { get; set; } = CultureInfo.CurrentCulture;
 
     IList<ISchedulerView> Views { get; set; } = new List<ISchedulerView>();
 
--- a/Radzen.Blazor/RadzenWeekView.razor
+++ b/Radzen.Blazor/RadzenWeekView.razor
@@ -8,21 +8,24 @@
     public override string Text { get; set; } = "Week";
 
     [Parameter]
+    public string TimeFormat { get; set; } = "h tt";
+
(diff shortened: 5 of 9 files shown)


Radzen.Blazor/RadzenMonthView.razor
```diff
@@ -8,7 +8,9 @@
     {
         get
         {
-            return Scheduler.CurrentDate.ToString("MMMM yyyy");
+            var sb = new System.Text.StringBuilder(Scheduler.CurrentDate.ToString("MMMM yyyy", Scheduler.Culture));
+            sb[0] = char.ToUpper(sb[0], Scheduler.Culture);
```
akorchev, line 12: Why is this needed?

akorchev:
Could you elaborate on the need to have a different Culture per component? As far as I know all string methods use the CurrentUICulture anyway.

Radzen.Blazor/RadzenMonthView.razor
ThnxDaniel: Some cultures don't treat months as names, resulting in incorrect capitalization for this usecase, sv-SE is one of them.

akorchev: Shouldn't we leave it as it is per the Culture rules then? If a culture does not capitalize its month names we shouldn't do that as well.

ThnxDaniel: Sure, but in this case its also the start of a heading/sentence, which is why should be capitalized even though the months aren't treated as names.

ThnxDaniel:
> Could you elaborate on the need to have a different Culture per component? As far as I know all string methods use the CurrentUICulture anyway.

In my case I need to override the culture, which is not possible by simply changing the value of CurrentUICulture

akorchev:
Have you tried the way described in the[Blazor documentation](https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0&pivots=server#dynamically-set-the-culture-from-the-accept-language-header-1)?

By uncommenting the [localization](https://github.com/radzenhq/radzen-blazor/blob/master/RadzenBlazorDemos/Startup.cs#L55-L67) [code](https://github.com/radzenhq/radzen-blazor/blob/master/RadzenBlazorDemos/Startup.cs#L87-L99) in the Radzen sample application and changing "de-DE" to "bg-BG' I was able to localize the Scheduler in Bulgarian without any other changes:

image

We believe this provides a robust way to implement localization at application level. There isn't much need to add a Culture property at component level.

javiercampos:
@akorchev I'm commenting here because you closed my thread, but that way of doing localization is problematic (and it's indeed found problematic by the Blazor team, they are targeting NET 7.0 to address it). Blazor server has a shared server session context, so actions from one user (in one culture) may affect data that updates the UI on other use (which may use a different culture according to the localization set on the http connection, and not on the signalr circuit).

So even if you can reload and show the main rendering in the culture set by the middleware (which again, is not invoked on the SignalR circuit), say one user updates the actual data on the scheduler (to cite the example you put here), so it forces a rendering on all users showing that on screen... no matter the actual middleware-based culture for those other users, anything rendered/updated will have the "de-DE" culture (and will show with the german culture to bulgarian users): this includes date formatting or localized strings (if you use the `IStringLocalizer`).

Using the ASP.NET localization middleware for setting the Blazor Server culture is just a very bad idea for many reasons (note that also `CurrentUICulture` is not kept between async calls on the blazor synchronization context either), and until either Blazor has a more suitable method of setting the culture for either the components or the actual SignalR circuit, a workaround is needed.

I have that workaround for my own components, but I just cannot do it for third-party components in a generalized way (without having to derive from -all- components), whereas it would be "relatively easy" to do for you on the base components.

akorchev:
I am not sure why you think this way of doing localization is problematic as it the the official one mentioned in the Blazor documentation. We have been using it in all Radzen Blazor applications since 2019 and we haven't seen the issues you are describing. Still I do agree that if you don't want to change the culture via the official means you will encounter problems. One has to reload the application (via HTTP request) to set the current culture. This should reestablish a new SignalR connection  with the right culture (AFAIK). 

Radzen applications (ones generated via the Radzen IDE) change the current culture by navigating to a controller.

```
            var redirect = new Uri(UriHelper.Uri).GetComponents(UriComponents.PathAndQuery | UriComponents.Fragment, UriFormat.UriEscaped);

            var query = $"?culture={Uri.EscapeDataString((string)args)}&redirectUri={redirect}";

            UriHelper.NavigateTo("Culture/SetCulture" + query, forceLoad: true);
```

The controller itself is this:

```
    [Route("[controller]/[action]")]
    public partial class CultureController : Controller
    {
        public IActionResult SetCulture(string culture, string redirectUri)
        {
            if (culture != null)
            {
                HttpContext.Response.Cookies.Append(
                    CookieRequestCultureProvider.DefaultCookieName,
                    CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(culture)));
            }

            return LocalRedirect(redirectUri);
        }
    }
```

This approach should not cause a problem where one user changes the culture of a different user. Here is a test that I made - two users browser the same application with different cultures.
localization

javiercampos:
@akorchev I made a very simple (based off the blazor template in VS code, I added radzen just to show off, but it's not needed):

https://github.com/javiercampos/CultureProblemSample

In particular: https://github.com/javiercampos/CultureProblemSample/blob/master/Pages/Index.razor

I have Chrome (on the left) set to English and Firefox (on the right) set to Spanish (although I made the controller so you can switch via a cookie with the exact code you are using). Look what happens when I update the date in Firefox (in Spanish), to the Chrome window (or vice-versa)

BpHKDgZh8F

When updating something in the Spanish browser which raises a "StateHasChanged" (even if Invoked on the component), the `CurrentUICulture` on the renderer for the "English browser" is still "Spanish" (because blazor server doesn't know about browsers, and doesn't maintain the culture for their Invokation).

Of course this is a contrived example with a simple event, but I personally use domain bus events (locally via events like this or distributed using rabbitmq) to update data that might need a rerender (and the invocation is done from other user's circuit, or from the rabbitmq listener thread, which do not have its `CurrentUICulture` set by the middleware).

I have some workarounds for it (and again, the Blazor team needs to flow the CurrentUICulture from different invocations and/or the blazor renderer, but *they don't as of now*). Implementing that workaround on Radzen components would mean deriving off all components that need it, and again, as this PR shows, it's a relatively easy thing to do if you set it on your base classes.

akorchev:
Can you reproduce the problem without using a static class? I have no idea how static classes (and events) behave in ASP.NET core environment - for example how they capture the current thread etc. It seems after commenting `StaticData.UpdateDate();` everything behaves as expected - both browsers remain with the same culture.

By the way should the default value of Culture be CurrentCulture instead of CurrentUICulture? According to [this SO answer](https://stackoverflow.com/questions/6795891/fxcop-ca1305-currentculture-vs-currentuiculture) this seems to be the case.

akorchev:
@ThnxDaniel there seems to be a problem with the format in Day view (should display only hours - not it displays the date and time):

image

ThnxDaniel:
> @ThnxDaniel there seems to be a problem with the format in Day view (should display only hours - not it displays the date and time):
> 
> image

Had forgotten to pass it along from the RadzenDayView, but should be fixed now

akorchev:
@ThnxDaniel thanks! What do you think about the default value of the Culture property? Shouldn't it be CurrentCulture as per [this](https://stackoverflow.com/questions/6795891/fxcop-ca1305-currentculture-vs-currentuiculture) and [this](https://stackoverflow.com/questions/329033/what-is-the-difference-between-currentculture-and-currentuiculture-properties-of)?

Radzen.Blazor/RadzenMonthView.razor
akorchev: Let's remove that capitalization from the code. If needed it can be achieved with CSS.

ThnxDaniel:
> @ThnxDaniel thanks! What do you think about the default value of the Culture property? Shouldn't it be CurrentCulture as per [this](https://stackoverflow.com/questions/6795891/fxcop-ca1305-currentculture-vs-currentuiculture) and [this](https://stackoverflow.com/questions/329033/what-is-the-difference-between-currentculture-and-currentuiculture-properties-of)?

Reading through the documentation, it's not entirely clear as to which one would be correct for the UI portion of a web app. But I've changed it to CurrentCulture to be consistent with the rest of the codebase

javiercampos:
> Can you reproduce the problem without using a static class? I have no idea how static classes (and events) behave in ASP.NET core environment - for example how they capture the current thread etc. It seems after commenting `StaticData.UpdateDate();` everything behaves as expected - both browsers remain with the same culture.

The fact that is "other thing updating the bound values other than a direct response to the UI in that same circuit" is the culprit. I used a static class and an event to illustrate the problem, but anything that forces a re-render (updating the bound data from a background service, from other user's circuit, etc) that doesn't happen in the thread (and I mean the thread... even doing invocations in an async method where the thread -not the synchronization context, because the blazor one doesn't capture the thread-) where the middleware set the culture in, will exhibit this problem. It's only not a problem if you don't do any kind of realtime updates and only respond to the UI events on the same signalr circuit as the user. 

> By the way should the default value of Culture be CurrentCulture instead of CurrentUICulture? According to [this SO answer](https://stackoverflow.com/questions/6795891/fxcop-ca1305-currentculture-vs-currentuiculture) this seems to be the case.

CurrentUICulture should be preferred, but that also means having to specify it -anywhere- you render values. I myself have no preference, to me, Current Culture is always the same as the UI one, but the only reason of CurrentUICulture exists is to be used on UI paints

akorchev:
@ThnxDaniel Thanks! AFAIK CurrentUICulture is mostly used for resx files and satellite assemblies. Date and number formatting should probably follow CurrentCulture (according to [this](https://stackoverflow.com/questions/329033/what-is-the-difference-between-currentculture-and-currentuiculture-properties-of) and my prior experience. Still as @javiercampos mentions CurreentCulture and CurrentUICulture are set to the same thing in most cases. 

I will remove the capitalization code in the month view and merge the pull request.

akorchev:
@ThnxDaniel to capitalize the heading with CSS you can use the following:

```
.rz-scheduler-nav-title {
   text-transform: capitalize;
}
```

javiercampos:
Taken this has been merged, can we reopen https://github.com/radzenhq/radzen-blazor/issues/206 (since it's a problem for more components than the scheduler) and I may try to make some time to do a PR myself if noone else is interested?

PR #232, opened 2021-09-27
Add culture and default culture cascaded. Use in datagrid and numeric templates

javiercampos:
Based on https://github.com/radzenhq/radzen-blazor/issues/206

*Warning*: there's a test, `Numeric_Respect_MinParameter` not passing for `RadzenNumeric`, however it doesn't seem to be related to any of the changes here (it didn't pass either without making any changes to `RadzenNumeric` itself)

There might be places where I haven't found templates using automatic conversion that might require "Culturization".

The `RadzenTree` in particular I've left out uses a `Func<object, string>` value for making the conversion, but seems to only handle (it builds a expression with a cast on the function `Getter`) cases where the property is already a string, and strings should not try to be converted.


Radzen.Blazor/RadzenAutoComplete.razor
```diff
@@ -23,7 +23,7 @@
                     @foreach (var item in LoadData.HasDelegate ? Data != null ? Data : Enumerable.Empty<object>() : (View != null ? View : Enumerable.Empty<object>()))
                     {
                         <li role="option" class="rz-autocomplete-list-item" @onclick="@(() => OnSelectItem(item))" onmousedown="Radzen.activeElement = null">
-                            <span>@PropertyAccess.GetItemOrValueFromProperty(item, TextProperty)</span>
+                            <span>@PropertyAccess.GetItemOrValueFromProperty(item, TextProperty).ToCulturedString(Culture)</span>
```
akorchev, line 26: I think everything that uses TextProperty should be left as it is. In 99.9% it is used with a string property which should not be localized. Let's leave all similar code as it was.

javiercampos: (note that in any case, if you don't want to go there... I can revert those uses and make a custom template for those in my code -which is what I'm doing now-)

akorchev: Were you setting TextProperty to a Date or numeric property? The thing is ToCulturedString will do nothing in 99% of the cases which is needless overhead. The only legitimate use I see is in the DataGrid template.

javiercampos: @akorchev yes, I have (to a datetime in particular, in a dropdown, and a numeric in an autocomplete). Can't remember if it gave problems because I replaced it with a custom template to actually select the Culture (I currently do a `ToString(Culture)` on those), but I'd have had the case for it (that's one of the reasons I started the issue in first place)... I can see it may not be a typical case though, but it's indeed a legitimate case (I use a dropdown to select one in a series of specific dates that come from the database).

In any case, you decide, if you want me to revert those in `TextProperties` I can do, no problem (and keep using a custom template where needed). **My** main use case is the data grid (I have others, but I pass most of those values through a custom localization solution -which includes culture-aware conversions- so I need custom templates for those anyway).

akorchev: Yes, please revert all changes related to TextProperty. You can leave the one in the DataGrid though - it is legit as a grid column can display any type not just strings.

javiercampos: I've reverted the changes on those now. I've also changed the title of the PR tos pecify that it's only datagrids and numerics now (and I made it compatible with https://github.com/radzenhq/radzen-blazor/pull/229 so that it works now on both the direct parameter and the cascading parameter).

Radzen.Blazor/RadzenCompareValidator.cs
```diff
@@ -30,7 +30,7 @@ private int Compare(object componentValue)
             switch (componentValue)
             {
                 case String stringValue:
-                    return String.Compare(stringValue, (string)Value, false, CultureInfo.CurrentCulture);
+                    return String.Compare(stringValue, (string)Value, false, GetComponentCulture());
```
akorchev, line 33: Why not just `Culture`? What is the purpose of the `GetComponentCulture` method?

javiercampos: That's right, `GetComponentCulture()` was the original method in the first commit to get the specified/default/currentculture and made a virtual method to make it like the GetComponentCss()/etc. methods, but I changed it later to resolve on the property itself to allow it to work with `IScheduler` (instead of creating a specific interface implementation) and left that there. I'll remove it.

By the way, I've noticed that `IScheduler` has a setter on the `Culture` property which is not used (and I believe that should be removed from the interface). This is generally bad in Blazor because changing a parameter property directly on the property (and not through a parameter change) does not raise `SetParametersAsync` nor any of the parameter-related component lifecycle methods.

Removing the setter from the interface does not seem to bear any consequences (because it's only used for reading in child components). Should we remove that setter, while we're at it?

PR #268, opened 2021-11-04
Bound private filter properties to the public properties.

carle50:
Bound private filter properties to the public properties.


EthanB911 (approved):

enchev:
Hey @carle50,

We cannot merge this since parameters should be use to keep state. We use private properties for that. Check this issue for more info: https://github.com/dotnet/AspNetCore.Docs/issues/17178

PR #277, opened 2021-11-12
Filter DataGrid by Multiselect ComboBox with LoadData event defined

nstuller:
This block of code (or something similar) needs to be added to the QueryableExtension file to properly handle a multi-select filter. I have seen sample code for multi-select dropdowns filtering data grids, but they do not work. This is because the FilterValue is stored in an IEnumerable<string> but there is nothing in QueryableExtension that handles that in the ToFilterString method. I was able to get my scenario to work by using this code.

This code works with a LoadData event handler that is used to appropriately filter the grid items by the filter string generated across all columns.

### DIFF
--- a/Radzen.Blazor/QueryableExtension.cs
+++ b/Radzen.Blazor/QueryableExtension.cs
@@ -174,7 +174,7 @@ namespace Radzen
 
                         List<string> multivalueFilterList = new List<string>();
                         IEnumerable<string> filterValues = (IEnumerable<string>)column.GetFilterValue();
-                        //Combine all the above linqoperator stuff
+                        //Combine the linq queries for all filter values selected
                         foreach (string item in filterValues)
                         {
                             multivalueFilterList.Add(GetColumnFilter(column, item));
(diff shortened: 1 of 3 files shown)


enchev:
Hi @nstuller,

Thanks for the pull request!

I've tried the demo for custom filtering template with multiple selection however I've got the following error:
image

nstuller:
Happy to help. I have resolved the issue that @enchev pointed out. I also added a workaround because otherwise the multi-select doesn't get it's parameters set until after the reload that occurs in the OnChange method.

enchev:
Hi @nstuller,

I've reviewed further your code and I've noticed that it will work only for `IEnumerable<string>` (PropertyAccess.GetElementType(column.GetFilterValue().GetType()) == typeof(string)) and only if the DataGrid is bound using LoadData (ToFilterString() method is used only in this case). 
Unfortunately we will unable to accept this pull request. We will do our best to provide built-in filtering support for any `IEnumerable<>` in some of our next updates together with example.

PR #288, opened 2021-11-21
Expandable Indicator Column Visibility Flag

beeterry:
Currently on a DataGrid, if a template is set, it is assumed that we want the expandable indicator column rendered. I added a parameter that gives us the ability to toggle whether we want that column rendered.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -51,7 +51,7 @@
                 {
                     <col>
                 }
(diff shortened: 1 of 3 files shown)


enchev:
Hi @beeterry ,

Thanks for the pull request! Maybe we can name the property simply `ShowExpandColumn`? If you agree I'll perform the change and will merge the pull request.

enchev:
I've proceeded with the rename and I've fixed RadzenDataGridRow to take into account the new property.

PR #302, opened 2021-12-03
DropDownDataGrid panel width and IsoCurrency property on NumericTextBox

kostasxagoraris:
I have added 3 new features

1. I added the ability to set the panel width on dropdownDatagrid 
2. I added the ability to allow to change column header width on dropdownDatagrid 
3.  I added the ability to define the currency on numeric text box unrelated to the current culture (useful in multicurrency applications)

### DIFF
--- a/Radzen.Blazor.Tests/NumericTests.cs
+++ b/Radzen.Blazor.Tests/NumericTests.cs
@@ -350,5 +350,24 @@ namespace Radzen.Blazor.Tests
 
             Assert.Contains($" value=\"{valueToTest.ToString(format)}\"", component.Markup);
         }
+        [Fact]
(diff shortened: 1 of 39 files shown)


enchev:
Hi @kostasxagoraris,

We appreciate your effort however we will unable to accept your pull request. There is no need for `ShowPagesDropDown` since this DropDown will not be shown if you simply do not set `PageSizeOptions`. Currency symbol also for Numeric is not relevant since the component is not only for editing currencies but for general numeric types, maybe you can create your own CurrencyNumeric component by wrapping our Numeric with the additional functionality you need. The DropDown panel min-width can be controlled with CSS also, no need for such property.  We will expose `AllowColumnResize` for the DropDownDataGrid in our next update.

PR #313, opened 2021-12-19
Allow manual control legend elements visibility for plots

ArtemkaKun:
**ShowInLegend** property - public setter was added
The default value was set as _true_ to not to break the previous default value (Visible property is _true_ by default)
Preventing legend from rendering, when ShowInLegend is _true_ but Visible is _false_

### DIFF
--- a/Radzen.Blazor/CartesianSeries.cs
+++ b/Radzen.Blazor/CartesianSeries.cs
@@ -146,7 +146,7 @@ namespace Radzen.Blazor
         }
 
         /// <inheritdoc />
-        public bool ShowInLegend { get => Visible; }
+        public bool ShowInLegend { get; set; } = true;
 
         /// <summary>
         /// The name of the property of <typeparamref name="TItem" /> that provides the Y axis (a.k.a. value axis) values.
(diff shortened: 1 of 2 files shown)


akorchev:
This implementation is a breaking change. Toggling the Visible property of a series won't hide it from the legend.

ArtemkaKun:
Hi, thanks for a reply, this clear things :)

If I understood you right, you assume that with this change, when any plot will have ```Visible == false```, a series will be visible in a legend anyway.

I believe my second commit 633ba160114a1da3ee2ea971beeb6dd773beb9f0 prevents that kind of behavior.

Yes, I want to have a series that doesn't appear in the legend. As an example, I will attach a screenshot from my current project where I'm using Radzen lib.

image

- I want to define my custom "header" for this line plot.
- On phones screens because of legend label graph part is squished, so I want to place it on top of the section.

akorchev:
> I believe my second commit 633ba16 prevents that kind of behavior.

Check the Area chart demo in your pull request:
current

Compare it to how it behaves [live](https://blazor.radzen.com/area-chart):
live

I think you can hide the legend via CSS on mobile devices without this pull request:

```
@media (max-width: 768px) {
   .rz-legend {
       display: none;    
   }
}
```

ArtemkaKun:
😮 Ok, that behavior with clicking on a legend's member change a lot. Thanks for clarifying. 
I will close the PR, because it's useless in that case

PR #318, opened 2021-12-28
StepBase attribute

refactorer1:
Added support for a StepBase attribute which can be used to apply the step value to an other value than 0. if specified as "min", the value of the Min attribute is considered. If StepBase attribute is omitted, the control behaves like before, and Step value is based on 0.


akorchev:
We can't accept this pull request:

1. You seem to have change the formatting of the whole file. As a result everything is listed as changed. We won't be able to git blame anything if we accept it like it is.
2. The newly introduced property can be a number or the string "min". This is ambiguous.

We decided to make the default slider behavior based on the min value as it seems to be what `<input type="range">` (the built-in slider) does by default.

PR #338, opened 2022-01-20
DataGrid - CheckBox Only Selection 

vertonghenb:
If the datagrid is using multiselect or selectable rows, it should be possible to only Select items and not trigger the OnRowClick, For example when navigation to another page on normal cell click, but only select items when using the Checkbox.
This feature introduces no breaking changes, only additional opt-in behavior.

This feature is similar to Telerik's 
https://demos.telerik.com/blazor-ui/grid/checkbox-only-selection

checkboxselect


enchev:
Hey @vertonghenb,

Thanks for the pull request! I see what you are after, maybe it will be easier like this:
https://github.com/radzenhq/radzen-blazor/commit/db78b831b3a685ca468116bd43ce3726e933b1e6

PR #343, opened 2022-01-21
remove obsolete tabindex from surrounding radzeninputfile span

### LINKED ISSUE #342: RadzenFileInput redundant tabindex
Using RadzenFileInput render a formatting span including tabindex surrounding the input field
2022-01-21 12_57_22-STEAM-Cockpit

this causes the span to be focused first, then the input field when tabbing through. this breaks with accessibility regulations.

acordts:
remove obsolete tabindex from surrounding radzeninputfile span.
closes #342

### DIFF
--- a/Radzen.Blazor/RadzenFileInput.razor
+++ b/Radzen.Blazor/RadzenFileInput.razor
@@ -9,7 +9,7 @@
 {
     <div @ref="@Element" style="@Style" @attributes="Attributes" class="@GetCssClass()" id="@GetId()">
     <div class="rz-fileupload-buttonbar">
-        <span class="@ChooseClassList" label="Choose" tabindex="@TabIndex">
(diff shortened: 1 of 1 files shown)


enchev:
Hey @acordts,

Thanks for your pull request however this will break immediately TabIndex property which is not acceptable. What we can do is to execute file open dialog on this span enter key.

acordts:
hi, thanks for review that fast. 
your idea doesn't solve the problem and cause a new issue: open the dialog twice on using "enter" the input. I close the PR for the moment and try another approach.

acordts:
Dear @enchev,

excuse my ignorance, the tabindex="@TabIndex" must be part of the span-tag? If not, a shift of the tabindex parameter to the input-tag could be a way out.

Bests

PR #344, opened 2022-01-21
Add ColumnPicker

vertonghenb:
Functionality to choose which columns you want to show.

picker

This PR does not include breaking changes, but opt-in. Visible parameter is not impacted and is still the primary source. Unless you override it.

@enchev, maybe we can also put it in the footer, but I'm not sure if it makes sense.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -5,6 +5,19 @@
 @typeparam TItem
 @inherits PagedDataBoundComponent<TItem>
 
+@if(Visible && AllowColumnPicking)
+{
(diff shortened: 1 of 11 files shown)


vertonghenb:
@enchev  any news about this one?

enchev:
We like what you’ve submitted however we would like @yordanov to review this and maybe provide more consistent way for such DataGrid extensions, maybe via some special template, toolbar or similar. Unfortunately at the moment we are busy with Radzen 3.0.

enchev:
Hi @vertonghenb,

@yordanov made some changes to the pull request however I have one more note. Visible property of the column is public and often used. Changing the type from bool to bool? might brake existing code, please use different property and do not change the type of Visible property.

PR #353, opened 2022-01-27
Diacritics symbol insencitive

kostasxagoraris:
I have added the ability to perform an accent insencitive search on all components that inherit from dropdownbase since it is a hashle for those who right on languages with accents and other diacritics to search with or without accents. I also included the option to ignore symbols as well. I share the code with you in case you are interested on including it in the upstream. I have not yet added support for search on datagrid because I must also take into account linw to sql search.


enchev:
Hey @kostasxagoraris,

Thanks for your pull request however I'm afraid that we cannot accept it. Your change breaks immediately IQueryable support since such expression cannot be translated:
image

kostasxagoraris:
Yes ..what I have made does not work on IQueryable . I guess you are right
My bad! I was thinking according to my project (Because I use the grid and
the databound controls like this : datagrid --> database access,
dropdowns-->lists.

2B || ! 2B ?

Kostas X...

PR #354, opened 2022-01-28
Add Click Parameter on MenuItem

vertonghenb:
Currently it's only possible to provide a `Path` parameter to the children of the `<Menu>` and the `<Menu>` takes full control. 

This PR makes it possible to also add the `Click` EventCallback to the children (`MenuItem`), this way we can use additional logical before navigation and have more control of the component if we don't want to use the `Click` of the parent

This PR introduces no breaking changes, but an additional parameter `Click` to the `MenuItem`

menuitem-click

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Forms;
+using Microsoft.AspNetCore.Components.Web;
 using Radzen.Blazor;
 using System;
 using System.Collections;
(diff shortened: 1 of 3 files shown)


enchev:
Hey @vertonghenb,

Thanks for your pull request however RadzenMenuItem already have onclick propagated to the menu:
https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenMenuItem.razor#L5
https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenMenuItem.razor.cs#L84-L90

Using Value of the event argument you can check which exactly menu item is clicked.

vertonghenb:
@enchev true, but it's quite tedious and can be error prone if you forget one of the values.

PR #369, opened 2022-02-10
RTL Support Added. (Styles fixed)

vrassouli:
To add RTL Support, I've changed all styles of paddings, margins, any left-right styles like border-left, ... to use their 'inline' variant.
For example, 
`
margin: 1rem 2rem 3rem 6rem;
padding-left: 2.5rem;
padding-right: 4.5rem;
border-left: 1px solid red;
`
is converted to:
`
margin: 1rem 0 3rem 0;
margin-inline: 6rem 1rem;
padding-inline-start: 2.5rem;
padding-inline-end: 4.5rem;
border-inline-start: 1px solid red;
`

### DIFF
--- a/Radzen.Blazor/themes/components/blazor/_accordion.scss
+++ b/Radzen.Blazor/themes/components/blazor/_accordion.scss
@@ -8,10 +8,11 @@ $accordion-item-font-weight: 600 !default;
 $accordion-icon-width: 1.5rem !default;
 $accordion-icon-height: $accordion-icon-width !default;
 $accordion-icon-font-size: 1.5rem !default;
-$accordion-icon-margin: 0 0.375rem 0 0 !default;
+$accordion-icon-margin-inline: 0 0.375rem !default;
 $accordion-selected-color: $default-link !default;
 $accordion-hover-color: $accordion-selected-color !default;
-$accordion-content-padding: 0.5rem 0.5rem 0.5rem 2rem !default;
+$accordion-content-padding-vertical: 0.5rem 0 !default;
+$accordion-content-padding-inline: 2rem 0.5rem !default;
 $accordion-content-font-size: $accordion-item-font-size !default;
 $accordion-border-radius: $border-radius !default;
 $accordion-item-border: none !default;
@@ -42,7 +43,7 @@ $accordion-shadow: none !default;
 
   .rzi {
     font-size: $accordion-icon-font-size;
-    margin: $accordion-icon-margin;
+    margin-inline: $accordion-icon-margin-inline;
   }
 }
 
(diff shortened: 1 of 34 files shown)


vrassouli:
The next step, is to fix some javascripts
For example, while resizing column width of a datagrid, the direction should be switched.

akorchev:
Hi @vrassouli,

How can we test your PR in RTL mode? There are three major issues we see in it:

1. You have changed the names of some SASS variables which breaks Radzen's Theme Builder.
2. Support for the `padding-inline` CSS attribute was [only recently added to Safari](https://caniuse.com/?search=padding-inline). If we accept your pull request all Radzen themes will get broken on iOS devices that are not up-to-date.
3. RTL CSS rules are included in LTR themes as well which is redundant.

While the first issue is solvable - just don't rename existing property names the second is a dealbreaker for us as it is a breaking change.

We are not RTL experts but can't this be implemented by other means? For example with an extra selector for `dir=rtl` and a helper SASS function which reverses the left and right values?

```
.rz-accordion-content {
   font-size: $accordion-content-font-size;
   background-color: $accordion-item-background-color;
   padding: $accordion-content-padding;

   [dir=rtl] & {
      padding: rtl($according-content-padding);
   }
 }
```

Or we can make it simpler by hiding RTL support behind a SASS boolean variable and build a separate rtl version of the theme. This would solve the third issue too as RTL CSS won't be downloaded in LTR themes and vice versa. Something in the lines of:

```
$accordion-content-padding: rtl(0.5rem, 0.5rem, 0.5rem, 2rem) !default
```

where the rtl function (which does not exist yet) would reverse the direction depending on a global $rtl flag.

We will discuss the details internally and post our thoughts later.

vrassouli:
Hi @akorchev 
Thanks for the review

Here is a link, so you can view the output of this change:
http://radzen.vhdr.ir/datagrid
There is a RTL-switch button on the header of the page, next to the theme-switch dropdown.

I think, we have 3 options to support RTL:

1. To provide a mini CSS file, containing only the required style overrides.
2. To provide a full CSS file(s), containing the RTL version of the default styles.
3. To make the default style, support both directions natively

Maintaining option 1, would become a nightmare as the original styles will be changed over time, and we should always check and sync the overrides.
Option 2, is good and can use saas functions to convert styles of LTR to RTL and provide a new set of css files which are all in RTL direction. But, this would require a separate distribution and the user, has to include both set of (mostly identical) versions of the styles. And to prevent including both, the user has to do a complete refresh of the page, and dynamically switch the included styles.
Option 3, is the way I decided to go. this way, we only distribute one set of styles, with some minor changes and additional styles (like the ones you mentioned at issue 3). So the user would include only one set of files, and changing directions (may be in response to changing language\culture in a multilingual application) wouldn't require a full refresh of the page!

akorchev:
I don't think it makes sense to include both RTL and LTR styles in the same CSS file. We have to think of a better way to handle it. Creating a separate RTL version of every CSS file could be handled at build time.

Unfortunately we can't accept your pull request considering those requirements that we have (RTL CSS should not be downloaded by non-RTL users and we should support iOS devices from 2021).

vrassouli:
Ok
Please let me know if I can help with the this.
I may try to separate the RTL version, and would request for another pull for review.

Thanks

Issue #373, opened 2022-02-12
JSDisconnectedException in RadzenDataGrid Dispose


dxrdxr:
**Describe the bug**
I added a NotificationService.Notify to my server side Blazor app and it is now repeatedly throwing `JSDisconnectedException` when I refresh the page.

The call stack:
```
Microsoft.AspNetCore.Components.Server.dll!Microsoft.AspNetCore.Components.Server.Circuits.RemoteJSRuntime.BeginInvokeJS(long asyncHandle, string identifier, string argsJson, Microsoft.JSInterop.JSCallResultType resultType, long targetInstanceId) Line 116
	at /_/src/Components/Server/src/Circuits/RemoteJSRuntime.cs(116)
Microsoft.JSInterop.dll!Microsoft.JSInterop.JSRuntime.InvokeAsync<Microsoft.JSInterop.Infrastructure.IJSVoidResult>(long targetInstanceId, string identifier, System.Threading.CancellationToken cancellationToken, object[] args) Line 132
	at /_/src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs(132)
Microsoft.JSInterop.dll!Microsoft.JSInterop.JSRuntime.InvokeAsync<Microsoft.JSInterop.Infrastructure.IJSVoidResult>(long targetInstanceId, string identifier, object[] args) Line 91
	at /_/src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs(91)
Microsoft.JSInterop.dll!Microsoft.JSInterop.JSRuntime.InvokeAsync<Microsoft.JSInterop.Infrastructure.IJSVoidResult>(string identifier, object[] args) Line 69
	at /_/src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs(69)
Microsoft.JSInterop.dll!Microsoft.JSInterop.JSRuntimeExtensions.InvokeVoidAsync(Microsoft.JSInterop.IJSRuntime jsRuntime, string identifier, object[] args) Line 32
	at /_/src/JSInterop/Microsoft.JSInterop/src/JSRuntimeExtensions.cs(32)
[Resuming Async Method]
[...]

PR #375, opened 2022-02-16
JSDisconnectedException in RadzenDataGrid Dispose

dxrdxr:
As described in #373, in .Net 6.0, there is a race condition where the `JSRuntime` will become unavailable
before a component's `Dispose` is able to be called (may take upwards of 60 seconds). As a result, Blazor will throw a `JSDisconnectedException` on each subsequent call to `JSRuntime.InvokeVoidAsync`.

The fix is to wrap the calls in `Dispose` to catch `JSDisconnectedException`  and set `IsJSRuntimeAvailable` to `false`.
Because `JSRuntime.InvokeVoidAsync` is in fact async, the code must `await` the result in order to catch exceptions.
This pattern of calling into the runtime in `Dispose` is in many components so a wrapper was created to handle the
common boilerplate code needed to wrap the component specific functionality.

The improvement is most notable in the DataGrid component which calls the runtime once per column, which may trigger 4-5 exceptions being thrown per column! After this fix the first exceptions are caught and further exceptions prevented.

This change in behavior is only observed in .Net 6.0, so as part of this fix a new Target was enabled and the associated changes were made.

### DIFF
--- a/Radzen.Blazor/RadzenSidebar.razor
+++ b/Radzen.Blazor/RadzenSidebar.razor
@@ -2,7 +2,7 @@
 
 @if (Visible)
 {
-    @if (Responsive)
+    @if (IsResponsive)
     {
         <RadzenMediaQuery Query="(max-width: 768px)" Change=@OnChange />
     }
--- a/Radzen.Blazor/RadzenSidebar.razor.cs
+++ b/Radzen.Blazor/RadzenSidebar.razor.cs
@@ -23,9 +23,12 @@ namespace Radzen.Blazor
         /// <summary>
         /// Toggles the responsive mode of the component. If set to <c>true</c> (the default) the component will be 
         /// expanded on larger displays and collapsed on touch devices. Set to <c>false</c> if you want to disable this behavior.
+        /// Responsive mode is only available when RadzenSidebar is inside <see cref="RadzenLayout" />.
         /// </summary>
         [Parameter]
         public bool Responsive { get; set; } = true;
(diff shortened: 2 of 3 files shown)


enchev:
Hey @dxrdxr,

Thanks for your pull request however the SASS build is broken after your change. I don't see also where NET5_0_OR_GREATER is defined.

dxrdxr:
> 

I see the SASS problem. I think I have an idea what is broken. I look at it later today.

enchev:
Maybe we need to restore Before/AfterTargets and the Condition for SASS related targets.

dxrdxr:
Fixed. In trying to optimize the multi-target case I created a bug, in fixing that bug I introduced this bug. :-(

In figuring out what went wrong, I noticed that the csproj file is duplicating the `SASS` Target without making any changes other than using `Sass` instead of `SassFile`. By changing the `ItemGroup` to `SassFile` it allows the redundant target to be removed and also enables the `LibSassBuilder` optimizations such as change detection based on a cache of the hashes, which can save 10 seconds off an incremental build by skipping unnecessary Sass builds.

enchev:
Unfortunately there is another issue - navigation seems broken:
image
There are no errors however clicking the navigation menu items and event getting started button will not navigate - just the url in the address bar is changed.

dxrdxr:
The problem was a deadlock in the JSRuntime caused by misuse of `await`.

New Fix/PR later today.

PR #378, opened 2022-02-19
Handle JSDisconnectedException when JSRuntime goes away before Dispose

dxrdxr:
As described in [#373](https://github.com/radzenhq/radzen-blazor/issues/373), in .Net 6.0, there is a race condition where the `JSRuntime` will become unavailable
before a component's `Dispose` is able to be called (may take upwards of 60 seconds). As a result, Blazor will throw a `JSDisconnectedException` on each subsequent call to `JSRuntime.InvokeVoidAsync`.

The fix is to wrap the calls in `Dispose` to catch `JSDisconnectedException` and set `IsJSRuntimeAvailable` to `false`.
Because `JSRuntime.InvokeVoidAsync` is in fact async, the code must await the result in order to catch exceptions.
This pattern of calling into the runtime in `Dispose` is in many components so a wrapper was created to handle the
common boilerplate code needed to wrap the component specific functionality.

The improvement is most notable in the DataGrid component which calls the runtime once per column, which may trigger 4-5 exceptions being thrown per column! After this fix the first exceptions are caught and further exceptions prevented.

This change in behavior is only observed in .Net 6.0, so as part of this fix a new Target was enabled and the associated changes were made.

### DIFF
--- a/Radzen.Blazor/DropDownBase.cs
+++ b/Radzen.Blazor/DropDownBase.cs
@@ -16,7 +16,7 @@ namespace Radzen
     /// <typeparam name="T"></typeparam>
     public class DropDownBase<T> : DataBoundFormComponent<T>
     {
-#if NET5
+#if NET5_0_OR_GREATER
         internal Microsoft.AspNetCore.Components.Web.Virtualization.Virtualize<object> virtualize;
 
         /// <summary>
@@ -74,7 +74,7 @@ namespace Radzen
         /// <returns><c>true</c> if virtualization is allowed; otherwise, <c>false</c>.</returns>
         internal bool IsVirtualizationAllowed()
         {
-#if NET5
+#if NET5_0_OR_GREATER
             return AllowVirtualization;
 #else
             return false;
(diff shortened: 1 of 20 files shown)


enchev:
I still see the same exception after refresh in Visual Studio output:
image

dxrdxr:
Correct. That exception will be thrown because the JSRuntime is not in the initialized state. Unfortunately there is no documented interface to query the state, only the exception.

The fix is to stop calling the runtime once it is detected that an exception occurred. For a simple component this is not significant, but on a complex component like a DataGrid it can save dozens of unnecessary exceptions.

An alternative implementation that was considered but not done was to rewrite all of the component's to use DisposeAsync(). That would provide a cleaner way to catch the exception as a proper await could be used. I felt the benefit was very minor and the risk of breaking something else was not worth it.

enchev:
Unfortunately I'm not sure in this case what's fixed in your pull request. @akorchev let us know your opinion.

akorchev:
If there is no way to safely call JSRuntime I don't think we should jump through hoops for that. Even if we catch the exception it will still break in the debugger. I am also not a fan of wrapping all JSRuntime access code - Microsoft don't seem to do that in their Blazor code so it must not be required.

enchev:
I'm afraid that we cannot accept this pull request. The changes are too many and drastic and I don't see any important benefit.

PR #383, opened 2022-02-21
add AllowRowSelectOnRowClick property at RadzenDropDownDataGrid

pigwing:
when use DropDown component in RadzenDropDownDataGridColumn template,change the DropDown value will trigger RadzenDataGrid.OnRowSelect


pigwing:
动画
here's the problem. this requests can solve it.^_^

PR #386, opened 2022-02-25
Updated the behavior of is checked to correctly evaluate the bottom l…

grwomack:
Hi Everyone,

I updated the behavior of the RadzenTree and TreeItem when the user chooess to display them as checkboxes.

I added a new property to RadenTree:
[Parameter] public string ChildrenPropertiesChain { get; set; } = string.Empty

If the user adds the chain of children as a string, the checkboxes will maintain the correct IsChcecked state.  

I created an example on the https://blazor.radzen.com/tree-checkboxes page.

Best,
Garrett

### DIFF
--- a/Radzen.Blazor/RadzenTree.razor.cs
+++ b/Radzen.Blazor/RadzenTree.razor.cs
@@ -156,6 +156,22 @@ namespace Radzen.Blazor
         /// </summary>
         [Parameter]
         public IEnumerable<object> CheckedValues { get; set; } = Enumerable.Empty<object>();
+        
+        /// <summary>
+        /// Gets or sets the checked values. Use with <c>@bind-CheckedBottomLayerValues</c> to sync it with a property.
+        /// </summary>
+        [Parameter]
+        public IEnumerable<object> CheckedBottomLayerValues { get; set; } = Enumerable.Empty<object>();
+        
+        [Parameter] public string ChildrenPropertiesChain { get; set; } = string.Empty;
+        
+        internal Dictionary<object, IEnumerable> LowestLevelChildrentDict { get; set; } 
+        private string[] ChildrenPropertyNames => ChildrenPropertiesChain.Split('.');
+        
+        internal void SetBottomLayerCheckedValues(IEnumerable<object> values)
+        {
+            CheckedBottomLayerValues = values.ToList();
+        }
 
         internal List<RadzenTreeItem> items = new List<RadzenTreeItem>();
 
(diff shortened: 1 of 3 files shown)


akorchev:
Thank you for the pull request. Unfortunately I do not understand the implementation and the intent of this feature.

1. What is the second property even doing?
2. Why is the ChildrenPropertiesChain property a string? 
3. Why is it `.` separated? 
4. What if the tree is self-referencing? 
5. What if there are unlimited number of levels?

If it is related to [this forum thread](https://forum.radzen.com/t/multilayer-checkbox-tree-bad-checkbox-display-when-collapsing-layers/10170/2) - we recognize this as an issue and will search for a solution.

grwomack:
1) What is the second property even doing?
    i) [Parameter] public IEnumerable<object> CheckedBottomLayerValues { get; set; } = Enumerable.Empty<object>();
    ii) This property is used to hold the values of the lowest level layer that are checked in the tree.  If the user checks value has 
        children then only the lowest level children are added to this enum.

2) Why is the ChildrenPropertiesChain property a string?
   i) Given the implementation of TreeLevel where you provide the Children property, I tried to mimic this by passing a string 
      separated by "." to the tree so it knew how many levels there were and what the proepty name of the children at each level 
      should be.
   ii) I am open to it being any implementation.  I just tried to follow the style that I saw in the definition of the tree layer.

3) Why is it . separated?
   i) The dot separates each IEnumerable that would be the children from the top layer.  
   ii) On the https://blazor.radzen.com/tree-checkboxes, if you look at the parent class, to get the lowest level  you would need to call Parent.Children.Grandchildren.GrandGrand.Children.  I tried to mimic this in this input to the ChildrenPropertiesChain property.  
   iii) Open to any other suggestion on how to implement this.
4) What if the tree is self-referencing?
   i) Not sure how that changes my implementation.  The tree has property named Data which contains the data to be represented in the tree.  I create a dictionary which maps the value of each node in the tree to the lowest children that the node can have in the tree.  I checked the "checked" status of the lowest level items in the tree only when the IsChecked method is called.  That is why I added the [Parameter] public IEnumerable<object> CheckedBottomLayerValues { get; set; } = Enumerable.Empty<object>(); because it keeps track of the checked status of the lowest level children only. 
5) What if there are unlimited number of levels?
   i) This implementation cannot handle an infinite amount of layers.
   ii)  I found that the base implementation couldn’t not handle more than 2 layers so I added this code to handle N layers.

PR #390, opened 2022-03-02
Bread Crumb Component

MPapst:
Introduces a new BreadCrumb Component, which takes a `IEnumerable` of custom objects adn displays them in a bread crumb menu style.

Also optionally allows to customize the display using a template and a template for the item separation.

I added a new class `rz-breadcrumb` without any style information in css.

Let me know if I should change anything


akorchev:
Thank you for the pull request @MPapst!

A breadcrumb component would be a great addition to Radzen.Blazor!

I have some feedback that we would like to be addressed before we merge the pull request:

1. The RadzenBreadCrumb component should support the `Visible`, `Style` and custom attributes as any other Radzen Blazor component. You can check how other components are implemented for example RadzenTabs. Its CSS class should be `rz-breadcrumb`.
2. RadzenBreadCrumbItem does not need to inherit from RadzenComponent. Its CSS class should be `rz-breadcrumb-item`.
3. We think RadzenBreadCrumb should not be a data-bound component and does not need a Data property - it would require the user to create a class just for the breadcrumb. Also breadcrumbs are rarely database driven. We think it should be defined in a ways similar to RadzenPanelMenu/RadzenMenu (or at least support that option as well as having a `Data` property):
    ```
    <RadzenBreadCrumb>
       <RadzenBreadCrumbItem Text="Text" Path="/path" Icon="icon" />
    </RadzenBreadCrumb>
    ```
4. The rendering appearance of the separator should be controlled from the theme CSS and not via properties such as Separator. A `_breadcrumb.scss` file would be needed in the `themes/components/blazor` directory. I would probably make `rz-breadcrumb` have `display: flex; align-items: center`.

MPapst:
Hi @akorchev ,

I think I understand your points, I will add the stylings as soon as possible.
Regarding point 3, the data bound component. My thoughts were to have a the breadcrumb menu created by a service, based on the current navigation position or based on the user navigation (e.g. capture the `NavigationManager` navigation event and display the last 5 locations) on a central point in the application.

What would you think about offer both options?
I could create a `RadzenBreadCrumb<TItem>` Component offering a Type Property and Data Bound components and a `RadzenBreadCrumbMenu` using the `RadzenBreadCrumb<TItem>` to allow putting the elements into the ChildContent RenderFragment. 
I am not sure, but this could also work with the same component.

Unfortunately I cannot offer a tight timeline for the upcoming changes as I will be without internet for two weeks starting on saturday.

MPapst:
I managed to have some time today.

Unfortunately there is one point I can not solve - maybe you've got an idea about that:
When using Child Content Items I still need to supply a Type Parameter for the component (it is not used, but it is in the definition). Adding a non generic class inheriting `RadzenBreadCrumb<object>` does not work with the same name.
Removing the type param would mean to have an object parameter passed to the Template if used.

Possible solution would be:
```csharp
public class RadzenBreadCrumbMenu : RadzenBreadCumb<object> { }
```
But this introduces a different component.

akorchev:
When a component has child items it should be untyped (similar to RadzenTree). I think this is OK - a cast in the template would not be too bad.

MPapst:
@akorchev Please have a look, I did a rework and removed the data bound part.

akorchev:
There are a few items from my previous review which are not addressed. Check them out.

MPapst:
Am I missing something?

1. added `Visible`, `Style`, `attributes`, CSS class is `rz-breadcrumb`
2. CSS class of item is `rz-breadcrumb-item`
3. Removed Data binding in favor of child content. Replaced `Items` with `ChildContent` property
4. CSS is added and controls how the separator looks

akorchev:
Yes, here are they:

1. You have modified files that you probably shouldn't have - Radzen.Blazor.csproj, RadzenCheckBoxList, bootstrap files. Please revert those to their original state.
2. Probably there isn't a need to inherit RadzenBreadCrumbItem from RadzenComponent.
3. The Link property should be Path for consistency with other components.
4. We render the CSS class of components like this: https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenLabel.razor#L5 Please use the same approach.

MPapst:
Corrected. Thanks for the summary.

I would like to leave the item inheriting the RadzenComponent, as this brings the the standard properties even for the items (like Style).

PR #400, opened 2022-03-16
Datepicker footer placeholder

mmmmmarco:
This PR adds the ability to optionally add custom functionality to the Radzen datepicker, in an easy and flexible way. For example a today button, as added as a demo.

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor
+++ b/Radzen.Blazor/RadzenDatePicker.razor
@@ -96,6 +96,9 @@
                                 </tbody>
                             </table>
                         </div>
+                        <div class="rz-datepicker-footer">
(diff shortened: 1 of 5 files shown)


Radzen.Blazor/RadzenDatePicker.razor
```diff
@@ -96,6 +96,9 @@
                                 </tbody>
                             </table>
                         </div>
+                        <div class="rz-datepicker-footer">
+                             @FooterTemplate
```
enchev, line 100: It will be good to check if the FooterTemplate is defined before rendering this element

mmmmmarco: Done

PR #409, opened 2022-03-30
Add Minimum Date, Maximum Date, Initial Month/Year, and SameStyleForA…

andrewaggb:
I added some functionality to the DatePicker.

I think it's straightforward and the changes seem quite safe to me.  None of these are breaking changes and none are required.  omitting them results in identical behavior to the current calendar.  

New Parameters

InitialMonth.  There wasn't a way to have the calendar load on a different month than DateTime.Today.  This is a very simple change to let you pick some other time period so the initial popup doesn't have to be the current month.  It was the reason I forked the project.  My main use case is to select a future date at least a week away and no more than 3 months away.  Near the end of the month that means no days are available in the current month so I'd rather the calendar started with next month.

MinimumDate and/or MaximumDate.  Simple way using parameters to disable everything before and/or after a particular date (and handles the year drop down).  You could accomplish the same thing a custom daterender function and setting the YearRange parameter but I think this is simple and intuitive and similar to other calendars.

SameStyleForAllDays. (Better name is welcome).  Currently days belonging to a month other than the selected month are visible in the calendar but look disabled because the razor page is applying the class 'rz-state-disabled'.  I understand the style choice but I don't want those days to look disabled if my daterender function says it isn't.

Notes:
One thing that's slightly different is the initial value of Disabled in DateRenderEventArgs could be True if a date is less than the minimum or greater than the maximum.
DateRender can still override it.
But if minimum and/or maximum dates are in use your daterender could look like this where you take the initial value into account.
    void DateRender(DateRenderEventArgs args)
    {
        args.Disabled = **args.Disabled** || dates.Contains(args.Date);
    }


Radzen.Blazor/RadzenDatePicker.razor.cs
```diff
@@ -249,6 +251,35 @@ protected override void OnInitialized()
         [Parameter]
         public string InputClass { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Minimum Selectable Date.
+        /// </summary>
+        /// <value>The Minimum Selectable Date.</value>
+        [Parameter]
+        public DateTime? MinimumDate { get; set; }
```
enchev, line 259: Maybe these properties can be named simply Min and Max similar to other components like RadzenNumeric, RadzenSlider, etc .

Radzen.Blazor/RadzenDatePicker.razor.cs
```diff
@@ -249,6 +251,35 @@ protected override void OnInitialized()
...
+        /// <value>The Maximum Selectable Date.</value>
+        [Parameter]
+        public DateTime? MaximumDate { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Initial Date/Month View.
+        /// </summary>
+        /// <value>The Initial Date/Month View.</value>
+        [Parameter]
+        public DateTime? InitialMonth { get; set; }
```
enchev, line 273: The name should include date in my option since it's DateTime

Radzen.Blazor/RadzenDatePicker.razor.cs
andrewaggb: MinDate and MaxDate or just Min/Max?

enchev: Just Min/Max are good enough in my opinion - will match other Radzen components with min/max settings.

Radzen.Blazor/RadzenDatePicker.razor.cs
enchev: InitialViewDate is OK!

enchev:
> rz-state-other-month

Yep, this will be better! Maybe just `rz-datepicker-other-month`. Since the `span` is child of the `td` the span style can be easily overwritten if you set custom CSS class to the `td`.

andrewaggb:
I've updated this with the changes discussed.  Renamed parameters to Min/Max/InitialViewDate, removed SameStyleForAllDates, added rz-datepicker-other-month.  I'm not sure where the 'Source' tab in the demo comes from.

PR #475, opened 2022-06-02
New waiting component

GioeEng:
New component for blocking a user waiting on a background run process with simplified startup.
Directly customizable block layout for your project in the main layout.

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor
+++ b/Radzen.Blazor/RadzenDatePicker.razor
@@ -109,19 +109,19 @@
                     <div class="rz-timepicker">
                         <RadzenNumeric TValue="int" Disabled="@Disabled" Value="@(HourFormat == "12" ? ((CurrentDate.Hour + 11) % 12) + 1 : CurrentDate.Hour)" 
                            Min="@(HourFormat == "12" ? 1 : -1)" Max="@(HourFormat == "12" ? 12 : 24)" TValue="double" Step="@HoursStep"
-                           Change="@UpdateHour" class="rz-hour-picker" @oninput=@OnUpdateHourInput />
+                           Change="@UpdateHour" class="rz-hour-picker" @oninput=@OnUpdateHourInput Format="@(PadHours ? "00" : "")" />
                         <div class="rz-separator">
                             <span>:</span>
                         </div>
                         <RadzenNumeric TValue="int" Disabled="@Disabled" Value="CurrentDate.Minute" TValue="double" Step="@MinutesStep" Min="0" Max="59"
-                           Change="@UpdateMinutes" class="rz-minute-picker" @oninput=@OnUpdateHourMinutes />
+                           Change="@UpdateMinutes" class="rz-minute-picker" @oninput=@OnUpdateHourMinutes Format="@(PadMinutes ? "00" : "")" />
                         @if (ShowSeconds)
                         {
                             <div class="rz-separator">
                                 <span>:</span>
                             </div>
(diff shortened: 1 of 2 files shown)


akorchev:
Hi @GioeEng,

Thank you for opening a pull request and your effort! 

Here is my high level feedback:

1. I am not sure if I like the idea of the WaitingService running the background task (who makes the syncronisation and error handling?). I think this should be the developer's responsibility and the WaitingService should just have simpler methods Open/Close similar to how the other Radzen.Blazor services work. I imagine something like:
    ```
    waitingService.Open(/* options */);
    await Thread.Sleep(5000);
    waitingService.Close();
    ```
2. We prefer not to use System.Drawing.Color as a property. Component styling is ideally done via the theme CSS.
3. It seems the default implementation is mostly look-less and does not display anything. You have added loading animations in the examples wwwroot directory. Ideally there should be some default implementation in the theme so developers don't have to use custom CSS.

GioeEng:
Hi akorchev,

Thanks for the great tips. My commit is actually a proposal for integrating the waiting control.

I implemented it all within the last commit.

1) In fact, OpenAsync handled with a background task did not allow a proper error trap. I changed the method by launching an action and updated the sample page with the various solutions. Now you can launch the waiting with the classic Radzen Open and Close method or with the OpenAsync method (which I personally find very convenient).
2) I changed the type of MaskColor option variable to string
3) I removed loaders.css from the demo project and implemented the animated loader’s CSS directly into the scss by adding several customization variables. Now the default colors are the same used in the other Radzen controls and change depending on the theme used. However, it is always possible to customize the waiting of the application directly in MainLayout.

Personally (often working with data processing pages that require a few seconds for processing) I find the control really convenient and very functional.

akorchev:
The OpenAsync method still executes the task (as well as Render). We won't accept such API. The current implementation seems to be a bit too heavy for what a loading indicator should be - for example it has almost the same amount of JS required as the HTML editor. We would accept a waiting service which is a lot leaner - with less configuration options and simpler API:

- The OpenAsync method which shows a loading indicator without running a task (with options).
- Background color and z-index should be controlled via single option called Style as with other Radzen components.
- JavaScript should be minimal of any. Check the internal Popup component in the Radzen.Blazor.Rendering namespace. It should provide some of the required implementation.
- WaitingPlaceholders seems overly complicated. Not needed IMHO.

GioeEng:
The process has been simplified to the recommended specifications. The javascript has been reduced to the maximum, in case you would like to simplify it further, you must do without the timer which appears in the waiting.

akorchev:
More comments/questions:

1. What is the purpose of FullCustomContent? I don't think it is needed - the Render method should probably take care of that internally.
2. The JS creates a timer and updates some HTML. We can scrap that and handle it via custom code similar to the other customizations.
3. We will have to design our own animation as we don't want to use someone else's. We can handle that internally but won't happen any time soon due to other priorities.

About AllowClear - unfortunately it won't work as it is. I think positioning the clear icon needs a wrapper element to be done properly (which however is kind of a breaking change as custom attributes will start to render on top of it).

Right now it doesn't work if one changes the height of the textbox:

image

Fortunately AllowClear can be achieved via the built-in type="search":

```
<RadzenTextBox type="search" Change=@(args => OnChange(args, "TextBox")) Class="w-100" />
```

Maybe we can even style it to match the Radzen appearance: https://stackoverflow.com/questions/20804016/editing-input-type-search-pseudo-element-button-x

Anyway we can probably take this pull request from here and work on it further when time permits. We don't want to waste more of your time! Thank you for the great idea and the effort you put into it!

Issue #477, opened 2022-06-03
Make TagName of HeadlineTitle configurable


cheinrich01833:
To improve the accessibility it is necessary to mark the heading of a modal window with a <h...> tag.

akorchev:
Closing as it has been handled via #492

PR #478, opened 2022-06-03
Make TagName of HeadlineTitle configurable

cheinrich01833:
To improve the accessibility it is necessary to mark the heading of a modal window with a <h...> tag.
#477

### DIFF
--- a/Radzen.Blazor/DialogService.cs
+++ b/Radzen.Blazor/DialogService.cs
@@ -192,7 +192,6 @@ namespace Radzen
                 Bottom = options != null && !string.IsNullOrEmpty(options.Bottom) ? options.Bottom : "",
                 Height = options != null && !string.IsNullOrEmpty(options.Height) ? options.Height : "",
                 ShowTitle = options != null ? options.ShowTitle : true,
-                TitleTagName = options != null && !string.IsNullOrEmpty(options.TitleTagName) ? options.TitleTagName : "span",
                 ShowClose = options != null ? options.ShowClose : true,
                 Resizable = options != null ? options.Resizable : false,
                 Draggable = options != null ? options.Draggable : false,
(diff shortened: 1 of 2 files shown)


akorchev:
I think we can just use a `<h3>` or `<h4>` in the .razor file without an extra option. I can imagine 99.99% of the users won't need to change it ever :)

cheinrich01833:
We are using nested RadzenTabs. Depending of the level of nesting we need different h-tags `<h3>` up to `<h6>`.  To avoid a breaking change, I think, it would be better to use a option. 99.99% of users can ignore this option but the remaining would be free. ;-)

akorchev:
I am afraid we can't accept the pull request like it is. This requirement is too niche to add as an extra dialog option.

cheinrich01833:
As software developer I agree with you, but the colleagues of our test department (Deutsche Telekom AG) are of the opinion, that using of right header html tags in the right order without gaps are not a niche case... ;-(
Your suggestion is to change fix the `<span>` tag to `<h4>` ? 
Are there any other ideas to achieve this goal?

akorchev:
> Are there any other ideas to achieve this goal?

I would just add the necessary heading in the dialog content itself. You can optionally hide the title of the dialog via the API.

akorchev:
@cheinrich01833 did you try my suggestion to include the title markup in the dialog content itself?

cheinrich01833:
Yes I did, but it is not the best idea. You loose the function of the close button in the title bar and you must change the css because of padding in .rz-dialog-titlebar and .rz-dialog-content. We will not do it that way.

akorchev:
At the moment the Title is rendered as markup string. This means it accepts HTML and you can use `<h3>My title</h3>` when you open the dialog.

```
  await DialogService.OpenAsync<DialogCardPage>($"<h3 style='display:inline'>Order {orderID}</h3>",
               new Dictionary<string, object>() { { "OrderID", orderID } },
               new DialogOptions() { Width = "700px", Height = "570px", Resizable = true, Draggable = true });
```

image

akorchev:
We would accept a pull request changing the span to a div. Then it would be valid I think.

cheinrich01833:
Thats great. Thanks. I have changed the code.

akorchev:
Could please rebase your commits so only the latter one goes in git history? Alternatively close this pull request and open a new one with just the last one.

PR #487, opened 2022-06-13
LegendClick event in PieSeries

paulo-rico:
While using a donut chart as both information and as a navigation aid to drill down to a segment's detail, I noticed that with a large variance (e.g. 600 to 1) in the chart data, the clickable area is too small to action. I therefore created a LegendClick event to address this using almost identical code that exists for the SeriesClick Event.

Below is a partial file of the changes made. I will place them in RadzenPieSeries.razor.cs

`using Microsoft.AspNetCore.Components;
using Radzen.Blazor.Rendering;
using System.Threading.Tasks;

namespace Radzen.Blazor
{
  /// <summary>
  /// Renders donut series in <see cref="RadzenChart" />.
  /// </summary>
  /// <typeparam name="TItem">The type of the series data item.</typeparam>
  public partial class RadzenDonutSeries<TItem> : RadzenPieSeries<TItem>, IChartDonutSeries
  {
    /// <summary>
    /// A callback that will be invoked when the user clicks on a legend item. 
    /// </summary>
    [Parameter]
    public EventCallback<SeriesClickEventArgs> LegendClick { get; set; }
    /// <inheritdoc/>
    public override RenderFragment RenderLegendItem()
    {
      return builder =>
      {
        foreach (var data in Items)
        {
          builder.OpenComponent<LegendItem>(0);
          builder.AddAttribute(1, nameof(LegendItem.Text), TooltipTitle(data));
          builder.AddAttribute(2, nameof(LegendItem.Class), $"rz-series-item-{Items.IndexOf(data)}");
          builder.AddAttribute(3, nameof(LegendItem.MarkerSize), MarkerSize);
          builder.AddAttribute(4, nameof(LegendItem.MarkerType), MarkerType);
          builder.AddAttribute(5, nameof(LegendItem.Color), PickColor(Items.IndexOf(data), Fills));
          builder.AddAttribute(6, nameof(LegendItem.Click), EventCallback.Factory.Create(this, () => OnLegendItemClick(data)));
          builder.CloseComponent();
        };
      };
    }
    /// <inheritdoc />
    private async Task OnLegendItemClick(object data)
    {
      if(LegendClick.HasDelegate)
      {
        await InvokeLegendClick(LegendClick, data);
      }
    }

    /// <inheritdoc />
    public async Task InvokeLegendClick(EventCallback<SeriesClickEventArgs> handler, object data)
    {
      var category = Category(Chart.CategoryScale);

      await handler.InvokeAsync(new SeriesClickEventArgs
      {
[...]

PR #488, opened 2022-06-13
Call Chart.SeriesClick when legend is clicked

paulo-rico:
Overrides the following Pull Request made in error
[https://github.com/radzenhq/radzen-blazor/pull/487#issuecomment-1153591621](https://github.com/radzenhq/radzen-blazor/pull/487#issuecomment-1153591621)

If a pie or donut chart is created and the variance between data items is large, i.e. Series1: 600, Series2: 1, Series2 becomes difficult/impossible to 'SeriesClick'. Therefore, saw a requirement to enable that same click through the legend.

Moved Event to Chart rather than Series as advised. But the new Event could be cumbersome as it doesn't apply to all charts. There's already an action for LegendItemClick on some charts. So just made the call to SeriesClick instead.

### DIFF
--- a/Radzen.Blazor/RadzenPieSeries.razor.cs
+++ b/Radzen.Blazor/RadzenPieSeries.razor.cs
@@ -3,6 +3,7 @@ using Radzen.Blazor.Rendering;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Radzen.Blazor
 {
@@ -116,13 +117,38 @@ namespace Radzen.Blazor
                     builder.AddAttribute(3, nameof(LegendItem.MarkerSize), MarkerSize);
                     builder.AddAttribute(4, nameof(LegendItem.MarkerType), MarkerType);
                     builder.AddAttribute(5, nameof(LegendItem.Color), PickColor(Items.IndexOf(data), Fills));
+                    builder.AddAttribute(6, nameof(LegendItem.Click), EventCallback.Factory.Create(this, () => OnLegendClick(data)));
+
                     builder.CloseComponent();
                 };
             };
         }
 
-        /// <inheritdoc />
-        public override bool Contains(double x, double y, double tolerance)
+      private async Task OnLegendClick(object data)
+      {
+        if(Chart.SeriesClick.HasDelegate)
+        {
+          var category = Category(Chart.CategoryScale);
+
+          var args = new SeriesClickEventArgs
+          {
+            Data = data,
+            Title = GetTitle(),
+            Category = PropertyAccess.GetValue(data, CategoryProperty),
+            Value = PropertyAccess.GetValue(data, ValueProperty),
+            Point = new SeriesPoint
+            {
+              Category = category((TItem)data),
+              Value = Value((TItem)data)
+            }
+          };
+
+          await Chart.SeriesClick.InvokeAsync(args);
+        }
+      }
+
+      /// <inheritdoc />
+      public override bool Contains(double x, double y, double tolerance)
         {
             if (Items.Any())
             {


akorchev:
I think a separate event is needed as SeriesClick is fired when clicking the series itself. Also it would be awkward if only the legend of pie and donut series triggers the event.

paulo-rico:
Hi akorchev,

Not sure where to go with this now. 

The behaviour of clicking a legend is already different between chart types. RadzenPieSeries overrides to not include a LegendItem.Click. The existing function for the other chart types is to make that series invisible.

Does it call for another property on a chart to set the behaviour of a legend click? If so, how would the existing behaviour work with a pie chart?

Stay well

Paul

akorchev:
Pie and donut series render a label per value instead of per series hence the difference in the behavior.

paulo-rico:
"...it would be awkward if only the legend of pie and donut series triggers the event"

I think this is where the confusion is. Currently, LegendItem.Click only applies to non-pie charts. It's not exposed to set a delegate, it just happens on those charts (functionality: hiding a series).

It wouldn't make sense to piggy-back on that functionality and hide the series, then call a delegate.

One thought (in the Legend.Click code that hides the series) is to check if there is a delegate and then ignore the default code to hide the series and proceed to call the delegate. That means that if a LegendClick delegate is set, all chart types will call the delegate. Otherwise, the default behaviour for non-pie-type charts will action and hide the series.

Does this sound acceptable?

akorchev:
> It wouldn't make sense to piggy-back on that functionality and hide the series, then call a delegate.

Why though? Sounds fine to me to call the event handler after the default behavior.

paulo-rico:
The way I'm using it is to drill down through the data.
So the series of events would be -
1. Click the legend
2. Automatically hide the series based on chart (currently out of a developers control)
3. Show the detail of the series that you've in effect just hidden

It's not really intuitive.

akorchev:
Hiding the labels when clicking the legend items is a different issue which isn't related. It can be handled via separate setting altogether. To be honest I don't think there is an API to hide series in the same way as from the legend.

paulo-rico:
I think we may be getting our wires crossed somewhere along the line.
I've created my own website with the Radzen Demo's on it. I will adjust the demos to accommodate the LegendClick to show you what I mean. . I'll be in touch when I've done this, it should hopefully demonstrate what I'm doing and what my concerns are. I'll post the link when finished.

akorchev:
Just to summarize my feedback so far:
- We will not accept a pull request which triggers SeriesClick when a legend label is clicked. 
- We don't want series to have events of their own.
- We don't want to have events that only work when certain series are used.

paulo-rico:
[http://radzen.delaci.co.uk/bar-chart](http://radzen.delaci.co.uk/bar-chart) (and the other charts).
I've changed all the chart demos to show a console and log the events.
My initial concern was in hiding a series, but then calling LegendClick, which in my application shows a detail of the series. It didn't make sense to hide, but then refresh a linked list showing the detail of a series you've just hidden. 
Resolution - adding the IsVisible property to the LegendClickEventArgs so the developer can then choose what to do.

akorchev:
Adding IsVisible to the event args sounds reasonable. Unfortunately there are some conflicts that have to be resolved. Also the Common.cs file has too many changes - what did you do to it? Lastly - please format the new code according to the rest of the source. Right now it uses two spaces for indentation whereas we use four.

paulo-rico:
Please check if that is now ok. 
!!!   Common.cs - I only added the LegendClickEventArgs class   !!! Can't see how it messed up so much.
I reset back to branch and applied changes (after changing default tab spacing to 4).
Checked all files changed in forked repository and the look just fine.

akorchev:
There are still conflicts and Common.cs is all changed. Visible in the `Files changed` tab. You may want to rebase your branch to the current master, or get the contents of Common.cs from the master and add the LegendClickeventArgs again. Or start the PR from scratch. The are also changes to Radzen.Blazor.csproj which are not needed.

paulo-rico:
Seems the reset didn't work. Looked fine in diff in Visual Studio but obviously not.
I will delete my fork and start again. It must work this time. I think I'm only adding about 20 lines over four files. I will not commit until I'm 100% sure that the changes are all that you see.
Appreciate you bearing with me. As I may have mentioned before, first contribution to an open source project. Already learnt a lot of what and what not to do.

PR #491, opened 2022-06-14
Placeholders for htmleditorcomponents

rebnis-johan:
This PR adds possibility to set the Placeholder-texts for the sub components to the RadzenHtmlEditor, specifically RadzenHtmlEditorFontName, RadzenHtmlEditorFontSize and RadzenHtmlEditorFormatBlock. As all titles/tooltips colud already be set, this was the only(?) thing missing to make it possible to change language for the html-editor.


akorchev:
Thanks!

Issue #410, opened 2022-03-31
Tabs - Tab Position - Allow Top Right


DanRogers99:
Allow TabPosition.TopRight to TabPosition to align tabs in the top right-hand corner.

![image](https://user-images.githubusercontent.com/33861884/161101169-26a6acee-a3ab-4830-9320-83dd442d1a5c.png)

Can be achieved by adding 
`justify-content: flex-end;` to .rz-tabview-nav

PR #493, opened 2022-06-14
Add TopRight and BottomRight position to TabPosition

thedomingo:
This PR adds `TopRight` and `BottomRight` position to `TabPosition` and implements issue #410.

`TopRight` example
image

`BottomRight` example
image

### DIFF
--- a/Radzen.Blazor/RadzenTabs.razor.cs
+++ b/Radzen.Blazor/RadzenTabs.razor.cs
@@ -181,11 +181,11 @@ namespace Radzen.Blazor
             }
             else if(TabPosition == TabPosition.TopRight)
             {
-                positionCSS = "rz-tabview-top-right";
+                positionCSS = "rz-tabview-top rz-tabview-top-right";
             }
             else if (TabPosition == TabPosition.BottomRight)
             {
-                positionCSS = "rz-tabview-bottom-right";
+                positionCSS = "rz-tabview-bottom rz-tabview-bottom-right";
             }
 
             return $"rz-tabview {positionCSS}";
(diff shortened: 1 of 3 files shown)


Radzen.Blazor/RadzenTabs.razor.cs
```diff
@@ -179,6 +179,14 @@ protected override string GetComponentCssClass()
             {
                 positionCSS = "rz-tabview-left";
             }
+            else if(TabPosition == TabPosition.TopRight)
+            {
+                positionCSS = "rz-tabview-top-right";
```
yordanov, line 184: Let's add a CSS class instead of replacing it to reduce the duplicated code in the SCSS file. In this case it would be `positionCSS = "rz-tabview-top rz-tabview-top-right";`

Radzen.Blazor/themes/components/blazor/_tabs.scss
```diff
@@ -17,193 +17,263 @@ $tabs-icon-margin: 0 !default;
...
+    border: $tabs-border;
+    box-shadow: $tabs-shadow;
+    flex: 1;
+    overflow: auto;
+
+    .rz-tabview-top > & {
+        border-radius: 0 0 $border-radius $border-radius;
+    }
+
+    .rz-tabview-top-right > & {
```
yordanov, line 258: We don't need this rule as it will be inherited from the `.rz-tabview-top` rule above. Same applies to the `bottom-right` rule below.

yordanov (changes requested):
Thank you for this PR! I have a suggestion - instead of replacing the `rz-tabview-top` CSS class with `rz-tabview-top-right`, why don't we just append the latter. This way we avoid code duplication in the SCSS file. See my comments.

thedomingo:
Thank you for your feedback. I haveve updated this with the requested changes.

yordanov (approved):

PR #499, opened 2022-06-20
LegendClick after fork recreate 202206181900

paulo-rico:
Absolutely checked that it only has the few code changes that I've applied! 57 changes in four files.


akorchev:
Thanks!

PR #509, opened 2022-07-05
Chart overlay to draw a Constant, Mean, Median and Trend

glutio:
Added a feature to display chart overlays. Implemented a Constant, Mean, Median and Trend (linear regression).

image

```
<RadzenChart>
    <RadzenBarSeries Data="Data" CategoryProperty="Key" ValueProperty="Value">
        <RadzenSeriesConstantLine TItem="KeyValuePair<string,double>" Value="30" StrokeWidth="2" Stroke="red" LineType="LineType.Solid" />
        <RadzenSeriesMeanLine TItem="KeyValuePair<string,double>" @bind-Value="Mean" StrokeWidth="2" Stroke="green" LineType="LineType.Dotted" />
        <RadzenSeriesMedianLine TItem="KeyValuePair<string,double>" @bind-Value="Median" StrokeWidth="2" Stroke="yellow" LineType="LineType.Dashed" />
        <RadzenSeriesTrendLine TItem="KeyValuePair<string,double>" StrokeWidth="2" Stroke="pink" LineType="LineType.Solid" />
    </RadzenBarSeries>
    <RadzenValueAxis>
        <RadzenGridLines Visible="true"></RadzenGridLines>
    </RadzenValueAxis>
</RadzenChart>
<p>Mean: @Mean</p>
<p>Median: @Median</p>
```


akorchev:
Hi @glutio,

Thank you for the pull request! It would really be a great addition to the Radzen.Blazor chart. 

I have a few comments over the API and implementation that I would like to discuss with you.

I think the API and implementation can be made simpler if you use `IChartSeries` instead of `CartesianSeries<TItem>` - you can add the methods that compute the median, mean and trend as methods of `IChartSeries`. This would achieve two things - remove the need of TItem in the *Line components and also remove the need of Value/ValueChanged (if one wants the median, mean or trend as a number they can get a `@ref` to the series and call the corresponding methods).

glutio:
Hi, I refactored the code based on your comments and added another overlay component - text label - which uses CartesianSeries<TItem> but the type is inferred from Data parameter. One thing about using @ref vs @bind-Value 
```
<Series @ref="refSeries"/>
@refSeries.Median
```
It looks like refSeries is not initialized when it first renders the above markup.

I have one concern about SVG, it looks like the overlays are drawn multiple times, I can see in DevTools that it just appends to the svg. I thought @key supposed to fix it but apparently not. Do you know what this is and how to fix it?

image

glutio:
Thanks for the hint! I fixed the SVG issue - I now check if the instance is already in the Overlays list before adding it. Maybe make Overlays a hashset instead of a list?.. Also changed Properties to Methods per your comment.

image
image

glutio:
Added tooltip for overlays to display mean/median value

image

akorchev:
I think the data labels should be enabled per series without the need of a loop (it should be internal).

```
<RadzenSeriesDataLabel Stroke="red" OffsetX="10" OffsetY="10" />
```

Having the loop makes it cumbersome to use.

In addition there should be some offset (horizontal or vertical) so the text of the label does not appear over the series.

Great job by the way!

glutio:
The idea about the DataLabel is that it is a universal component for displaying any text (it uses your SVG Text component that does not have stroke but has takes class), and it already has offsetX and offsetY and css class parameters, and the loop is just an example of how this universal label component can be used to display values. I think it may be useful to have a component like this to display any text, not just values.. But I will also add some component that adds values labels without the need for a loop but since your svg Text component takes css class instead of stroke I will use a class parameter.

akorchev:
In this case you can create a separate component RadzenSeriesLabel (which can display arbitrary text) and RadzenSeriesDataLabels (which display the data labels). Adding Stroke, StrokeWidth and other properties can be done by us. By the way let me know if you feel you can't spend any more time on this - we can take it from here. It would just take some time to finish it up.

glutio:
Thinking about it I will need to add Data to IChartSeries or the labels component will need TItem, which do you prefer?

akorchev:
Adding Data to IChartSeries sounds better.

glutio:
Ok here's the implementation for data labels. It supports offsetX in polar coordinates too but the labes are clipped if they go out of chart bounds.. I tried using padding on the chart to add more space but that didn't work nice.
image

glutio:
I think it is more or less a complete feature now so if you want to take over and merge please go ahead, thank you!

akorchev:
Hi @glutio,

Could you please rebase your branch as currently there are merge conflicts? We will take it from there then.

glutio:
Done

akorchev:
I don't know how to use other email. We merge pull requests from the github UI and I am not aware of such an option.

glutio:
Ok no problem

glutio:
Btw, I think there’s some design bug in the charts where in case of the bar series I need to put my value formatter on the category axis. For user it should be value axis whether inverted or not but in my testing I concluded that in inverted axes scenario I need to use category axis formatter. Did I misunderstand that? You can see I do it in the data labels implementation. Thanks!

akorchev:
Our [bar chart demo](https://blazor.radzen.com/bar-chart) seems to work fine - the ValueAxis is formatted. 

```
                    <RadzenValueAxis Formatter="@FormatAsUSD">
                        <RadzenAxisTitle Text="Revenue in USD" />
                    </RadzenValueAxis>
```

glutio:
Ok, fixed. thanks

PR #516, opened 2022-07-08
Alternative simple busy/loading wrapper 

glutio:
One issue is that the spinner is at the center of the control, so if you scroll the control the spinner will scroll with it. Maybe it's a feature.

image

https://user-images.githubusercontent.com/22550674/178044310-4b118ea6-221d-43f7-98d0-bea7b3f8cafa.mp4

### DIFF
--- a/Radzen.Blazor/BusyLoadingService.cs
+++ b/Radzen.Blazor/BusyLoadingService.cs
@@ -1,0 +1,22 @@
+﻿namespace Radzen.Blazor
+{
+    /// <summary>
+    /// A service to control a RadzenBusyLoading component
+    /// </summary>
+    public class BusyLoadingService
+    {
+        internal delegate void StateChanged(bool val);
+        internal event StateChanged BusyChanged;
+        internal event StateChanged LoadedChanged;
+
+        public void Busy(bool val)
+        {
+            BusyChanged.Invoke(val);
+        }
+
+        public void Loaded(bool val)
+        {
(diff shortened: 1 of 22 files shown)


akorchev:
Hi @glutio,

There is a similar [pull request](https://github.com/radzenhq/radzen-blazor/pull/475) which we are still not ready to merge as it requires UX design. We will keep both open to evaluate the pros and cons.

glutio:
sure, no worries, just offering this simpler no JS alternative because I already had it available from my other project..

glutio:
Added fullscreen mode 

image

image

glutio:
Now supports three usage options:

Using @ref directly 
Using CascadingParameter passed to the child content
Using scoped service BusyLoadingService which may be shared by multiple wrappers

glutio:
BusyLoadingService usage

image

image

image

akorchev:
After adding the the service this pull request became very similar to the one opened earlier. This is why we can't merge them both. TBH the other one is more inline with the existing Radzen.Blazor API and services and we will probably merge it after it receives the required UX design.

Please discuss with us via issues, forum threads or email any future work like to contribute. This will prevent conflicts like this in the future and won't lead to any waisted efforts.

glutio:
Ok, no worries, although, frankly, I think mine's better ;) Cheers!

https://user-images.githubusercontent.com/22550674/178134236-5ae13196-f4f1-49bc-a796-5bc78fe9dd9f.mp4

PR #531, opened 2022-07-14
Do not show trend tooltip if the line is not visible and fix line names

glutio:
- Do not show trend tooltip if the line is not visible and fix line names
- Revert accidental removal of commit 6b9d2c19a7f8e5ac36b4b528e912e6bd84264a7f (FilterCleared event)


glutio:
Nice work putting a demo together! I think it would also be useful to demo the ConstantLine, maybe have two to show some kind of corridor on the chart. Also, I noticed some chart width calculation issue in the lines. Is it by design or does it need to be fixed?
image

PR #534, opened 2022-07-20
add DisableEmptyText attribute

pigwing:
add DisableEmptyText attribute fix when AllowPaging = true && LoadData.HasDelegate && Count == 0 DataGrid InLine Editing not working

https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenDataGrid.razor#L275
because DataGrid InLine Editing template radzenGrid.InsertRow  must render DrawRows(visibleColumns) fontend,


enchev:
Hi @pigwing,

Maybe ShowEmptyMessage (with true as default value) will be better name since there is EmptyTemplate and EmptyText properties. Please add the property also to the RadzenDataGrid component only not in PagedDataBoundComponent.

PR #538, opened 2022-07-20
Add control over url matching for RadzenMenuItem

iliyan-kulishev:


akorchev:
Hi @iliyan-kulishev,

Thank you for the pull request! We have a comment though. In RadzenPanelMenu this functionality is [implemented](https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenPanelMenu.razor.cs#L34) as a "top-level" property and is set once per component and not on every item. For consistency we think it should be implemented the same way in RadzenMenu.

iliyan-kulishev:
Thanks for the reply. I didn't know about this. For now I'm using RadzenPanelMenu in my projects.

PR #548, opened 2022-07-26
Extract Interfaces for Radzen Services

### LINKED ISSUE #547: Components, relying on Radzen Service classes are not unit testable
I am using Radzen in a WASM project, trying to implement Unit Tests for my created components.
Therefore I am mocking all Services I inject. Unfortunately the services provided by the Radzen Library are not mockable, because they are not providing interfaces.
The change would be optional and would not break any existing implementations. Nevertheless I think it is binary incompatible (unsure about that).

Extract Interfaces for `NotificationService`, `TooltipService`, `DialogService`, `ContextMenuService`

An alternative solution would be to wrap the services on my own.

MPapst:
extract interfaces and provide a ServiceCollection initializer
See issue for discussion:
fixes #547

### DIFF
--- a/Radzen.Blazor/ContextMenuService.cs
+++ b/Radzen.Blazor/ContextMenuService.cs
@@ -37,7 +37,7 @@ namespace Radzen
     /// }
     /// </code>
     /// </example>
-    public class ContextMenuService : IDisposable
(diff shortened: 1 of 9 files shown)


MPapst:
Damn.. sorry - this needs more work. It breaks the RadzenNotification Component.

PR #550, opened 2022-07-28
RadzenDialog Improvements

BassetMan:
Fixed issue with RadzenDialog where closing a nested dialog removed the 'no-scroll' class from the page body, allowing the background to scroll. The javascript 'closeDialog' function is now only invoked when the last dialog is closed.

Added a  potential improvement to RadzenDialog, whereby the dialog mask is only shown for the top most dialog, rather than the accumulative effect of multiple masks. This might be personal preference, but makes more sense to me.

### DIFF
--- a/Radzen.Blazor/PagedDataBoundComponent.cs
+++ b/Radzen.Blazor/PagedDataBoundComponent.cs
@@ -303,15 +303,12 @@ namespace Radzen
         /// <param name="args">The <see cref="PagerEventArgs"/> instance containing the event data.</param>
         protected async Task OnPageChanged(PagerEventArgs args)
         {
-            if (skip != args.Skip || CurrentPage != args.PageIndex)
-            {
-                skip = args.Skip;
-                CurrentPage = args.PageIndex;
+            skip = args.Skip;
+            CurrentPage = args.PageIndex;
 
-                await Page.InvokeAsync(args);
+            await Page.InvokeAsync(args);
 
-                await InvokeAsync(Reload);
-            }
+            await InvokeAsync(Reload);
         }
 
         int? pageSize;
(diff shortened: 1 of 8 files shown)


enchev:
Hi @BassetMan,

Unfortunately I was unable to reproduce the issue you've described. Here is how I tested:
image
image

Closing last opened dialog will not background to scroll. Any idea how to reproduce the issue in our demos?

BassetMan:
Hey @enchev,

Thanks for getting back to me. I just tried to reproduce the issue with the demos, and found that because the layout page does not scroll the whole page, the issue does not occur. The easiest way I found to replicate the issue without too much effort is to create a new layout where the whole page scrolls. In the demos project I created a new layout called **EmptyLayout.razor**.

```
@using Microsoft.AspNetCore.Components
@using Microsoft.AspNetCore.Http
@using Radzen.Blazor

@inherits LayoutComponentBase

<RadzenDialog />

<div style="height:5000px;">
    @Body
</div>
```

Then I modified the **DialogPage.razor** to use this new layout. 

```
@page "/dialog"
@inject DialogService DialogService
@implements IDisposable
@layout EmptyLayout

<RadzenExample Name="Dialog" AdditionalSourceCodePages=@(new [] { "DialogCardPage.razor" })>
    <div class="row px-3">
        <div class="col-md-4 p-3">
            <RadzenCard>
            ...
```

Opening the order works as expected, then opening the nested dialog also works as expected. Closing the nested dialog removes the no-scroll class from the page body. Leaving the background content scrollable, while the initial dialog is still presented.

image

Also, did you see my proposal to only show the mask for the top most dialog, rather than having multiple masks appear when one or more nested dialogs are displayed. Which almost completely obscures the background content.

Issue #537, opened 2022-07-20
DropDown does not throw errors within an async Task set to the Change parameter.


Daeymon:
**Describe the bug**
Errors within an async method set to the Change EventCallback parameter of the RadzenDropDown do not get thrown back to main thread and therefore do not go through the logger.

**To Reproduce**
1. Create a default Blazor Server Project with Radzen added as per the Get Started instructions.
2. Add the following code to the bottom of Index.razor:

```
<span>This component correctly throws an error in the async OnChange method.</span>
<RadzenNumeric TValue=int Change=this.OnChangeAsync />

<span>This component does not correctly throw the error in the async OnChange method.</span>
<RadzenDropDown TValue=string Data=this.Values Change=this.OnChangeAsync />

@code {
    public string[] Values = { "Yes", "No" };

    public async Task OnChangeAsync()
    {
        throw new InvalidOperationException("This is an error");
        await Task.CompletedTask;
    }
}
```
3. Changing a value on the RadzenNumeric will throw an error as one would expect. This one is there to show expected behaviour. Changing a value on the RadzenDropdown will do nothing within the application. The only evidence of the error will be a line in the debug console.

**Expected behavior**
Expected behaviour is what happens when changing the value in the RadzenNumeric in the above code.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 103.0.5060.114

PR #563, opened 2022-08-04
fix: RadzenDropDownItem behavior on older versions of the blazor.js f…

GioeEng:
Unfortunately in older versions of NET CORE 6 the framework blazor.js does not include the passage in SignalIR of the field "type" (as in the figure appears comparing the version 6.0.1 and 6.0.7). The dropdown, if used with the old version, is no longer selectable with the mouse (type='click' is always false as type is always empty). 

compare


enchev:
Please check if this will not brake again this: https://github.com/radzenhq/radzen-blazor/issues/537

GioeEng:
> Please check if this will not brake again this: #537

I went back to commit a0d49f8 and also in the previous bc66795 to do a check. I don’t seem to be stopping this again: #537. The change works correctly even in case of throw

I saw the merger. 
Ok thanks

PR #568, opened 2022-08-08
feat: Dropdown with badges view

GioeEng:
Proposal:
Dropdown display with synchronized badges with multiple selection.
It’s convenient when the dropdown elements are numerous and you need to delete a selected item.


akorchev:
Hi @GioeEng,

Thank you for your contribution! It will be a valuable addition to the Radzen.Blazor components.

We will make a few changes in the rendering and property name (we prefer pill instead of badge). The changes will be pushed to this branch.

yordanov:
This is how the chips look with the last commit:
image

Styles will be further refined once we introduce a standalone Chip component.

GioeEng:
Perfect! Thanks for the implemantation.

PR #584, opened 2022-08-23
Fix Grid width is wrong in responsive mode

jiaming0708:
The column width should be 100% in responsive mode
- before
image
- after
image

### DIFF
--- a/Radzen.Blazor/themes/components/blazor/_grid.scss
+++ b/Radzen.Blazor/themes/components/blazor/_grid.scss
@@ -663,9 +663,10 @@ $column-drag-handle-hover-color: $column-drag-handle-color !default;
       > table {
         table-layout: auto;
         display: block;
-      }
-      tbody {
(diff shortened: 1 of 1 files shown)


enchev:
Hi @jiaming0708,

Unfortunately I'm unable to reproduce this:
image
image

Anything missing on my end?

jiaming0708:
The `note` content is too long, making the width overflow, check the element width via the inspect
you can remove/mark the `note` column to reproduce this issue

enchev:
Thanks! In this case maybe we should simply move the tbody rules inside the table rules:

```
> table {
    table-layout: auto;
    display: block;

    > tbody {
        display: block;
    }
}
```

jiaming0708:
ok, changed the code, https://github.com/radzenhq/radzen-blazor/pull/584/commits/b4c642caf3f0f43d0f2842babcf08f3013d1fd92

PR #589, opened 2022-08-30
fix ProgressBar template display error

pigwing:
about
Snipaste_2022-08-30_09-32-23


enchev:
Hi @pigwing,

Thanks for the pull request! The problem is already fixed.

PR #618, opened 2022-09-26
Save/Load DataGrid runtime settings added

enchev:


elylv:
This is great, and saves me hacking around with a wrapper to do this myself.  I am experiencing an apparent bug, though #633

PR #622, opened 2022-09-28
[Docs] Fix miscellaneous annoyances in the docs

beaudry:
In the docs website, I saw this and it annoyed me

image

So I adjusted it.


beaudry:
Update: saw this and it annoyed me as well.

image

beaudry:
I guess this is something as well, but I don't know if it should be removed, or have something inside? 
image

PR #690, opened 2022-11-09
New Sidepanel component

MPapst:
Introduces a new Sidepanel component including a service (similar to the dialog).

The Sidebar Component allows to open a single instance of a panel which is displayed on the right side. It renders a given component in the content div and allows some customization through an optional options object. It also allows to pass back values when closed.

If you put me some feedback in here, I will implement that and create tests and documentation afterwards.

### DIFF
--- a/Radzen.Blazor/RadzenSidepanel.razor
+++ b/Radzen.Blazor/RadzenSidepanel.razor
@@ -1,0 +1,74 @@
+﻿@inject SidepanelService _service
+@inherits RadzenComponent
+
+
+@if (Visible && isOpen)
+{
+    <aside @ref="@Element"
+        class="@GetCssClass()"
+        tabindex="@(isOpen ? "0" : "-1")"
+        style="@GetStyle()"
+        aria-labelledby="rz-sidepanel-label"
+        @onmouseenter="@OnMouseEnter"
+        @onmouseleave="@OnMouseLeave">
+        @if (options.ShowTitle)
+        {
+            <div class="rz-sidepanel-titlebar">
+        <div class="rz-sidepanel-title" style="display: inline" id="rz-sidepanel-label">@options.Title</div>
+                @if (options.ShowClose)
+                {
+                    <a href="javascript:void(0)" class="rz-sidepanel-titlebar-close" @onclick="@(_ => _service.Close(null))" role="button">
+                        <span class="rzi rzi-times"></span>
+                    </a>
+                }
+            </div>
+        }
+        <div class="rz-sidepanel-content" style="@options.Style">
+            @ChildContent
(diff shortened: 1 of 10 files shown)


akorchev:
Hi @MPapst,

We discussed your PR in detail and we think that:

1. The service adds complexity - it needs to be registered at startup, the component should be defined in the layout (and not nested anywhere), using it later with blazor content requires inline render fragments which have their quirks. Last but not least we have to add explicit support for such a service in Radzen and Radzen Blazor Studio which is a lot of extra work.
2. We already have a component called RadzenSidebar. It would be confusing to have RadzenSidepanel as well.

Having this in mind we believe this functionality should be a feature of RadzenSidebar - some new behavior that opens it on top of other content (via `position: fixed`). It should probably be a new property (Variant?) and should accept a position where to open at (Top, Left, Bottom, Right) and perhaps render an overlay which prevents other inputs from getting focus and closes the sidebar on click. Hiding and showing the sidebar can happen via the Visible property (similar to how RadzenAlert implements it).

If you like the idea, you are more than welcome to modify your pull request. Otherwise we will have to close it and implement it as a new behavior of RadzenSidebar in the near future.

MPapst:
Hey @akorchev,

sounds like it can work that way. I thought about the sidepanel similar to the dialog or a dialog extension avoiding a lot of dialog madness in applications.

It will not solve all the points you mention, but what whould you think by extending the `RadzenDialog` / `DialogService` with two new methods and the possibility to open the dialog as an side panel instead?
Could also be a property to the DialogOptions (e.g. `DialogPosition = DialogPosition.SideRight` or similar). This would solve all the complexity issues and make it a low easier to implement in the Studio as well, as you wil need another configuration option instead of implementing a completely new dialog variant.

akorchev:
That's another way to solve it. We are however against extending DialogOptions for the following reasons:

1. There are features that won't work in this mode - drag and drop, resizing, custom positioning (top, left)
2. Mobile display should only apply in the current mode (we currently display it on top on mobile devices - a sidebar should stay as it is).

How about new API in DialogService?

```
var result = await dialogService.OpenSidebar<SomePage>(new SidebarOptions { Position=SidebarPosition.Right, Style = "width: 500px" });
```

SidebarOptions will omit the properties of DialogOptions that won't be supported. By using the dialog we would get the overlay functionality for free.

MPapst:
Yea, I like that idea. 
When I've got some time after work today I will change the implementation according to that..

MPapst:
To leave a short update on this task: Unfortunately I haven't had enough time to complete the change before my vacation. I'll be offline for a week now, but will tanke care of the implementation afterwards.

PR #694, opened 2022-11-11
Components - Added id/name-attribute to support more accessiblity

msdevcode:
I added id/name-attribute to some components:

- AutoComplete
- DropDown
- DropDownDatagrid
- Mask
- Numeric
- Password
- TextAera
- TextBox

Im using this components with RadzenLabel and "Component"-attribute as "for"-attribute. The controls couldnt match with the labels "for"-attribute. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attributes
https://www.w3schools.com/tags/att_label_for.asp

Accessibility is a more and more important topic and i tried to add or fix the "id"-attributes. 

grafik

grafik

Im using Chrome and an add-on called "WAVE Evaluation Tool" to analyse it:

grafik

Im not sure if there are more components that are missing right now. 

Your feedback is welcome in any cases.

### DIFF
--- a/Radzen.Blazor/RadzenMask.razor
+++ b/Radzen.Blazor/RadzenMask.razor
@@ -6,6 +6,6 @@
 @if (Visible)
 {
     <input @ref="@Element" disabled="@Disabled" readonly="@ReadOnly" name="@Name" style="@Style" @attributes="Attributes" class="@GetCssClass()" tabindex="@(Disabled ? "-1" : $"{TabIndex}")"
-           placeholder="@Placeholder" maxlength="@MaxLength" autocomplete="@(AutoComplete ? "on" : "off")" value="@Value" @onchange="@OnChange" id="@GetId()"
-           oninput="Radzen.mask('@GetId()', '@Mask', '@Pattern', '@CharacterPattern')"/>
+           placeholder="@Placeholder" maxlength="@MaxLength" autocomplete="@(AutoComplete ? "on" : "off")" value="@Value" @onchange="@OnChange" id="@Name"
+           @oninput="@OnInput"/>
 }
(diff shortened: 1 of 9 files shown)


Radzen.Blazor/RadzenMask.razor
```diff
@@ -6,6 +6,6 @@
 @if (Visible)
 {
     <input @ref="@Element" disabled="@Disabled" readonly="@ReadOnly" name="@Name" style="@Style" @attributes="Attributes" class="@GetCssClass()" tabindex="@(Disabled ? "-1" : $"{TabIndex}")"
-           placeholder="@Placeholder" maxlength="@MaxLength" autocomplete="@(AutoComplete ? "on" : "off")" value="@Value" @onchange="@OnChange" id="@GetId()"
-           oninput="Radzen.mask('@GetId()', '@Mask', '@Pattern', '@CharacterPattern')"/>
+           placeholder="@Placeholder" maxlength="@MaxLength" autocomplete="@(AutoComplete ? "on" : "off")" value="@Value" @onchange="@OnChange" id="@Name"
+           @oninput="@OnInput"/>
```
enchev, line 10: This change will cause interop calls and might introduce performance problems during input.

akorchev:
This is a a breaking change. Right now we ensure the id is unique even if there are duplicate component names (in case someone has more than one form). We also support setting the Id of a component via custom attribute - your PR seems to disable that - Name will override it.

`<RadzenTextBox id="someid" />`

The id attribute indeed does not set the id of the input element for say RadzenAutoComplete or RadzenDropDown as they are more complex and render wrapper elements. Perhaps a better solution is to introduce a new property for those components called `InputId` which has only one purpose as sets the id of the inner input element. It will be unrelated from Name (which is used to perform validation mostly). What do you think?

akorchev:
The latest change renders the Mask unusable unless Name is set. This is yet another breaking change. We won't accept the pull request in this state. If you want to implement a new parameter InputId let us know.

msdevcode:
Hey @akorchev, thanks for taking the time to explain me the issues of my pull request, appricated. It wasnt my intension to commit breaking changes via a pull request. I fully understand your decision. 

The option of a new attribute sounds fine for me but im not sure if accessibility evaluation tools will accept it. I only find examples where id-attribute of the input component and the for-attribute of the label are matched.

https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML

grafik


Nevertheless i will close the thread.

akorchev:
> Edit: The new "InputId"-attribute would look like this? `<input id="@InputId" />`

Yes.

> Which class would be the right class to include the new attribute? Maybe at the "FormComponent"?

Probably not as just setting the id attribute of some form components should work. Only components that render wrapper elements (e.g. DropDown, Numeric) should need the InputId parameter.

PR #695, opened 2022-11-11
Select value in multiple DropDown on `Enter` press

EkaterinaTikhomirova:
1. It's a bit confusing to select values in drop-down by pressing on `Space` button. 
 I suppose this control should behave like [select2](URL) or another that is similar to `RadzenDropDown`. Most users are used to pressing `Enter` to select an item from a list.
Moreover, there is some unpredictable behavior when drop-down contains values with `Space`.
2. I'd like an option to clear the search in drop-down every time the user selects an item from the list. So, I've add new parameter to allow developers to set up this behavior.

### DIFF
--- a/Radzen.Blazor/DropDownBase.cs
+++ b/Radzen.Blazor/DropDownBase.cs
@@ -598,7 +598,7 @@ namespace Radzen
                     //
                 }
             }
-            else if (Multiple && key == "Space")
+            else if (Multiple && key == "Enter")
             {
                 if (selectedIndex >= 0 && selectedIndex <= items.Count() - 1)
                 {


Radzen.Blazor/DropDownBase.cs
```diff
@@ -598,7 +598,7 @@ private async System.Threading.Tasks.Task HandleKeyPress(Microsoft.AspNetCore.Co
                     //
                 }
             }
-            else if (Multiple && key == "Space")
```
akorchev, line 601: This is a breaking change and we can't accept it. Enter closes the dropdown at the moment and this is why space selects and item.

dmitry-vychikov: Hi @akorchev !

Is there any possibility we could make this change through?

## Using `Space` is not convenient and buggy.

Consider the following scenario:

Your dropdown contains a list of names (first name + last name):

- Dmitry Vychikov
- Dmitry Tikhomirov
- Dmitry ...
- Dmitry ... (a bunch of different Dmitry's)

Then, 
- You start typing `Dmitry`, then see a lot of them
- You move the cursor with arrow buttons to find what you need, but still there are too many of them
- You decide to add a surname to the search, so you start typing again
- You press `<Space>`

But instead of putting a space character in the input, it selects an item! This is very frustrating and not user friendly.

## UI/UX

> Enter closes the dropdown at the moment and this is why space selects and item.

I would say, most users expect that `Esc` button closes the dropdown, not `Enter`.


## Possible options or next steps

- @akorchev are u considering to improve this at all?
- Would you be happy if we made this improvement under say a feature flag or as a configurable parameter? We could leave the default settings in place for now, but provide consumers with the ability to override this behaviour.
    - So it wouldn't be a breaking change anymore

akorchev:
After some internal discussion we decided to change address this. This is how selection will work:

- Enter selects item in single and multiple selection. In single it will also close the popup as this is how `<select>` works.
- In multiple selection mode Esc will close the popup.

Seems to be what you are requesting and would solve the filter issue with values that contain a space.

Could you submit the other feature as a separate PR? Also please don't change unrelated files (you have updated target frameworks in the csproj file).

akorchev:
Please either submit the other feature as a new PR or rebase your branch to drop the unnecessary commits. You should also rebase against the current master branch as you seem to be using an older one - you keep changing the target framework.

EkaterinaTikhomirova:
Done.
[PR for another feature](https://github.com/radzenhq/radzen-blazor/pull/705)

enchev:
Implemented in https://github.com/radzenhq/radzen-blazor/commit/c6a1560a0f23a5a4c95f366d914d70d37a34b5d7

Issue #699, opened 2022-11-12
Support for Speech to Text on TextArea


groogiam:
**Is your feature request related to a problem? Please describe.**
This request is not related to a problem just a nice to have. Mobile users often use this feature on larger text areas that require some sort of narrative.

**Describe the solution you'd like**
Add the ability to initiate browser [speech recognition API](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition)  on a text area.  This would ideally be a new parameter that would enable a microphone icon to initiate the recognition.

**Describe alternatives you've considered**
I have not considered any other alternatives.

**Additional context**
Something like this.

![image](https://user-images.githubusercontent.com/1992395/201493651-744667be-295c-4a6d-807b-f4bb200baf81.png)

I'd be happy to implement and submit a PR for the functionality, I just want to make sure that this functionality is desirable before taking the time. Looking forward to your feedback. Thanks.

kelvinkipruto741-netizen:
give the link

PR #700, opened 2022-11-13
Implemented speech to text for RadzenTextArea radzenhq#699

groogiam:
Implementation for https://github.com/radzenhq/radzen-blazor/issues/699

### DIFF
--- a/Radzen.Blazor.Tests/TextAreaTests.cs
+++ b/Radzen.Blazor.Tests/TextAreaTests.cs
@@ -165,7 +165,7 @@ namespace Radzen.Blazor.Tests
 
             Assert.Contains(@$"autofocus", component.Markup);
         }
-        
+
         [Fact]
         public void TextArea_Raises_ChangedEvent()
         {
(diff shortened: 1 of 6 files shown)


akorchev:
Hi Greg,

We think the button should be a separate stand-alone component which can be used with other input components and not just RadzenTextArea. 

A drawback of your implementation is that the button is always rendered after the text area and there is no way to control its appearance and position via the RadzenTextArea API. Making it a separate component would solve this too.

groogiam:
I'm running into some issues when implementing this. I am trying to pass the reference from the text input to the component that should be updated.

```
<RadzenTextArea @ref="textToSpeechInput" Change=@(args => OnChange(args, "TextArea")) Class="w-100" />
<RadzenSpeechToText AttachedInput="textToSpeechInput" ></RadzenSpeechToText>
```

The problem here is that the @ref does not get populated until after the initial render so the button does not show up until an explicit re-render. Any idea on how to make this work or would another approach be to implement value binding and just have users wire up the data binding to the same property as the textbox? Looking forward to your input. Thanks.

groogiam:
After giving this some thought it seems like the best approach would be to just have the component raise an event call back with the speech results while it is active. This is the simplest implementation and gives the caller the most flexibility as to what they want to do with the results. It also allows for use cases where the user does not want to write the result to a component. Let me know if you agree. Thanks.

groogiam:
Closing in favor of https://github.com/radzenhq/radzen-blazor/pull/703

PR #722, opened 2022-11-28
Add interpolation in RadzenLineSeries to make rendering more generic

maxbruecken:
Add a new property `Interpolation` to the component `RadzenLineSeries` to allow other interpolation modes as actually supported by the property `Smooth` (line and spline). The property `Smooth` is supported furthermore for backward compatibility.


akorchev:
Thank you for the PR @maxbruecken ! Do you think you can also add support for the Interpolation property in RadzenAreaSeries? It is quite similar to RadzenLineSeries.

maxbruecken:
Hi @akorchev, yes sure, I can add support for interpolation in RadzenAreaSeries. Should I extract a base class for both components to avoid code duplication?

akorchev:
> Should I extract a base class for both components to avoid code duplication?

Preferable not for the time being. We can probably reduce the duplication via other means (or just leave it for now).

akorchev:
Decided to create a new demo instead of updating the existing ones. Thank you for the commitment @maxbruecken!

PR #738, opened 2022-12-08
Update RadzenScheduler.razor.cs

igortelheiro:
Currently there's no easy way to get the current selected view. This is because the selectedIndex that gets updated is the field, not the public property.
The Scheduler.IsSelected() is not an easy way to get it, as you need to test every view to get the current one.
Please consider updating the SelectedIndex property or making that SelectedView getter public.

This is the easier way I found to get synced with the current view and, as you can see, not a handful one.
image
image
image
image


akorchev:
Thank you @igortelheiro

PR #739, opened 2022-12-08
Feature: Open a Dialog on the side

MPapst:
superseeds #690

As discussed this implementation is based on the `DialogService` and the `RadzenDialog` component.

### DIFF
--- a/Radzen.Blazor/RadzenDialog.razor
+++ b/Radzen.Blazor/RadzenDialog.razor
@@ -11,7 +11,7 @@
 @if (isSideDialogOpen)
 {
     <aside
-        class="@($"rz-dialog-side {sideDialogOptions.CssClass}")"
+        class="@GetSideDialogCssClass()"
         tabindex="@(isSideDialogOpen ? "0" : "-1")"
         style="@($"{sideDialogOptions.Style}{(sideDialogOptions.Width != null ? " width: {sideDialogOptions.Witdth}" : "")}")"
         aria-labelledby="rz-dialog-side-label"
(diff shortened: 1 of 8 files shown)


Radzen.Blazor/RadzenDialog.razor
```diff
@@ -8,11 +8,51 @@
     <DialogContainer Dialog=@dialog ShowMask=@(dialog==dialogs.LastOrDefault()) />
 }
 
+@if (isSideDialogOpen)
+{
+    <aside
+        class="@($"rz-dialog-side {sideDialogOptions.CssClass}")"
+        tabindex="@(isSideDialogOpen ? "0" : "-1")"
+        style="@($"{sideDialogOptions.Style}{(sideDialogOptions.Width != null ? " width: {sideDialogOptions.Witdth}" : "")}")"
```
MPapst, line 16: Style with optional set Width, which overrides the "hard coded" 400px width in the css

MPapst:
Hi @enchev,

I'll have a look on the `DialogPosition`. Do you think bottom and top makes sense in this case?
Style is already implemented as part of the DialogOptions, see comment above.

enchev:
Thanks @MPapst! In our opinion Top and Bottom most probably will be requested immediately if we merge it with just Left|Right.

MPapst:
Hmm, will try to add them later today.
Those imply a height property over a width... that will bring a way more complex implementation.

MPapst:
@enchev if you don't mind, I fixed the Nullreference Exception introduced with the format replacement in RadzenNumeric

PR #745, opened 2022-12-12
Optimize rendering of chart tooltip

maxbruecken:
This pull request optimizes the rendering of chart tooltip. Only the tooltip will be rendered on mouse movement, it avoids the rerendering of the whole chart and can improve performance of charts with many values.


akorchev:
Hi @maxbruecken,

Could you provide some way to measure the performance of the proposed change?

maxbruecken:
Hi @akorchev,

I have added a synthetical performance test, it take on master few minutes and with optimization ca. 10 seconds.

akorchev:
Thank you @maxbruecken. Looks fine.

PR #761, opened 2023-01-05
Feature: Add a required marker to the label component.

DaveBagler:
This PR adds a new parameter to the `RazenLabel` component named `RequiredMarker`.

This parameter is displayed if it is not an empty string. It appears after the existing `Text` parameter and is span-wrapped (`<span class="required-marker">...</span>`) so that developers can target the required marker with CSS to give it a different presentation than the label text.

The odd line breaks in the razor template are to avoid extra white space. An alternative method to avoid extra whitespace would be to concatenate the `Text` parameter, the span wrapper and the `RequiredMarker` parameter in a class property. However, this method would require the template to output raw HTML, so I felt the odd line breaks were the better approach.

I also added a test for the required marker.

This lets you style form labels like the examples below:

Screen Shot 2023-01-04 at 10 44 21 PM

Screen Shot 2023-01-04 at 10 45 30 PM

### DIFF
--- a/Radzen.Blazor.Tests/LabelTests.cs
+++ b/Radzen.Blazor.Tests/LabelTests.cs
@@ -20,6 +20,22 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void Label_Renders_RequiredMarkerParameter()
+        {
+            using var ctx = new TestContext();
+
+            var component = ctx.RenderComponent<RadzenLabel>();
+
+            var text = "Test";
+            var requiredMarker = "required";
+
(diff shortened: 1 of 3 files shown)


enchev:
Hi @DaveBagler,

Thanks for your pull request however we believe that it will be more generic if we just provide ChildContent for the Label in order to support more scenarios:
https://github.com/radzenhq/radzen-blazor/commit/da442dc02b5249741a2ee7ebf2a1bd2e848aa5b3

DaveBagler:
@enchev, the ChildContent property looks excellent. And that feature's coming in v4.5.0?

PR #766, opened 2023-01-10
Add child content example to label API docs.

DaveBagler:
This change adds an additional example to show the new child content property for the label component.


DaveBagler:
@enchev, thanks for adding the child content property for labels! It is a far more general solution than what I included in https://github.com/radzenhq/radzen-blazor/pull/761, and it works great.

PR #772, opened 2023-01-13
Addition of a YearView for the Scheduler

paulo-rico:
Created a YearView component to be used in the Scheduler ([Year View Demo](http://radzen.delaci.co.uk/)). Although functionally working, there are a couple of issues that I could do with some help on.

1. I am not conversant with SCSS and to be honest, I can't find how the css is generated, so I have placed some styles within the YearView.razor file that need to be addressed. There are also some properties, namely MonthLabelColor, NotInMonthColor and WeekendColor that may or may not need to be css rather than properties.

2. While creating this, it became apparent that there needed to be an option to reduce the height of an appointment, as a year view is sort of more of a summary. Whilst I did change this for the YearView and added the property AppointmentHeight, I was forced to create a new component called YearAppointment that would react to another boolean property ShowAppointmentContent to hide or show the content. Otherwise it would have still drawn the text, there being no Template to override it, and if there was, would have applied to all views. That made me think that the Template should maybe reside at the View level rather than the Scheduler level. This would let you have a different template per view. Just a thought.

### DIFF
--- a/Radzen.Blazor/RadzenYearView.razor
+++ b/Radzen.Blazor/RadzenYearView.razor
@@ -1,0 +1,36 @@
+@using Radzen.Blazor
+@using Radzen.Blazor.Rendering
+
+@inherits SchedulerViewBase
+
+@code {    
+    public override RenderFragment Render()
+    {
+        var appointments = Scheduler.GetAppointmentsInRange(StartDate, EndDate);
+
+        var maxAppointmentsInSlot = 0;
+
+        if (MaxAppointmentsInSlot != null)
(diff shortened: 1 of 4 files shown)


paulo-rico:
I am going to close this Pull Request as it doesn't seem to have been reviewed yet and I have changes to make to resolve item 1 in the first comment.

I will then begin to send Draft Pull Requests so I can really triple check that the code submitted is correct before marking it "Ready for Review". Been burnt once already with an erroneous cut and paste.

PR #773, opened 2023-01-16
Addition of a YearView for Schedule

paulo-rico:
Created a new YearView for the Schedule component. [Test site showcasing YearView](http://radzen.delaci.co.uk/)

Setup as draft request while I ensure that I have pushed everything correctly and to open for discussion.

### DIFF
--- a/Radzen.Blazor/RadzenYearView.razor.cs
+++ b/Radzen.Blazor/RadzenYearView.razor.cs
@@ -1,0 +1,75 @@
+using Microsoft.AspNetCore.Components;
+using Radzen.Blazor.Rendering;
+using System;
+using System.Globalization;
+
+namespace Radzen.Blazor
+{
+    /// <summary>
+    /// Displays the appointments in a month day in <see cref="RadzenScheduler{TItem}" />
+    /// </summary>
+    /// <example>
+    /// <code>
+    /// &lt;RadzenScheduler Data="@appointments"&gt;
+    ///     &lt;RadzenMonthView /&gt;
+    /// &lt;/RadzenScheduler&gt;
+    /// </code>
+    /// </example>
+    public partial class RadzenYearView : SchedulerViewBase
+    {
+        /// <inheritdoc />
+        public override string Icon => "calendar_month";
+
+        /// <inheritdoc />
+        public override string Title
+        {
+            get => Scheduler.CurrentDate.ToString("yyyy", Scheduler.Culture);
+        }
+
+        /// <inheritdoc />
+        [Parameter]
+        public override string Text { get; set; } = "Year";
+
+        /// <summary>
+        /// Specifies the text displayed when there are more appointments in a slot than <see cref="MaxAppointmentsInSlot" />.
+        /// </summary>
+        /// <value>The more text. Set to <c>"+ {0} more"</c> by default.</value>
+        [Parameter]
+        public string MoreText { get; set; } = "+ {0} more";
+
+        /// <inheritdoc />
+        public override DateTime StartDate
+        {
+            get
+            {
+                var d = new DateTime(Scheduler.CurrentDate.Date.Year, 1, 1).StartOfWeek();
+                if (d.DayOfWeek == DateTimeFormatInfo.CurrentInfo.FirstDayOfWeek) d.AddDays(-7);
+
+                return d;
+            }
+        }
+
+        /// <inheritdoc />
+        public override DateTime EndDate
+        {
+            get
+            {
+                var d = new DateTime(Scheduler.CurrentDate.Date.Year, 1, 1).AddDays(DateTime.IsLeapYear(Scheduler.CurrentDate.Date.Year) ? 366 : 365).EndOfWeek();
+
+                return d;
+            }
+        }
+
+        /// <inheritdoc />
+        public override DateTime Next()
+        {
+            return Scheduler.CurrentDate.Date.AddYears(1);
+        }
+
+        /// <inheritdoc />
+        public override DateTime Prev()
+        {
+            return Scheduler.CurrentDate.Date.AddYears(-1);
+        }
+    }
+}
--- a/Radzen.Blazor/Rendering/YearView.razor
+++ b/Radzen.Blazor/Rendering/YearView.razor
@@ -1,0 +1,192 @@
+@using Radzen
+@using Radzen.Blazor
+@using Radzen.Blazor.Rendering
+@inject DialogService DialogService
+
+@{
+    var points = new Dictionary<AppointmentData, double>();
+
+    // delaci check usage
+    var year = StartDate.AddDays(7).Year;
+
+    DateTime date = StartDate.AddDays(7);
+    DateTime realstart = StartDate;
+    int daysinmonth;
+}
+
+
+<div class="rz-year-view-content">
+
+    @for (int month = 1; month < 13; month++)
+    {
+        realstart = new DateTime(year, month, 1);
+        daysinmonth = DateTime.DaysInMonth(year, month);
+        date = realstart.StartOfMonth().StartOfWeek();
+
+        <div class="rz-view-month">
+            <div class="rz-year-view-month-content">
+
+                <div class="rz-year-month-title">
+                    @realstart.ToString("MMMM yyyy")
+                </div>
+
+                <div class="rz-year-view-header">
+                    @for (var dateheader = StartDate; dateheader <= StartDate.EndOfWeek(); dateheader = dateheader.AddDays(1))
+                    {
+                        <div class="rz-year-slot-header">
+                            @dateheader.ToString("ddd", Scheduler.Culture).Substring(0,1)
+                        </div>
+                    }
+                </div>
+
+                @for (var row = 0; row < 6; row++)
+                {
+                    <div class="rz-year-week">
+                        <div class="rz-events">
+                            <div class="rz-year-slots">
+                                @for (var col = 0; col < 7; col++)
+                                {
+                                    var day = date.AddDays((row * 7) + col);
+                                    var appointments = AppointmentsInSlot(day, day.AddDays(1));
+                                    var excessCount = appointments.Count();
+                                    string classname;
+
+                                    @if (day.Month != month)
+                                    {
+                                        if (excessCount > 0)
+                                        {
+                                            classname = "rz-year-slot-not-in-month-with-appointments";
+                                            <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-year-slot")>
+                                                <div class="@classname">
+                                                    @day.Day
+                                                </div>
+                                            </div>
+                                        }
+                                        else
+                                        {
+                                            classname = "rz-year-slot-not-in-month";
+                                            <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-year-slot")>
+                                                <div class="@classname">
+                                                    @day.Day
+                                                </div>
+                                            </div>
+                                        }
+                                    }
+                                    else
+                                    {
+                                        if (excessCount > 0)
+                                        {
+                                            classname = "rz-year-slot-normal-with-appointments";
+                                            <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-year-slot")>
+                                                <div class="@classname">
+                                                    @day.Day
+                                                </div>
+                                            </div>
+                                        }
+                                        else
+                                        {
+                                            classname = "rz-year-slot-normal";
+                                            <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-year-slot")>
+                                                <div class="@classname">
+                                                    @day.Day
+                                                </div>
+                                            </div>
+                                        }
+                                    }
+                                }
+                            </div>
+                        </div>
+                    </div>
+                }
+            </div>
+        </div>
+    }
+</div>
+<style>
+
+
+
+</style>
+
+
+
+@code {
+    [Parameter]
+    public DateTime StartDate { get; set; }
+
+    [Parameter]
+    public DateTime EndDate { get; set; }
+
+    [Parameter]
+    public int MaxAppointmentsInSlot { get; set; }
+
+    [Parameter]
+    public string MoreText { get; set; }
+
+    [CascadingParameter]
+    public IScheduler Scheduler { get; set; }
+
+    [Parameter]
+    public IEnumerable<AppointmentData> Appointments { get; set; }
+
+    IDictionary<string, object> Attributes(DateTime start, string className)
+    {
+        var attributes = Scheduler.GetSlotAttributes(start, start.AddDays(1));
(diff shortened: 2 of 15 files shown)


Radzen.Blazor/RadzenYearView.razor.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <inheritdoc />
+        [Parameter]
+        public override string Text { get; set; } = "Year";
+
+        /// <summary>
+        /// Specifies the maximum appointnments to render in a slot.
+        /// </summary>
+        /// <value>The maximum appointments in slot.</value>
+        [Parameter]
+        public int? MaxAppointmentsInSlot { get; set; }
```
akorchev, line 39: I think it is better the MaxAppointmentsInSlot to have a set default numeric value (as in MonthView) for consistency.

Radzen.Blazor/RadzenYearView.razor.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <value>The more text. Set to <c>"+ {0} more"</c> by default.</value>
+        [Parameter]
+        public string MoreText { get; set; } = "+ {0} more";
+
+        /// <summary>
+        /// Specifies the background color of the Month label in the view />.
+        /// </summary>
+        /// <value>The color of the month labels. Will use current theme if not set.</value>
+        [Parameter]
+        public string MonthLabelColor { get; set; } = "";
```
akorchev, line 53: Styling is usually left to the theme. Is there a particular need for this to be a property?

Radzen.Blazor/RadzenYearView.razor.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        /// <summary>
+        /// Specifies the background color of the cells that don't fall in the month for that row in the view />.
+        /// </summary>
+        /// <value>The color of the cells in a month row that are not part of the previous or next month. Will use current theme if not set.</value>
+        [Parameter]
+        public string NotInMonthColor { get; set; } = "";
```
akorchev, line 60: Same as `MonthLabelColor` - is there a need for this to be a property and not set through CSS?

Radzen.Blazor/Rendering/YearView.razor
```diff
@@ -0,0 +1,255 @@
+@using Radzen
+@using Radzen.Blazor
+@using Radzen.Blazor.Rendering
+@inject DialogService DialogService
+
+@{
+    var days = 0;
```
akorchev, line 7: Seems unused (shown by the build action).

Radzen.Blazor/Rendering/YearView.razor
```diff
@@ -0,0 +1,255 @@
...
+    var points = new Dictionary<AppointmentData, double>();
+    var startyear = StartDate.AddDays(7).Year;
+    DateTime date = StartDate;
+    DateTime realstart = StartDate;
+    int daysinmonth;
+
+    const double NUMBER_DAYS_COLUMNS = 37.0;
+    const double MONTH_LABEL_WIDTH = 50.0;
```
akorchev, line 15: Also seems unused.

akorchev (changes requested):

akorchev:
Did some research and here is more feedback about the UI.

Most other vendors have implemented yearly view as a series of calendars next to each other (that go one below the other on a mobile device). Here is an example from Google calendar:

image

So basically 12 MonthViews next to each other with a header for the month name. All tested vendors rendered only an indicator (single dot most of the time) when there are events on a date. Clicking the dot behaves like the more button in the current month view - displays a dialog with the appointments for that day.

One of the vendors however had a view that was similar to what you have implemented (they have the google calendar-like year view as well). They called that view YearTimeline though.

For consistency's sake we prefer the YearView to look as in the pasted screenshot for Google Calendar. We believe this is what developers would expect for an early view. However YearTimeline view also seems to have its merits and should not go to waste. I am thinking the following:

1. Rename it to YearTimeline view.
2. Do not rotate the month headers to vertical. Keep it horizontal and instead use horizontal scrolling always (the month names should stay fixed as a frozen datagrid column). 
3. Having horizontal scroll always would allow to use longer text for the day names instead of a single letter. 
4. Remove the month headers at the end as they would be unnecessary.
5. Also remove the text "MONTH" on top of the month headers - seems unnecessary.
6. Bonus points - implement the Year view. Could probably reuse some of MonthView's implementation.

What do you think?

Radzen.Blazor/RadzenYearView.razor.cs
paulo-rico: This is set in RadzenMonthView.razor and RadzenYearView.razor. Calculated based on the height of the slot.

`@using Radzen.Blazor
@using Radzen.Blazor.Rendering

@inherits SchedulerViewBase

@code {    
    public override RenderFragment Render()
    {
        var appointments = Scheduler.GetAppointmentsInRange(StartDate, EndDate);

        var maxAppointmentsInSlot = 0;

        if (MaxAppointmentsInSlot != null)
        {
            maxAppointmentsInSlot = MaxAppointmentsInSlot.Value;
        }
        else
        {
            var slotHeight = (Scheduler.Height - 60) / 6;
            maxAppointmentsInSlot = Convert.ToInt32(Math.Floor(slotHeight / 24)) - 1;
        }

        return @<CascadingValue Value=@Scheduler>
            <MonthView StartDate=@StartDate EndDate=@EndDate MaxAppointmentsInSlot=@maxAppointmentsInSlot MoreText=@MoreText Appointments=@appointments />
        </CascadingValue>;
    }
}`

Radzen.Blazor/RadzenYearView.razor.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <value>The color of the cells that fall on the weekend. Will use current theme if not set.</value>
+        [Parameter]
+        public string WeekendColor { get; set; } = "";
+
+        /// <summary>
+        /// Specifies the height of Appointments in the year view />.
+        /// </summary>
+        /// <value>The height of each appointment in the view. Initially set to 1.5</value>
+        [Parameter]
+        public double AppointmentHeight { get; set; } = 1.5;
```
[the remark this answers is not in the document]
paulo-rico: On reflection, for this view, it doesn't fit to have appointment heights of 1.5 showing content as in the other views. This view is more of an annual overview, where detail can be achieved on appointment click or hover for text. I will remove this and hard code it to a thin appointment height.

Radzen.Blazor/Rendering/YearAppointment.razor
```diff
@@ -0,0 +1,74 @@
...
+    <div class="rz-event-content" title=@Data?.Text @attributes=@Attributes>
+        @if(ShowAppointmentContent)
+        {
+            @Scheduler.RenderAppointment(Data)
+        }
+    </div>
+</div>
+@code {
+    [Parameter]
+    public double? Top { get; set; }
```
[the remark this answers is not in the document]
paulo-rico: Done. As per AppointmentHeight above, I have removed it as a property of the View and it is hidden, defaulting to true for all views, and overridden to false for YearView (soon to be renamed YearTimelineView).

Radzen.Blazor/Rendering/YearView.razor
paulo-rico: Removed

Radzen.Blazor/RadzenYearView.razor.cs
paulo-rico: I've created scss and css variables for each of these in _scheduler.scss and removed them as properties.

Radzen.Blazor/RadzenYearView.razor.cs
paulo-rico: I've created scss and css variables for each of these in _scheduler.scss and removed them as properties.

Radzen.Blazor/RadzenYearView.razor.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <value>The color of the cells in a month row that are not part of the previous or next month. Will use current theme if not set.</value>
+        [Parameter]
+        public string NotInMonthColor { get; set; } = "";
+
+        /// <summary>
+        /// Specifies the background color of the cells that don't fall in the month for that row in the view />.
+        /// </summary>
+        /// <value>The color of the cells that fall on the weekend. Will use current theme if not set.</value>
+        [Parameter]
+        public string WeekendColor { get; set; } = "";
```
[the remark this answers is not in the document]
paulo-rico: I've created scss and css variables for each of these in _scheduler.scss and removed them as properties.

paulo-rico:
Hi @akorchev 

EDIT - Apologies in advance for reply. I do ramble on if I don't think I'm explaining well enough.

Key for conversation

- GV - The Google view in the snapshot that you sent in the last comment
- TL  - The Timeline view from one of the other vendors
- YP - The proposed Timeline view (was based on regular Year Planner found hanging in many an office)

1. Absolutely with you on the GV. The ability to stack the separate month views to accommodate different screen sizes is a must. One thing I would do differently to other vendors is I would highlight the whole day if there is an appointment that crosses that day rather than just a dot. I think from a user point of view, it looks quite confusing and doesn't give a clear picture on availability, booked, busy, e.t.c.

2. My thoughts on TL differ. I had searched whether there were any other vendors that offered a 'Year View' and I did come across what you're referring to myself. I didn't think it was intuitive at all. It didn't give any more than using a Month View and manually 'nexting' through the year. In fact, it may have been less intuitive than 'nexting' through MonthView. I'll try to explain.

With the existing DayView, WeekView and MonthView, they all have context, or a term of reference. Without even looking at the data content, you instinctively know the boundaries and the context of what you are viewing. This was the thinking behind YP. You are viewing a whole year. You barely need to reference any headings, you instinctively know where you are.
With TL, you scroll up, down, left and right, trying keeping one eye on the headers so you can track your position. But even then, there's no context, or term of reference to what you're finally presented with, so the view doesn't really mean anything, or offer anything more than MonthView.

Analogy
It's a bit like a town/city route on Google maps. If you were supplied with a series of zoomed snapshots of a route, you'd be able to find your way to your destination if they were kept in order and with a little work. But if you zoom out to encompass the whole route, you are immediately presented with a view that offers far more. You have terms of reference. Scale. Distance. Could be a couple of landmarks you already know. Although you couldn't guess the exact distance or duration of the route, the landmarks would aid in giving you a very good idea. When presented with the full picture and context, the human brain is extremely adept at such calculations, usually with very little effort from the owner.

My initial thought is that GV shouldn't be too difficult to realise if using the existing MonthView as a starting point. I will have a play around with it over the next couple of days.

As for TL. I think it has the appearance that it's been created by a component designer to plug a gap in a library, rather than offer a real world solution to a problem (just my own personal opinion).

Be good to know your thoughts.

Speak soon and stay well.
Paul

akorchev:
I couldn't understand what you are opposing to. We like the implementation you have created and just want to suggest some changes in the rendering - wider cells in exchange for horizontal scrolling (also would look better on a mobile device). 

As for year view vs year agenda - I think it is a matter of preference and both have different use case. One is more of a summary and a very high level overview, the other could serve as a more detailed look over the yearly tasks.

paulo-rico:
I was worried the changes would have made it a totally different view to what was intended. So, instead, I've created a new view that encompasses all the changes you mentioned above. Now, the existing view is YearPlanner and the changed view is YearTimeline.

[Test site](http://radzen.delaci.co.uk)

paulo-rico:
Hi @akorchev 

Created "stacked month" year view. [Test site](http://radzen.delaci.co.uk/)

Would like to know your thoughts on layout / usability or any changes you think may improve it.

I need to clean up the code, especially the css. A lot of trial and error using flex display. I will then upload the code as a draft so you can check it out before I set it to ready for review.

Many Regards
Paul

akorchev:
Wow! That's some great progress! 

The only thing that bothers me with the Planner view is that it is almost unusable on a mobile device. The cells are misaligned and the headers are on top of each other. Maybe this can be fixed somehow.

IMG_4885
IMG_4884

paulo-rico:
Thanks @akorchev 
You are probably aware of my reluctance to totally lose the Planner view. It is being used in a real world application (desktop only) where scrollable wouldn't have been an option and I do think it has it's place.
I do see how you need controls to be able to be used on any device without the developer having to program for the different media. Unfortunately, having all the data on the screen on a small device is just not practical. By it's design definition, it really is just for a desktop or large media device.

So how does this sound? We combine the Planner and Timeline views into one. We add a property called "FullScreen", which will affect the control like this -

if FullScreen == true
    if ScreenSize (or container size) > PredefinedMinimum
        Show Planner
    else
        Show Timeline
else
    Show Timeline

This will allow a developer to show a full screen planner view if they wish, but will fall back gracefully to Timeline if displayed on small form factor device.

Many Regards
Paul

paulo-rico:
Hi @akorchev 

I think it's just about there [Test Site](http://radzen.delaci.co.uk/)

I've used a few @media queries so that on small devices the cells are fixed and scroll bars are used, whilst on larger devices, the cells will expand or shrink to fit the container.

Regards

Paul

akorchev:
Hey @paulo-rico,

This looks a lot better indeed! Great work. I found one small issue that I am sure is easy to fix. There is some white space between the rows after making the window smaller:

image

paulo-rico:
Hi @akorchev

I think it's just about there (again) [Test Site](http://radzen.delaci.co.uk/)

Limited height of content so it cannot stretch past a certain threshold. 

Tried many different views and zooms in Dev Tools and looks to be o.k.

Many Regards

Paul

akorchev:
It definitely looks better. The only thing that bothers me is that there just isn't enough room for everything:

image

Here it isn't clear which cell this "+3 more" applies to. Also if this has to appear in two adjacent cells overlaps will occur. Maybe the text should just be "..." so it can fit in the cell always.

paulo-rico:
MoreText changed to "+{0}" [Test Site](http://radzen.delaci.co.uk/)

Many Regards

Paul

akorchev:
The text is still outside of the event though. I think the slot needs a bit more height or padding:

image

akorchev (approved):

paulo-rico:
Hi @akorchev 

A mistake in my media queries. I'm going to clean up the CSS as best I can from this end, but definitely need someone with experience in this field to sort it out.

My initial clone is probably battered after the alterations I've been making in the last few weeks. There are a couple of changes in one existing razor.cs file and then the changes to the _scheduler.scss file. The rest are all new files.

I think I may have to park this clone, get a fresh clone, and then paste my changes in.

What's the best way of doing this?

Thanks for your patience

Paul

paulo-rico:
Hi @akorchev 

[Test Site](http://radzen.delaci.co.uk/)

I removed the original commit and worked from a fresh clone. Managed to sort a great deal on the css. I'd created that many incremental changes through our back and forths, that I think I lost my way. Using @media query, I could just use same css as MonthView for large full-screen display, and then use a seperate css for small screens that scrolls.

I think the css still needs to be tidied up by a css guru.

Regards

Paul

akorchev:
Thank you so much @paulo-rico, we will review the CSS later this week and merge your PR.

PR #783, opened 2023-01-25
Update button status on DataGridInLineEditPage.razor

brunosilva5:
After inserting a new object the insert button was disabled.


brunosilva5:
I think you also have to add the order to the list on OnCreateRow method, because if you add an object and try to delete it without refreshing the page, it won't appear in the list, that is, the if of the DeleteRow(..) method ( `orders.Contains(order` )) won't happen.

PR #787, opened 2023-01-30
Fix : DataGrid.ExpandItem

chouteau:
Separate DataGrid.ExpandItem on two methods ExpandItem and CollapseItem, this is to avoid the toggle effect


enchev:
Hey @chouteau,

Not sure why you doing this, however the method is internal and it is used exactly to toggle the DataGrid row expanded state in case of defined Template:
https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenDataGridRow.razor#L21

With your change you will unable to collapse an item once expanded when ExpandMode="DataGridExpandMode.Multiple". I'm afraid we cannot merge this.

PR #805, opened 2023-02-09
Replacing the boolean autocomplete parameters with strings.

DaveBagler:
This pull request replaced the boolean **AutoComplete** parameter with a string **AutoComplete** parameter. This allows developers to use the auto-complete options supported by web browsers. See the available options on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) and the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill).

This PR also adds a new `static class` that acts as an `enum` with each option linked to the appropriate attribute string value. It serves as a shortcut to avoid typos. 

Also, the tests are updated to use the new string parameter while maintaining the same functionality.

### DIFF
--- a/Radzen.Blazor.Tests/MaskTests.cs
+++ b/Radzen.Blazor.Tests/MaskTests.cs
@@ -119,11 +119,11 @@ namespace Radzen.Blazor.Tests
 
             var component = ctx.RenderComponent<RadzenMask>();
 
-            component.SetParametersAndRender(parameters => parameters.Add<bool>(p => p.AutoComplete, false));
+            component.SetParametersAndRender(parameters => parameters.Add<string>(p => p.AutoComplete, ACG.Off));
 
             Assert.Contains(@$"autocomplete=""off""", component.Markup);
 
-            component.SetParametersAndRender(parameters => parameters.Add<bool>(p => p.AutoComplete, true));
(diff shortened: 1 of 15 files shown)


akorchev:
Hi @DaveBagler,

I really wish you have contacted us before submitting this PR. I am afraid we can't accept so many breaking changes. This will break the application of basically everybody who has ever set the AutoComplete property. We can't allow that to happen. Maybe a compromise is to support a custom HTML attribute "autocomplete". Not sure if it would work but it is worth trying.

DaveBagler:
Yep, that's fair.

What about an `AutoCompleteType` or `AutoCompleteGuide` property used by the components if it is set, and if not, it falls back to the existing functionality of the `AutoComplete` property?

akorchev:
Yes, an extra property `AutoCompleteType` could definitely work! People can set it to `AutoCompleteType.Phone`, `AutoCompleteType.Name` etc. We can make it not require `AutoComplete` to be set to `true`. How does this sound?

DaveBagler:
Sounds good; I'll update the PR.

DaveBagler:
Sorry, it took me so long to submit the updated PR. I'm closing this PR in favour of https://github.com/radzenhq/radzen-blazor/pull/827.

PR #806, opened 2023-02-10
[WIP] Improve clickability of chart series

y2kbugger:
This is the start of a commit series to improve clickability of chart series.

Looking specifically for feedback about approach and testing.

# Motivation
I've noticed that clicking a line series datapoint can be a bit flaky.

image

At first I thought the hitbox was just too small, but with further testing I've noticed that sometimes you can be directly on the datapoint and a click still doesn't register. Additionally there is no user feedback on clickability, e.g. a cursor change.

The tooltip implementation feels natural and reliable.

# Current approach.
https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenChart.razor.cs#L282
It appears that the current approach is to loop through possible targets, checking if it is within a certain `tolerance`, and then triggering the `RenderTooltip` or `SeriesClick`. Right now the `tolerance`s is set to `25` and `5` respectively.

The `RenderTooltip` check also includes all overlays such as adjacent line segments, but `SeriesClick` does not.

# Solution
I would like to do two things here:

1. Fix the flakyness
2. Communicate that a click will succeed, either by changing the cursor, or highlighting a point somehow.

# Technical details
I don't yet fully understand the failure method flaky clicking. When a click doesn't register, I can click 5 times and nothing will happen. Moving over by a pixel or two and clicking will usually cause it register.

I no longer think the hitbox `tolerance` is the main concern.

One issue with expanding the hitbox: when there is lots of crowded data, you want to return the closest, not just the first matching data point.
```
foreach (var series in Series)
{
    if (series.Visible && series.Contains(mouseX - MarginLeft, mouseY - MarginTop, 5))
    {
        var data = series.DataAt(mouseX - MarginLeft, mouseY - MarginTop);

        if (data != null)
        {
            await series.InvokeClick(SeriesClick, data);
        }

        return;
    }
}
```
This is already done within a series, but not between series
```
public virtual object DataAt(double x, double y)
{
    if (Items.Any())
    {
        return Items.Select(item =>
        {
            var distance = Math.Abs(TooltipX(item) - x);
            return new { Item = item, Distance = distance };
        }).Aggregate((a, b) => a.Distance < b.Distance ? a : b).Item;
    }

    return null;
}
```

For giving feedback on clickability, we could overlay a caret opposite of the hover overlay, or alternatively, just sync the implementation of click and hover and showing empty caret if a `SeriesClick` is on but `ToolTip.Visible` is False. This informs the user of exactly which data point is about to be clicked.

https://forum.radzen.com/t/hitbox-for-seriesclick-is-tiny-on-line-charts/13087

### DIFF
--- a/Radzen.Blazor/RadzenChart.razor.cs
+++ b/Radzen.Blazor/RadzenChart.razor.cs
@@ -279,7 +279,7 @@ namespace Radzen.Blazor
         {
             foreach (var series in Series)
             {
-                if (series.Visible && series.Contains(mouseX - MarginLeft, mouseY - MarginTop, 5))
+                if (series.Visible && series.Contains(mouseX - MarginLeft, mouseY - MarginTop, 25))
                 {
                     var data = series.DataAt(mouseX - MarginLeft, mouseY - MarginTop);
 


akorchev:
Hi @y2kbugger,

Do you plan to make additional changes apart from changing the click threshold?

y2kbugger:
Yes, I don't even plan for the click threshold to be part of it anymore at
all  I'm still waiting for comment as asked above.

akorchev:
>  I'm still waiting for comment as asked above.

I am not sure what you have asked. Can you clarify?

y2kbugger:
"Looking specifically for feedback about approach and testing."

It's not actually just too small of a hit box. The problem is that you can
actually click it and have no response. It's a true bug.

I hoping someone who was familiar with the click code could give feedback /
intuition of what might be causing the intermittent click problem.

As far as testing, I Just wanted to make sure I knew what needed to be
tested after making a change to the clicking.

I haven't had time to dig in to just do basic print statements to see if
it's hitting the RadZen code at all.

akorchev:
I am the one who has implemented that feature and to be honest don't know of a case where the click event won't fire provided all conditions are met: there is a click event and it is within some chart series according to the hit test implementation. If you have discovered a bug please provide some reproduction steps.

y2kbugger:
I'll host an example site sometime today.

Steps to reproduce:

Create a scatter plot with a click handler.
Attempt to click on various markers. About 30% of the time you can be right
on top of a marker and no click will fire moving around then coming back to
the exact same spot will allow the click to trigger.

y2kbugger:
Thank you, I'll test this and get back to you.

y2kbugger:
> I reproduced a problem with SeriesClick not firing always when the user clicks a marker. This commit should improve the behavior: [e5f0412](https://github.com/radzenhq/radzen-blazor/commit/e5f041208fd5b21bc9dd76a298e04178d7d6e848). The chart now ignores clicks on markers (in JavaScript) and the marker itself listens to click events and triggers SeriesClick.

This does in fact fix the intermittent problem.


I am now going to focus on improving "Close clicks", as well as a tooltip option for "Absolute closest, rather than ignoring when over a tool tip" Right now tooltips mask nearby points so messes up case of lots not close data.

closepoints

akorchev:
> as well as a tooltip option for "Absolute closest, rather than ignoring when over a tool tip" Right now tooltips mask nearby points so messes up case of lots not close data.

I am afraid I don't understand that. I don't think the tooltip should respond to click events at all as this is the behavior I have seen with other tooltips.

y2kbugger:
No, what I mean is an option to get rid of the hysteresis that is in the
system right now.

Currently, if a tool tip pops up it will stay popped up even if you move
the mouse around closer to a new point. This is obviously to reduce jitter
and make it easier to read but when you're creating a crowded cloud of data
that for exploring, you would want the tooltip to be the absolute closest
point to the mouse rather than just the first one that triggered a tool tip.

akorchev:
In that case the threshold for the tooltip should probably be a parameter instead of the hardcoded value. Maybe a property of TooltipOptions would suffice.

y2kbugger:
Ok I have found another large bug that makes data dense charts seem glitchy. The Cartesian `DataAt` function only used distance in `x` this meant that only left right distance was used for choosing the closest point. 

Original:

        public virtual object DataAt(double x, double y)
        {
            if (Items.Any())
            {
                return Items.Select(item =>
                {
                    var distance = Math.Abs(TooltipX(item) - x);
                    return new { Item = item, Distance = distance };
                }).Aggregate((a, b) => a.Distance < b.Distance ? a : b).Item;
            }

            return null;
        }

Versus:

        public virtual object DataAt(double x, double y)
        {
            if (Items.Any())
            {
                return Items.Select(item =>
                {
                    var dx = TooltipX(item)-x;
                    var dy = TooltipY(item)-y;
                    var distance_squared = dx * dx + dy * dy;
                    return new { Item = item, Distance = distance_squared};
                }).Aggregate((a, b) => a.Distance < b.Distance ? a : b).Item;
            }

            return null;
        }

y2kbugger:
Additionally, Line charts' contains function assumes data is neat using a polygon method that _often_ fails on certain cases:


        public override bool Contains(double x, double y, double tolerance)
        {
            var category = ComposeCategory(Chart.CategoryScale);
            var value = ComposeValue(Chart.ValueScale);

            var points = Items.Select(item => new Point { X = category(item), Y = value(item) }).ToArray();

            if (points.Length > 0)
            {
                if (points.Length == 1)
                {
                    var point = points[0];

                    var polygon = new[] {
                        new Point { X = point.X - tolerance, Y = point.Y - tolerance },
                        new Point { X = point.X - tolerance, Y = point.Y + tolerance },
                        new Point { X = point.X + tolerance, Y = point.Y + tolerance },
                        new Point { X = point.X + tolerance, Y = point.Y - tolerance },
                    };

                    if (InsidePolygon(new Point { X = x, Y = y }, polygon))
                    {
                        return true;
                    }
                }
                else
                {
                    for (var i = 0; i < points.Length - 1; i++)
                    {
                        var start = points[i];
                        var end = points[i + 1];

                        var polygon = new[] {
                            new Point { X = start.X, Y = start.Y - tolerance },
                            new Point { X = end.X, Y = end.Y - tolerance },
                            new Point { X = end.X, Y = end.Y + tolerance },
                            new Point { X = start.X, Y = start.Y + tolerance },
                        };

                        if (InsidePolygon(new Point { X = x, Y = y }, polygon))
                        {
                            return true;
                        }
                    }
                }
            }

            return false;

y2kbugger:
https://green-stone-000cd3d10-preview.centralus.2.azurestaticapps.net/testplot

here is the comparison.

akorchev:
I think most of the problem comes from the fact that you are trying to use a line chart to implement a scatter plot. The tooltip implementation uses only the X coordinate as there are rarely points with the same Y coordinate and different X in a regular line chart. How does the new algorithm perform with a normal line chart such as our [demo](https://blazor.radzen.com/line-chart)? Can you add such an example for us to compare? 

To be honest I think it would be times easier to implement a new ScatterPlot series. The tooltip implementation there would be trivial - the squares (or other shapes) will fire the mouse events and there is no need for additional math involved. Such an implementation would also render faster as there is no invisible line (with Stroke=0) to draw. What do you think?

y2kbugger:
I was thinking that. I'm sure there are other specialized features for scatters as well.

I think line plots still need fixing after doing more testing:
selectable
unselectable

Comparing the two, because we do the first pass `Contains` filter, we cause some points to become "unselectable". Additionally we actually lose that ability to move left/right as high charts does. I am pretty sure HighCharts is doing the absolute closests point.

This will entail adding a `Distance(Item)` to `Series`.

Then we could replace

```
for s in series {
    if series contains mouse {
        tooltip for closest point for series
    }
}
```
with
```
closestpoints = select closest for each series
tooltip for min(closestpoint)
```
This would may also be faster because the `Contains` with polygons is a heavy compared to euclidean distance.
We would also need to work out how with other series as well as interaction e.g. bar + line, bar+scatter. I think having `Distance(Item)` return distance to the centroid for pie and bar would be good enough, distance to nearest point would be harder.

We could threshold on distance/do some hybrid with domain if you don't want the highcharts effect of give tooltip no matter how far you are from line.

y2kbugger:
Let me produce a demo implementation for "absolute closest point" with a few more multi-series types.
Then tackle Scatter in a separate PR

akorchev:
I don't think pie series work with other types of series too well at the moment so you don't have to test them in combinations. Bar and column charts have very simple hit tests - they are recatangles. The problem is with curves and points that are too close together. We can indeed see if the closest point algorithm works better in the general case (or similar to the current one).

akorchev:
Hi @y2kbugger,

We've recently made a few changes thanks to another PR that affect the clickability of charts. Feel free to check the latest release of Radzen.Blazor to see if it makes a positive difference for your case.

PR #807, opened 2023-02-10
Fixed DatePicker Kind getting lost when type is DateTime

sergiuciudin:
Fixed DatePicker Kind getting lost when type is DateTime

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor.cs
+++ b/Radzen.Blazor/RadzenDatePicker.razor.cs
@@ -547,7 +547,7 @@ namespace Radzen.Blazor
                     DateTimeOffset? offset = DateTime.SpecifyKind((DateTime)Value, Kind);
                     await ValueChanged.InvokeAsync((TValue)(object)offset);
                 }
-                else if (Value != null)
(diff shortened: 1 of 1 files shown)


Radzen.Blazor/RadzenDatePicker.razor.cs
```diff
@@ -547,6 +547,10 @@ protected async Task ParseDate()
                     DateTimeOffset? offset = DateTime.SpecifyKind((DateTime)Value, Kind);
                     await ValueChanged.InvokeAsync((TValue)(object)offset);
                 }
+                else if (Value != null)
+                {
+                    await ValueChanged.InvokeAsync((TValue)(object)DateTime.SpecifyKind((DateTime)Value, Kind));
```
sergiuciudin, line 552: Well, there is a problem with DatePicker. Kind value is lost when updating the value and is defaulted to unspecified. I'm not too confident about this casting `(DateTime)Value` so please advise.

sergiuciudin: added also type checking

PR #809, opened 2023-02-12
Added Password with Reveal Eye option

codepend:
Hi,

I've added a Reveal Password option to the RadzenPassword class. 

A couple things to note: 
I could not figure out the CSS structure so I re-used some classes from rz-calendar. I would be happy to fix/update this if I knew how.
I tried to follow Radzen's naming conventions.
I added a test as well as updated the components demo page.

Also, if you don't like this or the way it was done, I will not take it personally. Just trying to contribute.
Thanks for any consideration!

### DIFF
--- a/Radzen.Blazor.Tests/PasswordTests.cs
+++ b/Radzen.Blazor.Tests/PasswordTests.cs
@@ -113,6 +113,23 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void Password_Renders_AllowPasswordRevealParameter()
+        {
+            using var ctx = new TestContext();
+
+            var component = ctx.RenderComponent<RadzenPassword>();
+
+            component.SetParametersAndRender(parameters => parameters.Add<bool>(p => p.AllowPasswordReveal, true));
+
+            Assert.Contains(@$"rz-button-icon-left", component.Markup);
+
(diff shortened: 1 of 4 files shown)


enchev:
Hi @codepend,

Thanks for the pull request! 
There is a problem when width is defined
image

image

codepend:
@enchev  I see. I will try and figure out.

codepend:
I have implemented some changes which I believe resolve previously mentioned issues as well as added ability to allow password reveal while the component is Disabled. Happy to work on any issues you see with it.

enchev:
please rebase your branch against the master in order to check your changes

codepend:
I rebased this, believe it is up to date now using the new "config" components for demo pages

akorchev:
The latest implementation is problematic as well. The Style is rendered twice which is wrong. Also there is another wrapper which seems unnecessary. I am closing this PR as this feature needs to be addressed by a more robust rendering implementation.

codepend:
Thanks again for the consideration. I will try and work on something more appropriate.

akorchev:
The biggest problem is the rendering. Adding a wrapper element is a breaking change and should be fully tested that it doesn't break too many cases. This is why we prefer to implement it ourselves. Thank you for your efforts though! We appreciate the good will.

If you need this for your application I would implement it in a custom component which renders RadzenPassword or RadzenTextBox depending on a property value which is toggled by the icon.

Issue #810, opened 2023-02-13
RadzenDonut chart display error when items in serie are zero


Shinigami9000:
Bug Description:

The radzen donut series throws a errors in the dev tool each time the given series (collection) has a value of 0 (zero). 
Chart is trying to be drawn but has no default fall back method when the given series is empty. Bug was discoverd when building my own application. But the given bug is fully repreducible on the radzen demo components as shown in the below pictures.
![demo error](https://user-images.githubusercontent.com/58690140/218408175-856cd09e-9aa8-4ec7-b9d9-7ff16152776f.PNG)

![demo variables](https://user-images.githubusercontent.com/58690140/218406112-6736e32a-0c0c-4559-92ac-7e3a827052fe.PNG)

To reproduce the BUG
1. go to https://blazor.radzen.com/donut-chart
2. Click on 'Edit Source'
3. Scroll down to '@code' and to the collection 'DataItem[]'
4. Edit all 'Revenue' values in collection to 0 (Zero) (as shown in the above pictures)
5. Select 'Run' in top left corner
6. Wait for example page to load in
7. Open Browser Dev Tool
8. Navigate in Browser Dev Tool to 'console' tab to view errors

Expected behavior:
When all items in series have a value of 0 (zero) show an empty circle with no fill in (color) to indicate that the given series is empty.

Current behavior:
No circle is drawn and only the circle body text is shown and errors are displayed in browser dev tool

Reference to original Radzen forum Post:
https://forum.radzen.com/t/radzendonut-series-display-error-when-items-in-collection-are-zero/13007/5

akorchev:
Should be fixed in #813

PR #813, opened 2023-02-13
Pie and Donut render errors

paulo-rico:
As per issue [Issue](https://github.com/radzenhq/radzen-blazor/issues/810#issue-1581898242) / [Forum Entry](https://forum.radzen.com/t/radzendonut-series-display-error-when-items-in-collection-are-zero/13007)

[Working demo](http://radzen.delaci.co.uk/)

Checks for instances where sum of data = 0 and draws an empty circle rather than nothing (this was causing rendering errors in the browser where some of the values were NaN).

Whilst looking at this, there were also render errors when drawing data labels. These have been resolved.

Also found an error during testing when Data was set to a null object, i.e. DataItem[] revenue; These will also render as an empty circle rather crashing (crashing, not browser rendering error).

Regards

Paul Ruston

### DIFF
--- a/Radzen.Blazor/CartesianSeries.cs
+++ b/Radzen.Blazor/CartesianSeries.cs
@@ -214,12 +214,6 @@ namespace Radzen.Blazor
         /// <value>The items.</value>
         protected IList<TItem> Items { get; set; } = new List<TItem>();
 
-        /// <summary>
-        /// Stores <see cref="Data" /> filtered to items greater than zero as an IList of <typeparamref name="TItem"/>.
-        /// </summary>
-        /// <value>The items.</value>
-        protected IList<TItem> ItemsGreaterZero { get; set; } = new List<TItem>();
-
         /// <inheritdoc />
         public RadzenMarkers Markers { get; set; } = new RadzenMarkers();
 
@@ -394,8 +388,6 @@ namespace Radzen.Blazor
                     {
                         Items = Items.AsQueryable().OrderBy(CategoryProperty).ToList();
                     }
-                    
-                    ItemsGreaterZero = Items.Where(e => Value(e) > 0).ToList();
                 }
 
                 await Chart.Refresh(false);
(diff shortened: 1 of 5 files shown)


akorchev:
Hi @paulo-rico,

Thank you for the PR. I think the approach should be different than checking if the sum of all series is 0. It should ignore 0 values altogether (in fact values that are `<= 0`). Also the sum can be zero in other cases (e.g. -1 and 1).

In such edge cases I always check how Highcharts behaves (the most capable JS chart IMHO) . Here is what it does when the sum is 0 but there is one positive value:

```
    series: [{
        name: 'Share',
        data: [
            { name: 'Chrome', y: -1 },
            { name: 'Edge', y: 1 },
        ]
    }]
```

image



The negative series is completely ignored but the positive is still rendered. Here is what it does when all values are 0.


```
    series: [{
        name: 'Share',
        data: [
            { name: 'Chrome', y: 0 },
            { name: 'Edge', y: 0 },
        ]
    }]
```

image

It displays data labels (with arrows which the Radzen Chart does not support) but the chart is just a circle.

And here is what it does when all values are negative:

```
    series: [{
        name: 'Share',
        data: [
            { name: 'Chrome', y: -1 },
            { name: 'Edge', y: -1 },
        ]
    }]
```
image

In short I think that the algorithm should be like this:

1. Filter values that are less than or equal to 0.
2. If the collection is empty - draw the empty circle. 

How does this sound?

paulo-rico:
Hi @akorchev 
Hope this looks better. Negative values had never registered with me for a Pie Chart. I must learn to have a component developer hat on instead of a system developer one 😁

[Demo Site](http://radzen.delaci.co.uk/)

Many Regards

Paul

akorchev:
That looks better! I have one small request - can you move this property to the series that need it (pie, donut) instead of the base class?

paulo-rico:
That's done.

Regards Paul

Issue #820, opened 2023-02-19
RadzenHtmlEditorFormatBlock does not allow example text translation


Arturace:
**Is your feature request related to a problem? Please describe.**
The `RadzenHtmlEditorFormatBlock` does not allow changing the example text, like "normal" and "heading #".
This is an issue when you have strict restrictions upon the language your app is allowed to use.

**Describe the solution you'd like**
Allowing to set additional parameters in the `RadzenHtmlEditorFormatBlock` component. Like `HEadingExampleText` and `NormalExampleText`.

**Describe alternatives you've considered**
Recreating the component in my project with the example texts translated to the needed language. The issue with that solution is that  `Title` and `Placeholder` are already parameters, so why not have the example texts be there as well.

**Additional context**
I will be making a pull request shortly.

PR #821, opened 2023-02-19
Add parameters for the example texts in RadzenHtmlEditorFormatBlock.

Arturace:
See #820.
Tests pass.


akorchev:
Hi @Arturace,

Thank you for the PR. Could you please rename the properties? They should not contain "Example" in their names.
`NormalExampleText` should be `NormalText`. Also please add properties for all headings instead of using just one.

Arturace:
Hi there, my pleasure, my team are having a pretty good time working with your library.

Will do what you are asking, even though I believe that "example" for the example texts does make sense.
I do have to ask what's the benefit of having a property per heading?

Arturace:
Done @akorchev.

PR #827, opened 2023-02-21
Adding an autocomplete type parameter.

DaveBagler:
This PR adds a new `AutoCompleteType` parameter to the `RadzenMask`, `RadzenNumeric`, `RadzenPassword`, `RadzenTextBox`, and `RadzenLogin` components. It allows developers to tell the browser what type of built-in autocomplete to use. It doesn't impact the existing functionality of the `AutoComplete`.

It also adds a static class that serves as a string-associated enum with each of the commonly supported browser autocomplete types, and a new unit test for each of the four form field components.

See more of the conversation here: https://github.com/radzenhq/radzen-blazor/pull/805


akorchev:
Hi @DaveBagler,

Wouldn't it be better if AutoCompleteType were an enum? It would lead to a better developer experience. You can use an extension method to convert it to the string representation. What do you think?

DaveBagler:
@akorchev, I'm just looking around to see if people use any non-standard options, but if it's just the standard set, it could work. It's a question of whether it should be a limited set or an open string parameter with a helper for the standard options.

DaveBagler:
@akorchev, I've replaced the static class with an enum and added a new unit test to ensure the enum values get converted into the expected string values.

akorchev:
Thanks! I will review it later today and merge it.

akorchev:
Outstanding PR! Thank you @DaveBagler .

PR #832, opened 2023-02-25
[PanelMenu] Show arrow icon when ChildContent exists

benghaner:
**How it's working today**
The arrow icon displays when the menu item has menu items inside.

**Proposal**
Also show the arrow icon when there is ChildContent inside, so that the menu item can be used both ways.

**Before**

before

**After**

after


enchev:
Unfortunately I reverted your change since suddenly all items in our demos navigation received expand icon even if they don't have child items.

PR #834, opened 2023-02-27
Fix Regression where previously StrokeWidth=0 allowed for "Scatterplots" without lines

y2kbugger:
Change last week prevents 0 width stroke: https://github.com/radzenhq/radzen-blazor/commit/82010353c69099056e3990873673162ed70d91dd

It is sometimes useful to set `StrokeWidth` to zero (default is 2) to make "Scatterplot" style plots.

image
image

As far as i'm concerned this one is ready to merge.


y2kbugger:
@akorchev

PR #838, opened 2023-03-01
Addition of a Source View for Html Editor

paulo-rico:
Added a toolbar button for switching between rendered and source view.
When in source view, the formatting tools are disabled.

This gave rise to a new property in toolbar buttons using the editor's 'View' to enable or disable the button. This is a flag, so you can create a button that will be active in 'Rendered', 'Source' or both (can't think what could be done in both, but I'm sure someone will have a need in the future). This defaults to Rendered so as not to break existing code.

I also tried, instead of disabling buttons, hiding them. The UI looked messy switching between the two views.

Look forward to your thoughts...

Best Regards

Paul Ruston


paulo-rico:
Hi @akorchev

The JavaScript is only used for RadzenHtmlEditorCustomTool (Templated). I found that even when disabled, the tool still looked active in contrast to the other toolbar elements. Because you cannot know what is going to be in this toolbar (template), I found the only way to make it appear and act disabled was to iterate the tree of the custom div and add some "disable look" classes as well as actually disable them.

"_The mode button should invoke a Blazor method that toggles the mode property and this would change the enabled state of the toolbar buttons too._"

Other than Custom Templated, this is how it functions. It changes the view of the editor. 

`    protected override async Task OnClick()
    {
        if (Editor.View == HtmlEditorView.Rendered)
        {
            Editor.View = HtmlEditorView.Source;
        }
        else
        {
            Editor.View = HtmlEditorView.Rendered;
        }
        await Editor.StateChanged();
    }`


All the buttons disabled properties are set by a logical OR of the editors disabled property and whether the buttons ViewEnabled (ModesEnabled) flag contains the editors View (Mode) flag.

`disabled=@(Editor.Disabled || Disabled || !ViewsEnabled.HasFlag(Editor.View))`

I'll get on with changing those names and that property scope.

Many regards

Paul Ruston

paulo-rico:
Hi @akorchev 

Totally with you on JavaScript.

The reason I had to resort to it was for the templated custom tool.
 
CustomTool has this code 

`    @if (Template == null)
    {
        <EditorButton Title=@Title Click=@OnClick Icon=@Icon Selected=@Selected Disabled=@Disabled ViewsEnabled="@ViewsEnabled"/>
    }
    else
    {
        <div @ref="divTemplate" class="rz-html-editor-custom-tool">@Template(Editor)</div>
    }`

EditorButton does all the disabling in Blazor as needed. It's only when a Template is used that the problem occurs. 
Setting the same disabled code i.e. `<div @ref="divTemplate" disabled=@(Editor.Disabled || Disabled || !ViewsEnabled.HasFlag(Editor.View)) class="rz-html-editor-custom-tool">@Template(Editor)</div>` doesn't work. Everything within the div (Template) is still enabled.

Not sure where to go with this. Any ideas?

akorchev:
I think rendering a new class when the template is disabled is enough. People can use it to act appropriately. Or they can use the Editor instance which is passed as argument to determine what the current mode is.

paulo-rico:
Hi @akorchev 

I'm happier with that. On reflection, the JavaScript code was untestable for all eventualities. There could have been anything in that Template!

Disabling CustomTool templated controls based on Editor.Mode works a treat. 

Thanks

Paul

akorchev:
Thank you @paulo-rico! I will review it once again and act accordingly. A bit overwhelmed with other stuff ATM.

akorchev:
I've pushed some changes:

1. Made the tool last instead of first.
2. The tool now blurs (unlike other EditorButton instances) in order to trigger the change event of RadzenTextBox and update the Html when switching modes.
3. No longer use the Mode property to track the current state. It is a `[Parameter]` and should not be used for that - strange things happen if a parameter is also used as state. Using `mode` , `GetMode()` and `SetMode()` instead.
4. Made some methods internal.
5. Simplified the CSS.

Things seem to work as they should. You can check it out and test too.

paulo-rico:
Looks great. Everything works fine except ability to disable templated custom controls. Need to have access to the mode property for that. Do you have any objections to `GetMode()` being made Public in order to achieve this?

akorchev:
> Do you have any objections to GetMode() being made Public in order to achieve this?

No. Should be fine to be public.

Issue #841, opened 2023-03-02
Support programatically changing selected view on RadzenScheduler


Conduci:
**Is your feature request related to a problem? Please describe.**

There seems to be no way to programatically select RadzenScheduler's view. SelectedIndex only dictates which index is selected on initialize. The user has to click a Scheduler view selection button at the top which calls a private `OnChangeView` handler.

**Describe the solution you'd like**

There should be `SelectView(ISchedulerView)` method on the Scheduler akin to `AddView(ISchedulerView)`

**Describe alternatives you've considered**

Tried forcing re-render after changing SelectedIndex and some other workarounds but they were buggy at best.

akorchev:
Setting SelectedIndex is supposed to work: https://forum.radzen.com/t/scheduler-programmatically-set-view-day-week-month/12111. What did you try?

Conduci:
I've tried exactly that. It changes it once. Then when you switch back to month view, and click again, nothing happens.

Here are some code snippets. I've just re-tried in in the radzen demo project.
```razor
<RadzenScheduler @ref=@scheduler SlotRender=@OnSlotRender style="height: 768px;" TItem="Appointment" Data=@appointments StartProperty="Start" EndProperty="End"
    TextProperty="Text" SelectedIndex="@SelectedIndex"
    ...
```
```razor
@code {
private int SelectedIndex { get; set; } = 2;
...
}
````


```razor
async Task OnSlotSelect(SchedulerSlotSelectEventArgs args)
    {
        SelectedIndex = 0;
        await scheduler.Reload(); 
     }
```

The Docs on the SelectedView even say "Specifies the initially selected view"

Conduci:
Regardless of whether this is a bug or not, I still believe it's beneficial to have #842

akorchev:
Try without calling Reload(). The scheduler should reload if selected index changes (check the linked code in the PR).

We prefer to have only one method of changing views instead of two.

Conduci:
I did. The issue persists. Tried with and without StateHasChanged(); too

I believe the issue here is that the SelectedIndex which is passed as the parameter stays the same. When you manually click on Month, only the scheduler's private `selectedIndex` changes to 2. When you select the slot again, the selected index is set again to 0, which is it current value, both inside and outside the Scheduler. Check line 376 in side the RadzenScheduler.razor.cs, maybe you should compare the parameter to selectedView, although that seems hacky.

[...]

PR #842, opened 2023-03-02
Added SelectView(ISchedulerView) to RadzenScheduler (#841)

Conduci:
Allows users to programatically change the selected view. 

More information in issue #841


akorchev:
Hi @Conduci,

There is code which should handle changing the selected index from code: https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenScheduler.razor.cs#L378

What exactly did you try?

PR #849, opened 2023-03-07
Created RadzenDataGridCustomHeader

ksli78:
Created DataGridCustomHeaderPage.razor
Created DataGridCustomHaderWithColumnPickerPageSamples Update RadzenDataGrid.razor to display the render fragment for RadzenDataGridCustomHeader if it is not empty Updated the ExampleService to have the menu options for RadzenDataGridCustomHeader


enchev:
Hi @ksli78,

Thanks for your contribution! In my opinion instead of whole new component maybe it will be easier to have just a template similar to Template property. You can name it for example HeaderTemplate.

ksli78:
> Hi @ksli78,
> 
> Thanks for your contribution! In my opinion instead of whole new component maybe it will be easier to have just a template similar to Template property. You can name it for example HeaderTemplate.

After looking at it, I agree ! Let me see what I can do.. I have a little time today

PR #853, opened 2023-03-09
Added IsInInlineEditOrAddState flag to the RadzenDataGrid.razor.cs

ksli78:
Came across a small bug in the RadzenDataGrid component where if the grid is in inline edit mode and the user is adding or editing a row and accidentally clicks on the cell header the grid sorts and the edits disappear. I added the IsInInlineEditOrAddState Boolean flag to the gird which is set by edit/insert state and unset after a save or cancel

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -30,6 +30,7 @@ namespace Radzen.Blazor
     /// </example>
     public partial class RadzenDataGrid<TItem> : PagedDataBoundComponent<TItem>
     {
+        private bool IsInInlineEditOrAddState = false;
 #if NET5_0_OR_GREATER
         /// <summary>
         /// Gets or sets a value indicating whether this instance is virtualized.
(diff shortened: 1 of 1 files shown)


enchev:
Hey @ksli78,

I'm afraid this is how the DataGrid works - there are other operations like filter, group, page, etc. and the editing will be canceled explicitly.  This is also a breaking change in the behavior of the DataGrid and not all users will be happy - we cannot accept this pull request. You can implement the same in your case outside of the DataGrid component.

ksli78:
That’s fair ! But then is there a way to make the new record appear at the bottom row instead of at the top so close to the cell header? My reason was I have many users with accessibility issues and I testing we noticed that they would click on a sort during an edit/new operation. I can fix it only I my app but I figure it may be something others would need. Perhaps I went about it the wrong way maybe it could be a property on the grid that that can be enabled or disabled by the developer and it can have a default value of allowing the sort to happen this way it won’t break anyone. Sent from my iPhoneOn Mar 10, 2023, at 1:19 AM, Vladimir Enchev ***@***.***> wrote:﻿
Hey @ksli78,
I'm afraid this is how the DataGrid works - there are other operations like filter, group, page, etc. and the editing will be canceled explicitly.  This is also a breaking change in the behavior of the DataGrid and not all users will be happy - we cannot accept this pull request. You can implement the same in your case outside of the DataGrid component.

Issue #857, opened 2023-03-13
Allow access to Element property when inheriting RadzenComponent


scottfridwin:
**Is your feature request related to a problem? Please describe.**
Creating custom component that inherit the RadzenComponent class are unable to use the Element property.

**Describe the solution you'd like**
Change the setter for RadzenComponent.Element to `protected` instead of `internal`.

**Describe alternatives you've considered**
No alternatives except removing the inheritance.

**Additional context**
I found a few other mentions of this issue.
https://github.com/radzenhq/radzen-blazor/issues/79
https://forum.radzen.com/t/wasm-components-inheritance/10281

There may be a design reason for keeping the setter internal, but if there is not it would be very convenient to allow inheritance.

PR #858, opened 2023-03-13
Change RadzenComponent.Element setter to protected internal

scottfridwin:
#857 
Changing this setter allows custom components inheriting the RadzenComponent class to reference the html element in @ref tags.


scottfridwin:
> Hi @scottfridwin,
> 
> Your change breaks the compilation...

Thanks! Since it was so small, I made the change directly in github and didn't actually compile it. Updated to `protected internal` which will maintain access to all internal types as well as allow inheritance.

PR #873, opened 2023-03-25
Link/MenuItem/PanelMenuItem/ProfileMenuItem - Add NavLink-Match Property

msdevcode:
1. Add NavLink-Match Property to RadzenLink/RadzenPanelMenuItem/RadzenProfileMenuItem 
2. Set the missing Parameter-Attribute at RadzenMenuItem


akorchev:
Hi @msdevcode  and thank you for the PR!

You have changed the formatting in RadzenLink in a way which would affects the appearance (added new lines in the span with class rz-link-text). Browsers visualise new lines in some cases which breaks the appearance. Please keep the existing formatting.

msdevcode:
Good evening @akorchev, 

thank you for your explanation. I reverted the new lines.

PR #874, opened 2023-03-25
Scheduler changes - Better handling of Yearview clicks and overridable MoreText event

paulo-rico:
Addressing following forum issues -
https://forum.radzen.com/t/scheduler-do-nothing-on-slot-click/13572
https://forum.radzen.com/t/radzen-scheduler-monthview-moretext-dialog/13573

[Demo Site](http://radzen.delaci.co.uk/)

YearView will now follow this logic

```
if(appointments)
    if(SelectAppointment has delegate)
        if(appointment count == 1)
            call SelectApppointment
        else
            open dialog showing appointments to click. If SelectSlot has a delegate, also show "Slot Clicked" button
    else
        if SelectSlot has a delegate call SelectSlot
else
    if SelectSlot has a delegate call SelectSlot

```
Some guidance required. On the multiple appointments dialog for YearView, I have currently hard-coded the button with text "Slot Clicked". I know it needs to be a parameter. Should I place it in YearView (it only applies to this view), or do you prefer to have all paramaters at Scheduler level?

MoreText Event
Is just that. When "More Text" is clicked, it checks whether the new SelectMore event has a delegate. If so, it calls it, else it behaves as normal showing a list of appointments.


akorchev:
Hi @paulo-rico and thank you for the PR.

> Some guidance required. On the multiple appointments dialog for YearView, I have currently hard-coded the button with text "Slot Clicked". 

I couldn't find this in the changed files. Where is that button and what does it do?

I have some feedback about the implementation. Please remove the `HasSelectMoreDelegate` and similar public properties. Those are implementation specific and better not appear in the public API and intellisense. You can just use `Scheduler.MoreSelect.HasDelegate` when needed. Also make the methods that trigger events internal (or again use `Scheduler.MoreSelect.InvokeAsync()`

paulo-rico:
Hi @akorchev 
                    
It's in the OnListClick event in YearView.razor in Rendering folder. Not so easy to spot. It's razor within dialog service. Because you only have a spot to click in the YearView, this gives the option to call SelectSlot method instead of AppointmentSelect (if SlotSelect has a delegate).

```
await DialogService.OpenAsync(date.ToShortDateString(), ds =>
    @<div class="rz-event-list">
        <CascadingValue Value=@Scheduler>
            @foreach (var item in appointments)
                {
                    <Appointment Data=@item Click="OnListEventClick" />
                }
            </CascadingValue>
            <br />
            @if (Scheduler.HasSelectSlotDelegate)
            {
                <RadzenButton Text="Slot Clicked" Click="((args) => OnSlotEventClick(date))" />
            }
        </div>
    );
}
```
As for the extra properties. I couldn't get to the SlotSelect, AppointmentSelect and MoreSelect to use HasDelegate property as they aren't in IScheduler (that's what is passed to the Views as opposed to Scheduler). Yet, when I put them in IScheduler, it wouldn't allow me to make them private in Scheduler.

Could we just pass Scheduler instead of IScheduler to the views?

akorchev:
> Could we just pass Scheduler instead of IScheduler to the views?
You can't as Scheduler is generic and it would make all views generic as well.

I don't understand what's the purpose of that button with text Slot Clicked though. The year view should not do anything if an empty slot (date) is clicked I think (as all other views do). We are probably safe to remove the current behavior.

akorchev:
I thought about it more and I got what you are trying to do with that button. It shouldn't be a button though. I will discuss it with the team next week. Probably we shouldn't' display a dialog at all in Year view and just navigate to day view. And fire SlotSelect.

paulo-rico:
Ok @akorchev I'll hang fire while you discuss with team. 
Stay well and speak soon

paulo-rico:
YearView
Don't know if this fits, but may be something to think about when you discuss with team.
In effect, the click on a day that has an appointment is the equivalent of clicking the More link in other views. We could check whether SelectMore has a delegate first and react to that. This gives the developer the opportunity to fully customize what they want to do when an appointment is clicked in YearView.

akorchev:
Here is what I think about YearView:

1. We support the SlotSelect event. If the developer handles it - we do nothing but fire it.
2. If it is not handled we keep the existing behavior - show "there are no events for this date" or list the events.
3. We fire AppointmentSelect if the default dialog that displays events is shown.
4. We don't do anything about MoreSelect as we don't display a "more" button in YearView.

How does this sound?

paulo-rico:
I did have another idea that I've just implemented here to see how it goes. We create a new event called DaySelect. Would only be used in the YearView (it is a bit of a unique situation not being able to click on Slot or Appointment). This day event is called like this (in YearView)
```
    async Task OnDayClick(DateTime date, IEnumerable<AppointmentData> appointments)
    {
        await Scheduler.SelectDay(date, appointments);
    }
```
This way the developer is given the date and the list of appointments, and they choose how to handle it. It feels a lot cleaner and we're not pre-empting any UI or functionality at the component level.

[Demo Site](http://radzen.delaci.co.uk/)

What do you think?

akorchev:
I like the idea. What if we just extend the SlotSelect event to pass the appointments always? This way we won't introduce a new event that works only in one view and keep the same idea.

paulo-rico:
I'd already put these changes together before your suggestion. However, looking at your suggestion, we would still have to know whether an event had a delegate, and I'm not sure how that could be implemented without exposing the event to the interface.

paulo-rico:
I like it. I'll revert back and try those changes this end.

paulo-rico:
Ok. I've put those changes in and it looks to be working with very minimal testing (the England match is kicking off!). I'll commit now so you can take a look. I'll test more after game, but initially looks well.
Speak soon

paulo-rico:
I've run it through and it does look OK. I'll leave it with you.

akorchev:
There seems to be quite substantial change in the rendering of MonthView. Why was it necessary? 

Also we have to keep the `Task SelectSlot(DateTime start, DateTime end);` method for compatibility reasons. It can pass an empty appointment list as it used to do.

akorchev:
> Git must be reporting all the tab changes as well. The code tabs were not aligned properly.

It is indeed nested one level deeper. While the code looks better this causes one level of extra indentation in the output HTML. Let's keep the original indentation.

paulo-rico:
Pushed twice. I reverted back to original tabs but VS saw no changes. So I had to put in a dummy comment, commit, then remove.

akorchev:
I see the exact same changes: https://github.com/radzenhq/radzen-blazor/pull/874/files#diff-b1dfc69fd75f9e0235a0b6728741421d6e4811badb27f41775f56b31f7d4db4a

The thing is the whole file will blame to you if we accept this commit which is obviously not intended.

paulo-rico:
Can't understand that. In order to revert back to absolute original, I took a copy of MonthView from radzenhq and just changed that SlotSelect line of code. I'm out for a couple of hours, so I'll work on it when I get back in.

akorchev:
Probably your VS code is configured to automatically format the code which is causing changes in the whole file. As a result every line would appear to be modified by you and would make it harder to use git to track issues.

paulo-rico:
Hi @akorchev
I've not got auto format on. I have to manually action a format for my code in Visual Studio 2022. That said, something is obviously amiss.

What I'm going to do is -

- Copy my project locally
- Revert back to start
- Put my changes in again, line by line, without touching any formatting
- Commit and Push changes

Because VS 2022 git doesn't report changes the same way as git, I'll have to push what VS 2022 tells me is OK and then check it online. So, for now, ignore any new commits while I make sure that git is only reporting actual code changes I've made rather than any spurious formatting changes. When git is only reporting code changes, I'll tag you in a comment for your review.

Thanks for your patience
Paul

akorchev:
Instead of checking online you can install git locally (if you don't have it yet) and just run `git diff` before committing the changes.

paulo-rico:
I've downloaded GitHub desktop and ensured that the only changes visible are code changes and not format changes. All looks well. YearView does have many changes and they are valid. Removal of redundant methods and condense some if...else.... to inline test ? true : false.

akorchev:
I confirm that it looks better now. Will test tomorrow and merge.

Thank you!

paulo-rico:
@akorchev
In testing the new SlotSelect signature and passing appointments, I think it has highlighted a fundamental error in the rendering of some views. As far as I can see, it applies to Month, Planner and Timeline. It only occurs in a specific circumstance.
I am publishing an example of the problem to my test site. I will follow this comment up with details to accompany the link. It may take a short while to word it correctly.

akorchev:
Yes, this is how rendering is currently implement. Project #0 does not appear in the second case as it is in the "+4 more". Currently events that start earlier than the current slot take precedence and will render before the other.

paulo-rico:
Sorry @akorchev I may not be explaining well. It's the second part that's the problem. February 8th. If you click the "+3 more", it shows, rightly, five appointments (projects #1 through #5). But only one is rendered. Project #2 should start to render here as the space has been made available with the termination of project #0. But project #2 doesn't begin to render until the next row.

I think I've identified a fix, but it would be better to keep it separate from this PR.

```
                            @if (item.Start >= realstart && item.Start <= end)
                            {
                                <Appointment Data=@item Top=@top Left=@left Width=@width Height=@height Click=@OnAppointmentClick />
                            }
                            else if (realstart == start)
                            {
                                left = startSlotIndex * slotWidth;
                                length = Math.Max(1, Math.Min(daysinmonth, Math.Ceiling(item.End.Subtract(date).TotalDays - (startSlotIndex - 1))));
                                width = length * slotWidth;

                                <Appointment Data=@item Top=@top Left=@left Width=@width Height=@height Click=@OnAppointmentClick />
                            }
```
These are the two render possibilities, but they don't take into account appointments that can come into view because of a newly created space, but have a StartDate prior to the current slot render date.

akorchev:
I see. Well the current rendering implementation has its issues for sure. You can indeed open a separate PR for that. If you are ready with this PR do let me know so I can perform a final review before merging it.

akorchev:
I tried the PR but something feels off. Maybe we underestimated it. Now nothing happens in year view even if SlotSelect is not handled. Keeping the existing behavior is a must I think. Maybe we need to implement the equivalent of preventDefault in JavaScript. Something like 

void OnSlotSelect(SlotSelectEventArgs<T> args)
{
    args.PreventDefault(); // sets some internal flag which tells the view what to do after firing the event e.g. display a dialog or not.
}

Also what happened with the MoreSelect event? It doesn't seem to exist in the latest PR.

paulo-rico:
> I thought about it more and I got what you are trying to do with that button. It shouldn't be a button though. I will discuss it with the team next week. Probably we shouldn't' display a dialog at all in Year view and just navigate to day view. And fire SlotSelect.

I reverted back and implemented the SlotSelect event as discussed above and further conversation adding appointments to the SlotSelect output. I was leaving MoreEvent for another PR. The reason being is, with my original commit, it was imperative that we knew if AppointmentSelect, SlotSelect and MoreSelect had a delegate in order to act accordingly. Other than passing this information via the interface, which was unacceptable, I couldn't work out how it would be achievable.

Regards the prevent default suggestion, I think this is too late in the chain. By the time OnSlotSelect has been called, the view has already reacted in it's own way. I think the cleanest way is have some knowledge in the view of the event delegate status. Or we could have three more properties on the Scheduler that PreventDefaultSlotSelect, PreventDefaultAppointmentSelect and PreventDefaultMoreSelect. I don't know. Just throwing ideas around.

Do you have any suggestion on how we could pass this "event delegate status" to the views without contravening the interface rules? Are the Prevent Properties acceptable? I'll leave it with you.

Thanks @akorchev 
Paul

akorchev:
It shouldn't be too late to prevent the default UI. Consider something like:

```

void OnListClick()
{
     var prevented = await Scheduler.SlotSelect();
     
     if (!prevented)
     {
         await DialogService.ShowDialog();
     }
}

we would have to introduce a new SlotSelect method though - one that returns a boolean.

```
        public async Task<bool> SelectAppointment(AppointmentData data)
        {
            var args = new SchedulerAppointmentSelectEventArgs<TItem> { Start = data.Start, End = data.End, Data = (TItem)data.Data };

            await AppointmentSelect.InvokeAsync(args);

            return args.IsDefaultPrevented;
        }
```

and SchedulerAppointmentSelectEventArgs has two new additions:

```
public bool IsDefaultPrevented { get; private set; }
public void PreventDefault()
{
    IsDefaultPrevented = true;
}
```

This would allow us to make the original idea of handling SlotSelect work. We also need to pass the current ISchedulerView as well so the developer has enough context. 

If this sounds as too much work to you let me know - I will implement it myself when time allows.

**Comment by akorchev** [comment]
My idea is for the developer to have control over when to prevent the default UI provided by the scheduler. I think it is better than to just unconditionally stop showing it if an event is handled. Here is how one would use it:

```
void OnSlotSelect(SlotSelectEventArgs args)
{
     if (args.View.Name == "Year" && !args.Appointments.Any())
     {
          args.PreventDefault(); // Do not show the default UI "There are no events for this date
     }

     // Show default UI if any for every other case.
}
```

Does this make more sense?

paulo-rico:
It does and on it.

paulo-rico:
As soon as I saw how a developer would use it, it immediately fell into place. Should have clocked it before. Hope this is what you envisaged. Thanks @akorchev, for your patience and persistence 👍

akorchev:
Look fine! The only thing which bothers me a little is that we have to update the demo probably not to show the add / edit dialog in Year view as it shows both. I will sleep on it and decide tomorrow.

akorchev:
I pushed a commit to update the demo. Everything else seems fine. Do you think you have anything else to add?

paulo-rico:
Not on this PR, Atanas.

I will create a new PR for rendering appointments that should come into view when a space is made available from the termination of an appointment in the same row. It's only the same 4/5 lines to be added to Month, Planner and Timeline views, so it won't be as problematic a solution as this one.

Again, thanks for your patience. I know with the time spent hand-holding through this PR, you could probably have just done it yourself. But component development is a different strategy that not all programmers will be accustomed to (that would definitely include me!), and hopefully your continued guidance will develop those skills for me so I can contribute to the library with minimum fuss/supervision in the future.

Thanks and stay well
Paul

akorchev:
Thank you for the continuous effort and contributions! I want to assure you that this is time well spent for both of us and I don't consider it hand-holding. Cheers!

PR #878, opened 2023-03-28
Reset DataGrid in Settings.set

cvanstone:
Currently the DataGrid Settings are only reset if the Settings property is called with null.  If an application is developed with multiple `DataGridSettings` that can be saved and loaded, any secondary set is loaded in addition to the first for column sorting.

This commit resets the DataGrid each time the Settings setter is called prior to the updated value being applied and Reload being called meaning the new Settings will be applied from fresh.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -2936,25 +2936,24 @@ namespace Radzen.Blazor
             {
                 if (settings != value)
                 {
+                    canSaveSettings = false;
+
+                    Groups.Clear();
+                    CurrentPage = 0;
+                    skip = 0;
+                    Reset(true);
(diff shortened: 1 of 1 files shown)


enchev:
Hi @cvanstone,

Thanks for the pull request however your change will cause endless loop in this demo:
https://blazor.radzen.com/datagrid-save-settings-loaddata

I'm afraid we cannot accept this pull request.

cvanstone:
Sorry, could you explain the endless loop?  The setter already has `if (settings != value)` protection so the Reset wouldn't be called each time any update is made, only when the object ref changes.  

I tested on your demo before committing and it works fine.  I also extended the demo to allow local caching (non-persistent) of 2 additional `DataGridSettings` objects that I could switch between to confirm that the new settings were applied correctly, thus resolving the issue I had noted.

Sorry if I'm missing something obvious here...

PR #882, opened 2023-03-29
Reset DataGrid in Settings.set

cvanstone:
Currently the DataGrid Settings are only reset if the Settings property is called with null. If an application is developed with multiple DataGridSettings that can be saved and loaded, any secondary set is loaded in addition to the first for column sorting.

This branch resets the DataGrid each time the Settings setter is called prior to the updated value being applied and Reload being called meaning the new Settings will be applied from fresh.

As highlighted by **enchev** in my first pull request, this caused a loop issue in the [LoadData Demo](https://blazor.radzen.com/datagrid-save-settings-loaddata).  The `DataGridSettings` were being deserialized and loaded on every `OnAfterRenderAsync` once Data has been `loaded`.  This only needs to happen on `firstRender` to avoid continual loop.

I also noted that the Reload button was targeting the IQueryable page.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -2936,25 +2936,24 @@ namespace Radzen.Blazor
             {
                 if (settings != value)
                 {
+                    canSaveSettings = false;
+
+                    Groups.Clear();
+                    CurrentPage = 0;
+                    skip = 0;
+                    Reset(true);
+                    allColumns.ToList().ForEach(c =>
+                    {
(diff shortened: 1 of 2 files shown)


enchev:
Hey @cvanstone,

Nothing changed - the loop is still there:
settings-loaddata-loop-filter

cvanstone:
Hi @enchev,

Sorry, I cannot replicate your behaviour with my updated commits.  Can you confirm that you were running my updated demo as it looks like the original.  

I have tried column picking, filtering, sorting and re-ordering multiple times and it never loops on the updated Load Data demo.

PR #907, opened 2023-04-07
Request to add an option to show pagenumbers or not.

MicrosoftHawk:


MicrosoftHawk:
I can work with that. As a suggestion I would like to say that pagenumbers are almost never really usefull so an option to hide / show them would seem kinda logical.

PR #908, opened 2023-04-07
Add ability to customize OData string by column

evandixon:
This allows users of the data grid to override OData expression generation for niche cases that aren't ready for the main repo. For my use-case, I have a column representing a List<string>, and I want to search for any one of the values in it. With this PR, I can inherit RadzenDataGridColumn to fit my specific needs:

```
    public class CustomProofOfConceptRadzenDataGridColumn<TItem> : RadzenDataGridColumn<TItem>
    {
        protected override string GetColumnODataFilter(object filterValue, FilterOperator filterOperator)
        {
            var property = GetFilterProperty().Replace('.', '/');

            if (FilterPropertyType == typeof(List<string>) && filterValue is string)
            {
                return $"{property}/any(x: x eq '{filterValue}')";
            }

            return base.GetColumnODataFilter(filterValue, filterOperator);
        }
    }
```

An idea for future improvement is to, instead of using a virtual method for the customization, add a Func<object, FilterOperator, string> parameter, though I didn't feel comfortable adding such an option without getting the opinions of the maintainers first.

### DIFF
--- a/Radzen.Blazor/QueryableExtension.cs
+++ b/Radzen.Blazor/QueryableExtension.cs
@@ -144,8 +144,8 @@ namespace Radzen
         public static string ToFilterString<T>(this IEnumerable<RadzenDataGridColumn<T>> columns)
         {
             Func<RadzenDataGridColumn<T>, bool> canFilter = (c) => c.Filterable && c.FilterPropertyType != null &&
-               (!(c.GetFilterValue() == null || c.GetFilterValue() as string == string.Empty) || c.GetFilterOperator() == FilterOperator.IsNotNull 
-                   || c.GetFilterOperator() == FilterOperator.IsNull || c.GetFilterOperator() == FilterOperator.IsEmpty 
+               (!(c.GetFilterValue() == null || c.GetFilterValue() as string == string.Empty) || c.GetFilterOperator() == FilterOperator.IsNotNull
+                   || c.GetFilterOperator() == FilterOperator.IsNull || c.GetFilterOperator() == FilterOperator.IsEmpty
                    || c.GetFilterOperator() == FilterOperator.IsNotEmpty)
                && c.GetFilterProperty() != null;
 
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/QueryableExtension.cs
```diff
@@ -583,18 +583,18 @@ private static string GetColumnODataFilter<T>(RadzenGridColumn<T> column, bool s
...
+        internal static string GetColumnODataFilter<T>(RadzenDataGridColumn<T> column, object filterValue, FilterOperator columnFilterOperator)
         {
             var property = column.GetFilterProperty().Replace('.', '/');
 
-            var columnFilterOperator = !second ? column.GetFilterOperator() : column.GetSecondFilterOperator();
             var odataFilterOperator = ODataFilterOperators[columnFilterOperator];
 
-            var value = IsEnumerable(column.FilterPropertyType) && column.FilterPropertyType != typeof(string) ? null :
-                    !second ? (string)Convert.ChangeType(column.GetFilterValue(), typeof(string)) :
-                        (string)Convert.ChangeType(column.GetSecondFilterValue(), typeof(string));
```
enchev, line 597: Looks like the second filter value is missing in the new implementation?

Radzen.Blazor/QueryableExtension.cs
```diff
@@ -654,9 +654,7 @@ private static string GetColumnODataFilter<T>(RadzenDataGridColumn<T> column, bo
             }
             else if (IsEnumerable(column.FilterPropertyType) && column.FilterPropertyType != typeof(string))
             {
-                var v = !second ? column.GetFilterValue() : column.GetSecondFilterValue();
```
enchev, line 657: Again, not sure why the second part of the filter is missing.

Radzen.Blazor/QueryableExtension.cs
evandixon: The responsibility for choosing which value to filter on has been moved to RadzenDataGridColumn.GetColumnODataFilter. This way, QueryableExtension can focus more on generating single OData expressions and less on the inner workings of the data grid column, improving maintainability. Maybe in the future there's a third filterable value, or even a list of acceptable values, for example.

PR #915, opened 2023-04-13
Add support for setting tab index for escape button in dialogs

DanRogers99:
Add support for setting the tab index for the escape button in dialogs within DialogOptions


enchev:
Hi @DanRogers99,

Can you provide additional info why you need this? You can create fully custom dialogs from HTML if you want some custom UI and settings - check our demos for reference.

enchev:
My idea was that you can define the content of the dialog including Ok and Cancel buttons and set desired tabindex:
image

DanRogers99:
This PR is to implement the setting of the tab index of the escape x in the top right-hand corner, not of the cancel button or anything within rz-dialog-content

enchev:
Sorry, I misunderstood the pull request completely! I'll merge it however in my opinion the property name should be change to CloseTabIndex - I'll do that after the merge.

PR #928, opened 2023-04-19
Add ReadOnly property to CheckBoxList and CheckBoxListItem

johanpolak:
Here is my pull request which adds the ReadOnly property to the CheckBoxList and the CheckBoxListItem. Additionally I added a commit to set the tabindex of the single CheckBox to -1 if the CheckBox is ReadOnly.

I have a few remaining questions about this approach.
1. Should the method 'SelectAll' ignore read-only and disabled checkboxes in the list?
2. Should the method 'IsAllSelected' ignore read-only and disabled checkboxes in the list?
3. How to handle CheckBoxListItems which are gerenated through the data-property. The CheckBoxList has parameters for the TextProperty and a ValueProperty, should we add similar properties for Disabled and ReadOnly?

I look forward to your reaction.

### DIFF
--- a/Radzen.Blazor/RadzenCheckBox.razor
+++ b/Radzen.Blazor/RadzenCheckBox.razor
@@ -5,7 +5,7 @@
 @inherits FormComponent<TValue>
 @if (Visible)
 {
-    <div @ref="@Element" @attributes="Attributes" class="@GetCssClass()" @onkeypress=@OnKeyPress @onkeypress:preventDefault style="@Style" tabindex="@(Disabled ? "-1" : $"{TabIndex}")" id="@GetId()">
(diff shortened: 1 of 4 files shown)


enchev:
Thanks @johanpolak,

Here are my comments:
1. In my opinion SelectAll() should not touch any item if ReadOnly is set to true
2. This method should work with all items not matter if they are disabled or set as read-only
3. There are similar implementations in other components - i.e. DisabledProperty and you can check them to know how to implement such functionality in the CheckBoxList component.

enchev:
Hey @johanpolak,

I think that adding just an example below is good enough for now. If you need to set ReadOnly property for items coming from data you can always loop the data and set desired property for the CheckBoxList item:
image

johanpolak:
Hi @enchev,

Thanks for completing the pull request. The methods 'SelectAll' and 'IsAllSelected' have to be changed to ignore read-only / disabled CheckBoxListItems. Will you do this or shall I do it and create another pull request?

PR #946, opened 2023-04-30
Color Picker: HSV picker base color fix + alpha resetting fix

Cosmatevs:
The color in the top right corner of the HSV stauration/value picker should be always the most saturated. Once you choose a predefined color or enter RGB values, it becomes the chosen color, not the most saturated one:
color picker hue issue

Here's a fix for this issue.


Cosmatevs:
There's also an issue regarding alpha when you choose a predefined color or change HEX:
color picker alpha issue
Fixed it as well: changing HEX won't reset alpha while choosing a predefined color will do it.

PR #950, opened 2023-05-03
Numeric: wraparound feature

Cosmatevs:
For circular ranges such as degrees of angle.
It could be also used in the DatePicker component in the future.


Cosmatevs:
It turns out there are javascript functions I haven't changed.
Is there a reason to use javascript events (`onpaste`, `onblur`) instead of blazor events (`@onpaste`, `@onblur`) in this component?

enchev:
Plain JavaScript is used to avoid entering out-of-range values.

PR #951, opened 2023-05-03
Exposed SearchText property and added Action SearchTextChanged

AndrzejKl:
Hi,
Maybe it will be useful to someone as it was useful to me.
Exposing `SearchText` and `SearchTextChanged` Action


AndrzejKl:
@enchev - thank you,
one question: how often You're updating demo page? I'm asking, because on example tab [https://blazor.radzen.com/dropdown-filtering](https://blazor.radzen.com/dropdown-filtering) my changes are not visible yet, but on tab `Edit source' are visible.

PR #962, opened 2023-05-05
fix pageSize change datagrid not reload work

pigwing:
@see https://github.com/radzenhq/radzen-blazor/commit/f5b7ddcf0e99e1a42f3359b11fef637032892a5d
PageSize changed should invoke Reload() only when needed
here if pageSize = value;
base.OnPageSizeChanged(value); await InvokeAsync(Reload);  will never be executed.


Cosmatevs:
Note that this order of operations will prevent page size settings from being saved. I'd go with:
```
await base.OnPageSizeChanged(value);
await PageSizeChanged.InvokeAsync(value);
SaveSettings();
```

Cosmatevs:
@pigwing Sure, I observed it today as well. I only wanted to say that after you commented out the `pageSize = value;` line, the `SaveSettings()` function has no chance to really save the new page size settings. That's because the new page size is now set with `base.OnPageSizeChanged(value)`.

pigwing:
go it

PR #966, opened 2023-05-08
Use method to get classes for RadzenStepsItem

johanpolak:
In our own project we want to add some classes to the RadzenStepsItem based on some extra parameters we addedin our own project. In order to achieve this, it is necessary that the classes for the RadzenStepsItem are determined via a method, so we can override the method 'GetComponentCssClass' for the RadzenStepsItem.

### DIFF
--- a/Radzen.Blazor/RadzenSteps.razor
+++ b/Radzen.Blazor/RadzenSteps.razor
@@ -43,7 +43,7 @@
         var step = steps[i];
         @if (step.Visible)
         {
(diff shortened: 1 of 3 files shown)


akorchev:
Hi @johanpolak,

Why did you remove the IsSelected method? It doesn't seem necessary for the implementation and is also a breaking change. Please revert it.

johanpolak:
Hi @akorchev,

That wasn't a good idea indeed. I reverted it.

Issue #963, opened 2023-05-05
Feature Request: RadzenCustomValidator


AndrzejKl:
**Is your feature request related to a problem? Please describe.**
No;

**Describe the solution you'd like**
I'm missing a CustomValidator, where I could indicate my own validation guidelines, e.g., check if a value is unique or does not occur more than x times.
I would see it as adding an Action that should return true or false.
Something like:
```
<RadzenText Name="Name" @bind-Value=customer.Name />
<RadzenCustomValidator Component="Name" Validate="@customValidation" Text="Name need to be unique"/>
```
``` c#
@code{
IEnumerable<Customer> customers;  
Customer customer;

  bool customValidation(Customer item)
  {
     return customers.Where(c=>c.Name.ToUpper().Equals(item.Name.ToUpper())).Count() > 0;
  }
}
```
**Describe alternatives you've considered**
None;

**Additional context**
None;

PR #968, opened 2023-05-08
Added CustomValidator component (source and Demo)

AndrzejKl:
What do You think? not to simple?

#963

### DIFF
--- a/Radzen.Blazor/RadzenCustomValidator.cs
+++ b/Radzen.Blazor/RadzenCustomValidator.cs
@@ -1,0 +1,59 @@
+﻿using Microsoft.AspNetCore.Components;
+
+namespace Radzen.Blazor
+{
+    /// <summary>
+    /// A validator component which compares a component value with a specified value.
+    /// Must be placed inside a <see cref="RadzenTemplateForm{TItem}" />
+    /// </summary>
+    /// <example>
+    /// <code>
+    /// &lt;RadzenTemplateForm TItem="Model" Data=@model&gt;
+    ///    &lt;RadzenTextBox Name="Email" @bind-Value=@model.Email /&gt;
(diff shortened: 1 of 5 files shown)


akorchev:
Hi @AndrzejKl,

We were postponing the implementation of a custom validator component as we believe it should support asynchronous validation. The implementation you are suggesting won't work in such scenarios. Adding asynchronous support requires a change in RadzenTemplateForm as well and maybe in the other validators.

akorchev:
I think using GetAwaiter().GetResult() isn't [considered best practice](https://stackoverflow.com/questions/39007006/is-getawaiter-getresult-safe-for-general-use) and can lead to [deadlocks](https://www.nikouusitalo.com/blog/why-is-getawaiter-getresult-bad-in-c/) in some cases.  In addition Blazor doesn't support async validation by design: https://github.com/dotnet/aspnetcore/issues/40244.

AndrzejKl:
@akorchev 
Understood, tomorrow I'll delete this pull request and branch

akorchev:
Don't delete it yet! Let us think it over a bit more. We may leave the initial pull request implementation.

akorchev:
Hi again @AndrzejKl,

I think the original implementation with `IsValid` as a boolean parameter should work. After all if the Blazor framework doesn't support async validation we should try to force it. 

Could you please update your PR?

akorchev:
Could you also rename CheckIsValid to just IsValid? The former sounds as a method but is a property.

akorchev:
It seems a boolean property doesn't work with onblur validation (only during submit). I will change it to a `Func<bool>` which seems to work better:

```
        [Parameter]
        public Func<bool> Validator { get; set; } = () => true;

        /// <inheritdoc />
        protected override bool Validate(IRadzenFormComponent component)
        {
            return Validator();
        }
```

PR #976, opened 2023-05-13
Improved chart tooltips

stlufred:
- Show closest tooltip in line chart (instead of the first close enough)
- Improved speed (tooltip performance unit test execution time reduced from 4.3 to 1.7 seconds)
- Show it over or under to greatly reduce chances of tooltip being offscreen

### DIFF
--- a/Radzen.Blazor.Tests/ChartTests.cs
+++ b/Radzen.Blazor.Tests/ChartTests.cs
@@ -45,9 +45,9 @@ public class ChartTests
         foreach (var _ in Enumerable.Range(0, 10))
         {
             await chart.InvokeAsync(() => chart.Instance.MouseMove(100, 80));
-            Assert.Contains("<div class=\"rz-chart-under-tooltip", chart.Markup);
+            Assert.Contains("<div class=\"rz-chart-tooltip", chart.Markup);
             await chart.InvokeAsync(() => chart.Instance.MouseMove(0, 0));
-            Assert.DoesNotContain("<div class=\"rz-chart-under-tooltip", chart.Markup);
+            Assert.DoesNotContain("<div class=\"rz-chart-tooltip", chart.Markup);
         }
         output.WriteLine($"Time took: {stopwatch.Elapsed}");
     }
--- a/Radzen.Blazor.Tests/ChartTests.cs
+++ b/Radzen.Blazor.Tests/ChartTests.cs
@@ -1,14 +1,22 @@
-﻿using System.Linq;
+﻿using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Bunit;
 using Radzen.Blazor.Rendering;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Radzen.Blazor.Tests;
 
 public class ChartTests
 {
+    private readonly ITestOutputHelper output;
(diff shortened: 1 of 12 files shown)


akorchev:
Hi @stlufred,

Thank you for the PR! I think one thing can be improved though - instead of passing `ref object cache` to TooltipX we can pass the actual Func which returns the value. The code which invokes TooltipX can do that. Something like:

```
internal virtual double TooltipX(TItem item, Func<TItem, double> value)
{
}
```

Do you think you can do that change in the PR?

stlufred:
> Hi @stlufred,
> 
> Thank you for the PR! I think one thing can be improved though - instead of passing `ref object cache` to TooltipX we can pass the actual Func which returns the value. The code which invokes TooltipX can do that. Something like:
> 
> ```
> internal virtual double TooltipX(TItem item, Func<TItem, double> value)
> {
> }
> ```
> 
> Do you think you can do that change in the PR? Also please use a named tuple for the DataAt method. It would improve the readability of the code a lot.

I'd like to make the change, however I fail to see how the solution you are suggesting can be implemented.

Most of the improvement in term of speed I'm suggesting is that TooltipX method can skip doing some work and re-use what was done in the previous execution when it is called multiple times in the same context.  And since the caller of TooltipX does not know (and should not know) how TooltipX is implemented, I fail to see how it would know how to implement the Func<TItem, double>.

Sorry if I missed something obvious...

akorchev:
I think I also found another issue with the shared cache approach. In theory one can have multiple types of series in the same chart and each of them could need a different func to calculate the tooltip position. So sharing the same cache between all series is potentially wrong to begin with.

Maybe we should pull the cache up and set it up in SetParametersAsync of CartesianSeries. Then we can eliminate it as a parameter of TooltipX - it would be a protected field instead. How does that sound?

stlufred:
I keep the cache per series so it shouldn't be a problem.  However I did not realized that people are supposed to always change values through the Blazor mechanics and not directly in C#.  Good idea, this way it should be even simpler than what I did.

stlufred:
Ok, I reworked the caching to move it in the Category method.  This allow caching to give gains in all parts calling the Category method on numeric and date categories.

stlufred:
@akorchev You are right, the caching only works when CategoryProperty is numeric or date.  I could also use it when it is of type scale, however I went toward that approach to "play safe" as some code could in theory call the method with different type of scale base without the cache being cleared which would cause problems.

akorchev:
I just tested the reverse tooltip feature and it looks worse that now TBH. Could you please revert that change?

BEFORE:
before

AFTER:
after

I have rebased the PR to the master branch and applied the category cache reset suggestion.

stlufred:
I must admit it looks better above, however I have problems that the tooltip goes "outside the window":
image

I admit the above / under approach is far from perfect (and does not solve going out of bound left or right).  A few ideas I have, which one you think is the best (or do you have an other one to suggest):

1. Add an option to activate / deactivate the under behavior
2. Find some other way to display my information (and remove that under tooltip mechanism I added)
3. Find a way to keep in the tooltip inside the chart area, but I found nothing easy...

akorchev:
Indeed this should be handled in some other way. Probably by using the TooltipService which has screen boundary detection as a built-in. Or at least its positioning JavaScript code.

PR #979, opened 2023-05-16
Change DatePicker icon based on TimeOnly parameter.

DaveBagler:
**Issue:** The **DatePicker's** button icon is `calendar_today` even if the **DatePicker's** `TimeOnly` parameter is set to `true`.

**Proposed Solution:** Substitute the `calendar_today` icon with the `schedule` icon when the **DatePicker's** `TimeOnly` parameter is set to `true`.

This PR updates the DatePicker's CSS, code and template to support the icon logic. I've also included two new unit tests in this PR to test which icon is displayed.


DaveBagler:
I noticed the build failed on the PR. It failed on data-grid-related code. I haven't touched that code in this PR, so I think it may be unrelated.

Issue #974, opened 2023-05-11
RadzenNumeric - Deleting all text leads to no data shown


fieldsj23:
** Describe the bug
Deleting all text from the RadzenNumeric box leads to the current value disappearing after hitting Enter or leaving the box.  Please note that it does work if you delete the value and either hit Enter or leave the box the first time.   After this, the value has been set back to zero or Min and when you delete a 2nd time, it leaves the box with an empty string.   To the end user, this box looks empty.

**To Reproduce**
1. Go to https://blazor.radzen.com/numeric
2. Click into field Min set to 1 and Max set to 10 
3. Delete the value in the RadzenNumeric 
4. Press Enter or click outside the RadzenNumeric 
5. It will change to a "1" in the box as the minimum is set to 1
6. Again, delete the value in the RadzenNumeric 
7. Press Enter or click outside the RadzenNumeric
8. No value will be seen, just an empty box

**Expected behavior**
After deleting the value, if it is empty when hitting Enter or leaving the box, it should display a number within the range.  This number could be the Min or the last valid value.

**Screenshots**
Following the steps above results in the following:
![image](https://github.com/radzenhq/radzen-blazor/assets/133258881/d9c43fa0-d032-4d68-a47a-e223b49aa298)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Edge Version 113.0.1774.35
                  Chrome Version 112.0.5615.140
 - Version Radzen.Blazor 4.10.4

**Additional context**

PR #980, opened 2023-05-18
#974

imatoria:
RadzenNumeric - Deleting all text leads to no data shown


enchev:
I'm afraid we cannot accept this pull request since it will brake immediately entering any number - will accept only empty string.

PR #989, opened 2023-05-29
Loop parameters once during SetParametersAsync

blake-fm:
Down from 6.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -1850,19 +1850,27 @@ namespace Radzen.Blazor
         /// <inheritdoc />
         public override async Task SetParametersAsync(ParameterView parameters)
         {
-            var emptyTextChanged = parameters.DidParameterChange(nameof(EmptyText), EmptyText);
+            bool emptyTextChanged = false, allowColumnPickingChanged = false, valueChanged = false, allGroupsExpandedChanged = false;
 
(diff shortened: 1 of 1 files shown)


blake-fm:
If you're interested, I can drop this over, also.  But it probably needs a bit more attention / clean up / add an allow param,so it's not a breaking change.
https://github.com/radzenhq/radzen-blazor/commit/e28565b712ae7ae54d9d0d8a3d834a026af92941

enchev:
Hi @blake-fm,

Thanks for your pull request however I'm not sure what's the purpose of this change. Will it solve something?

blake-fm:
By itself this is just optimisation.  `RadzenDataGrid` is currently looping through the list of parameters on every `SetParametersAsync` 6 times.  `SetParametersAsync` is called a lot more frequently than its name suggests.

The commit I mentioned in follow-up comment solves: `RadzenDataGrid` does not clean its list of selected items, when `SelectionMode` is changed from multiple to single.

When I added that ^ to my own code, I figured you could also benefit.  If you're not interested though, no worries!  Close away.

enchev:
The most significant difference in my opinion is using EqualityComparer instead parameters.DidParameterChange() and I'm not sure if this is correctly.

blake-fm:
That's an extension method in your `common.cs`.  It loops the parameter set and runs the same line.

enchev:
Well, yes, I need more coffee!

PR #991, opened 2023-05-29
Fixed Chart Click method

stlufred:
- Was not using x and y parameters
- Use the same maximum distance as the tooltip (so that click and tooltip returns the same thing)
- Fixed xDelta and yDelta that were not considering MarginLeft and MarginTop


akorchev:
Hi @stlufred,

Click and tooltip must not use the same distance otherwise it may not work with series that are less than 25px away from each other. Did you test for that case?

akorchev:
Nevermind. I just tested it and it seems to work fine. I will go ahead and create parameters for the magic value of 25 so people can tweak it if needed (ClickTolerance and TooltipTolerance).

PR #1021, opened 2023-06-13
Add parameter to determine whether to use OData filtering

### LINKED ISSUE #957: RadzenDataGrid doesn't know it's using OData until after first load
**Describe the bug**
In PagedDataBoundComponent.ReloadOnFirstRender, if Data is null on the first render, Reload will be called. RadzenDataGrid's implementation of Reload handles filtering and ordering. RadzenDataGrid.GetOrderBy gets the columns to order by, and calls IsOData to determine the format of the order. IsOData depends on knowing Data's type, which is null because that's required to trigger the reload on first render. Until https://github.com/radzenhq/radzen-blazor/commit/f5b7ddcf0e99e1a42f3359b11fef637032892a5d, an alternative was to set Data to `IEnumerable.Empty<MyClass>().AsODataEnumerable()` as a hint for IsOData, but now that solution results in the grid not loading on first render.

**To Reproduce**
1. Create a view model with an object such as the following:
```
public class GridItem
{
    public MyClass ViewModel { get; set; }
}

public class MyClass
{
    public string SomeProperty { get; set; }
}
```
2. Create a RadzenDataGrid with `TItem="GridItem"` with null Data and a LoadData delegate that uses OData to get data. Add a column referencing the property `ViewModel.SomeProperty` with `SortOrder="SortDirection.Ascending"`.
3. In the implementation of LoadData, observe how `LoadDataArgs.OrderBy` equals `np(ViewModel.SomeProperty) asc`. The "np" is a result of the grid thinking OrderBy is going to be used in a dynamic linq expression. (See RadzenDataGridColumn.GetSortOrderAsString)

**Expected behavior**
The LoadData implementation receives a LoadDataArgs.OrderBy equal to "ViewModel/SomeProperty asc" and not "np(ViewModel.SomeProperty) asc", perhaps by setting Data to Enumerable.Empty<GridItem>().AsODataEnumerable() or by having a separate parameter on the grid to force IsOData to true.

evandixon:
Fixes #957. With future refactoring can make ODataEnumerable obsolete, but for now not setting the new parameter preserves the original behavior.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -76,7 +76,7 @@ namespace Radzen.Blazor
             }
 
             await InvokeLoadData(request.StartIndex, top);
-            
+
             var totalItemsCount = LoadData.HasDelegate ? Count : view.Count();
 
             virtualDataItems = (LoadData.HasDelegate ? Data : itemToInsert != null ? (new[] { itemToInsert }).Concat(view.Skip(request.StartIndex).Take(top)) : view.Skip(request.StartIndex).Take(top))?.ToList();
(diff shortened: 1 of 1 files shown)


enchev:
Hi @evandixon,

Thanks for your pull request however in our opinion `ODataEnumerable` is the right way to check what kind of expressions should be used for filtering, sorting, etc. Sorry however we cannot merge this.

PR #1032, opened 2023-06-17
Allow dialog to resize dynamically after being dragged

Dreagen:


enchev:
Hey @Dreagen,

Thanks for your pull request however your change will reset any resized dialog to content size after drag.

Dreagen:
@enchev Good point, is there anything we can do about that? I could only null these values if it's not a resizable dialog?

PR #1044, opened 2023-07-01
Progress bar circular component + progress bar extension

Cosmatevs:
- ~~moved shared methods and properties to the ProgressBase class~~
- added `Min` property in progress components
- created demos for the ProgressBarCircular component
- updated the ProgressBar Max Value demo to include the new `Min` property

progress bar circular

Things I couldn't do:
- update documentation
- set styles in pro themes


MPapst:
Cool extension. Despite I am not part of the Radzen Team, but what do you think about a Size Property (like Button has one)?

Cosmatevs:
@MPapst I've already included it, even in the demo, but not mentioned it in the pull request description.

akorchev:
Hi @Cosmatevs,

Thank you for the pull request! The circular progressbar would be a great addition. Some feedback:

1. We can probably use the same enum for ProgressBarStyle and ProgressBarMode. This would lead to two extra enums less and you can move the Mode and ProgressBarStyle properties in the base class. Could you do that please?
2. We can also reuse a lot of the existing styling and probably not need a separate SCSS file. We will do that as well as the theme-ing.

akorchev:
Yes, ProgressBarCircular sounds better.

akorchev:
Thank you @Cosmatevs ! We will take it from here - we will make the required changes to the themes and merge the PR.

Cosmatevs:
@yordanov Your changes in color/determinate classes caused one of the progress bar tests to fail. I think that the new classes make more sense but I don't really know all the purposes of the failed test. Is it about backwards compatibility as well?

yordanov:
@Cosmatevs, Yes, renamed classes in ProgressBar to reduce redundancy:
Before: `rz-progressbar rz-progressbar-determinate rz-progressbar-determinate-success`
After: `rz-progressbar rz-progressbar-determinate rz-progressbar-success`

This CSS optimization should not lead to any visual or functional change. The test makes sure proper CSS classes are added based on the properties set.

PR #1047, opened 2023-07-05
do not hide 2 notification when clicking close

bc3:
solving issue found on forum : https://forum.radzen.com/t/closing-a-notification-closes-two-notifications/2020

when you remove a message, blazor will rebind and the message will become invisible, so no need to set visible to false

### DIFF
--- a/Radzen.Blazor/RadzenNotificationMessage.razor
+++ b/Radzen.Blazor/RadzenNotificationMessage.razor
@@ -54,7 +54,7 @@
    public void Close()
    {
       Service.Messages.Remove(Message);
-      System.Threading.Tasks.Task.Delay(0).ContinueWith(r => { Visible = false; });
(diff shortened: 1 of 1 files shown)


bc3:
if you need a test project, you can clone this from my GIT, just add a project reference to your Blazor.Radzen library and run. You will see a Toast button, when click, it will reveal 5 toasts. Then you you close 1 (not the last in the list), 2 will close

enchev:
Here is what I see with your example without any code modifications:
image

notification-close

All messages are closed normally one by one.

bc3:
@enchev please try it when you click an item in the middle .... the code indeed works if you start to click on the last item in the list


https://github.com/radzenhq/radzen-blazor/assets/2547576/ec45553b-d69a-4736-83b8-a3f898d6c3ea

PR #1054, opened 2023-07-08
add isLoading property to dropdown datagrid component

atafra92:
I have a minor component enhancement suggestion.
In the DropdownDataGrid component, there is no option for showing a spinner when moving through pages within data grid.
I am using Radzen components in my project extensively and this feature would help me a lot, as well as others I suppose.
I am already using the same functionality in all of my data grids so I would like to be consistent through my application.


enchev:
Thanks for the pull request @atafra92! Please restore the original formatting and add only the new code related to new IsLoading property.

atafra92:
@enchev sorry for the formatting issue, VS2022 sometimes messes up with code formatting..I reverted the commit and made only the required changes.

PR #1069, opened 2023-07-17
add * in RadzenformField using required parameter

bikotoru:
Example

```
   <RadzenFormField Text="Rut" Variant="Variant.Outlined" Required>
                       
                        <ChildContent>
                            <RadzenNumeric ShowUpDown="false" Max="99999999" Format="##,###,###" style="display: block; width: 100%" @bind-Value="@cliente.rut" Name="rut" />
                        </ChildContent>
                    </RadzenFormField>
```
image


akorchev:
Hi @bikotoru  and thank you for the PR!

I have a few comments on the implementation. The `Required` attribute suggest there is a RadzenRequiredValidator as well however there isn't one (and there shouldn't be one to be honest). Also the styling is hardcoded to `color:red`. 

I suggest a different implementation altogether which will solve all issues mentioned (and is also a lot simpler):

1. Remove the Required property
2. Revert the code to the original one.
3. Make it support MarkupString like this: https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenNotificationMessage.razor#L13

Then you can achieve the same result via `<RadzenFormField Text="Label <span style='color: red'>*</span>" ...>`.

Would you please update the PR?

PR #1070, opened 2023-07-17
Htmledit on input

DaveBagler:
Adds two parameters to the `RadzenHTMLEditor` component.

- `NoTools`: is a boolean that allows for an HTMLEditor component that doesn't have any tools.
- `Input`: is an event callback on input allowing developers to react to changes in the HTMLEditor without having to wait until blur.

I updated one of the HTMLEditor demos to include an implementation of `Input`.

### DIFF
--- a/Radzen.Blazor/RadzenHtmlEditor.razor
+++ b/Radzen.Blazor/RadzenHtmlEditor.razor
@@ -13,7 +13,7 @@
             {
                 @ChildContent
             }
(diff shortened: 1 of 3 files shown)


akorchev:
Hi @DaveBagler,

We will rename the NoTools parameter as we believe it is a strange name. We will also make it work with a custom set of tools as it will not work with the current implementation.

PR #1072, opened 2023-07-18
Added the ability to use oninput instead of onchange

bikotoru:
This is created so that it can be enabled that when a value is written it is modified immediately, and thus, for example, the validation does not have to wait for it to be executed when the focus of the input is lost


bikotoru:
> Hi @bikotoru,
> 
> Thank you for the PR. I am afraid I see one major issue in the implementation - the `@oninput` event is always handled which will cause a server roundtrip on every key press and affect all existing implementations potentially hurting performance. We can't accept this pull request in its current state as it changes the existing behavior too drastically.

There is a solution to this inconvenience since the interaction with the user makes it unnatural.??

PR #1073, opened 2023-07-18
fix progressbar no apply color

bikotoru:


akorchev:
Hi @bikotoru,

This is a breaking change and there is no description as to what issue it tries to resolve. We can't accept this pull request.

bikotoru:
> Hi @bikotoru,
> 
> This is a breaking change and there is no description as to what issue it tries to resolve. We can't accept this pull request.

solved

akorchev:
Still not sure what this PR is supposed to do and no description was given.

PR #1076, opened 2023-07-19
RadzenDataGridColumn: handle non-px widths of frozen columns

Cosmatevs:
Before this change, when you set the width of a frozen column using an unit other than px (e.g. 5rem or var(--my-width)) then it will be converted to 200px while calculating how far from the left side a frozen column should be. Here's an example result (in a [demo](https://blazor.radzen.com/datagrid-frozen-columns), I've set the first column width to 5rem):
image

After my change:
image

If there are more frozen columns on the left, then the function creates a calc(w1 + w2 + w3…) css expression.


enchev:
Hi @Cosmatevs,

Please resolve the conflicts in order to be able to merge it!

Issue #1084, opened 2023-07-24
Some notifications are not disappearing when multiple are opened


technight:
**Describe the bug**
If multiple notifications are triggered within a short amount of time, then some of them will not disappear.
If 3 is opened then 1 will not disappear, 5 open -> 2 will not disappear, 7 open -> 3 and so on.

**To Reproduce**
1. Go to https://blazor.radzen.com/notification
2. Click more than 3 times on one of the show notification buttons
4. See that some of the notifications will not disappear

**Expected behavior**
Notifications disappears after given time

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 115.0.2

**Additional context**
I think this issue may have started after the "Do not hide 2 notifications when clicking close." from release 4.13.6

bc3:
@technight I tested in in y local repo's, before update in PR #1084 , the same behaivour existed :-(

technight:
I belive #1092 should close this issue. The problem was that the loop in RadzenNotification.razor would update the Message parameter of an already existing component when a notification was closed. PR adds the key directive so that the parameters are fixed to the "correct" component instance.

PR #1085, opened 2023-07-24
Bugfix/notificat message close

bc3:
hello @enchev we have another PR for radzen notifications, there was no way to intercept clicking on x of a toast

in our case: 
1) x closes the toast, and marks it as closed on the server
2) opens a link rendered in the toasts

this PR makes that possbile, i again made some test snipped on my GIT https://github.com/bc3/RadzenShizzle


enchev:
Please check this please:
https://github.com/radzenhq/radzen-blazor/issues/1084

bc3:
@enchev not the same, there was no event to wire the x to

bc3:
@enchev I think i do understand the problem in bug #1084 : messages do not have a unique identifier, you can test it in my new sample for this case, then you see the close events that happen

PR #1090, opened 2023-07-25
Add ParseInput to DatePicker

### LINKED ISSUE #667: Datepicker enhancement: add option for expected input formats
Hello,

We use the datepicker like this:

```
<RadzenDatePicker 
    TValue="DateTime?" 
    Value="@_boundValue"
    DateFormat="dd MMM yyyy"
    Kind="DateTimeKind.Local"
    Change=HandleDatePickerChangeAsync
    DateRender="HandleDateRender"/>
```

We display the date in the input field with `dd MMM yyyy` as it is more unambigous for Europeans and Americans. 

However we have noticed some users, who choose to forego the picker and type the date will use the more familar `03/04/2023` format. We are based in the UK and people expect `03/04/2023` to equal `03 April 2023`, however Radzen parses that as `04 March 2023`

I can see this happens here, the `dd/MM/yyyy` format doesnt match `dd MMM yyyy` so it falls back to the default method:

https://github.com/radzenhq/radzen-blazor/blob/b5f414623e695d72de6fd79f21537d84aa5c7e2c/Radzen.Blazor/RadzenDatePicker.razor.cs#L515-L521

So the approach could be for the line 520 parse to use local standards (I thought `Kind="DateTimeKind.Local"` might do that)

But perhaps another solution could be an `InputFormat` param where you could list formats to try and parse with before falling back to the default on line 520

so something like: `InputFormat="dd MMM yyyy, dd/MM/yyyy"`

xs4free:
Add a custom ParseInput function to the DatePicker so applications can provide a custom parse method for input to match multiple exact formats.

This fixes #667.

### DIFF
--- a/Radzen.Blazor.Tests/DatePickerTests.cs
+++ b/Radzen.Blazor.Tests/DatePickerTests.cs
@@ -380,6 +380,71 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void DatePicker_Parses_Input_Using_DateFormat()
+        {
+            using var ctx = new TestContext();
+            ctx.JSInterop.Mode = JSRuntimeMode.Loose;
+
+            var component = ctx.RenderComponent<RadzenDatePicker<DateTime?>>();
+
+            var raised = false;
+            object newValue = null;
+
+            component.SetParametersAndRender(parameters =>
+            {
+                parameters.Add(p => p.DateFormat, "ddMM");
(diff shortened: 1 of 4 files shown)


enchev:
Thanks! Please add an example in DatePicker demos as well.

xs4free:
@enchev I added an example and a little bit of text to explain the use-case. Hope this helps.
Please let me know if you need anything else.

PR #1096, opened 2023-07-30
Add DataGrid Item Editing

TimChen44:
In many business scenarios, programs need to add, edit, and remove items in the DataGrid. This operation only requires adjustments to the data, does not require user editing, nor does it require reloading data, which is usually triggered by other operations.

The InsertRow and UpdateRow in the current DataGrid will enter row editing mode and are not applicable.

So I added three simple methods

InsertItem: Insert a row of data

UpdateItem: Updates a row of data. If the KeyProperty property is configured, it will be matched through the Key. The updated object data can come from a new object

DeleteItem: Remove a row of data

They only make adjustments to the current data

Other questions: Why are there many formats in the source code with some extra or missing spaces, such as after "if"

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -70,13 +70,13 @@ namespace Radzen.Blazor
             var view = AllowPaging ? PagedView : View;
             var top = request.Count;
 
-            if(top <= 0)
+            if (top <= 0)
             {
                 top = PageSize;
             }
 
             await InvokeLoadData(request.StartIndex, top);
-            
+
             var totalItemsCount = LoadData.HasDelegate ? Count : view.Count();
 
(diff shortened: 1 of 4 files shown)


enchev:
I'm afraid that I don't see the benefit of these new methods. Adding, removing and updating items in your collection assigned to Data does the same and since the DataGrid supports ObservableCollection will not even require reference of the component.

TimChen44:
There is no intuitive way to refresh the display after modifying Data

```csharp
    protected RadzenDataGrid<CompanyQueryDto> gridList;

    protected List<CompanyQueryDto> CompanyQueryDtos;

    public async Task OnAdd()
    {
        var dto = new CompanyQueryDto() { CompanyId = Guid.NewGuid() };
        CompanyQueryDtos.Add(dto);//Assuming to obtain another object that needs to be added

        // await InvokeAsync(StateHasChanged); //invalid
        gridList.Data = CompanyQueryDtos.AsQueryable();//The refresh method is not intuitive
    }
```

The example provided on the official website currently uses the Reload implementation after deleting data. There is no elegant method when I don't need to reload, so at least one Delete method is also missing

PR #1097, opened 2023-07-31
functionality is added so that when the filter is clicked a function is executed

bikotoru:
With the new template filter functionality it happens that if one has many filters, since the table has many columns, the site takes a long time to load, so I added a functionality so that every time the filter is clicked it can be call a custom function and thus achieve the following

```
 public async Task FuncionX(RadzenDataGridColumn<Comgesot.Models.comgesotdb.Ot> prueba)
        {
            if (prueba.Property.Equals("TipocargoOt.nombre"))
            {
                if(tipoCargoOt is null)
                {
                    tipoCargoOt = await comgesotdbService.GetTipocargoOt();

                }
            }
}
```

So when the filter is called, if I don't have the data, it will search the database, and I only bring the information I need and not all when loading the page

Example

```
            <RadzenDataGrid @ref="grid0" ColumnWidth="200px" AllowFiltering="true" FilterMode="FilterMode.Advanced" AllowPaging="true" AllowSorting="true" ShowPagingSummary="true" PageSizeOptions=@(new int[]{5, 10, 20, 30})
                               Data="@ot"  CustomFilterEvent="@FuncionX">

```

**when trying to run the unit tests with the code that is in MAIN they are not executed, even without modifying the source code**

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -132,7 +132,7 @@
                                 var columnIndex = visibleColumns.IndexOf(column);
                                 var sortableClass = AllowSorting && column.Sortable ? "rz-sortable-column" : "";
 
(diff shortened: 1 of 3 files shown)


bikotoru:
the error in the tests was not caused by my code, the code already came with this problem from the main branch

enchev:
In templates you know exactly when and what will happen since you've made the UI. I'm afraid that we cannot accept this.

bikotoru:
@enchev try that but the frontend is created from the start of the view and then hidden, so if I create a component with an afterrender , the afterrender is executed on site launch, not on hit of the filter button

PR #1111, opened 2023-08-11
Added demo datagrid save settings to url

k-Sacr:
Added a demo example of saving datagrid settings to url query using json serialization and base64 encoding.

Fix multi LoadData with settings:

Before:
chrome_2023-08-11_17-38-42

After:

If remove 
```
protected override async Task OnAfterRenderAsync(bool firstRender)
{
    if (firstRender || loaded)
    {
        await LoadStateAsync();

        if (loaded)
        {
            loaded = false;
            await Task.Yield();
            await grid.Reload();
            loaded = false;
        }
    }
}
```
and move LoadStateAsync() to OnInitialized/OnInitializedAsync/SetParametersAsync

chrome_2023-08-11_17-50-36

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -1970,7 +1970,7 @@ namespace Radzen.Blazor
             {
                 await InvokeLoadData(skip, PageSize);
             }
-            else
+            else if (settings == null)
             {
                 await base.ReloadOnFirstRender();
             }
--- a/RadzenBlazorDemos/Pages/DataGridSaveSettingUrlLoadData.razor
+++ b/RadzenBlazorDemos/Pages/DataGridSaveSettingUrlLoadData.razor
@@ -1,0 +1,167 @@
+﻿@using Radzen
+@using RadzenBlazorDemos.Data
+@using RadzenBlazorDemos.Models.Northwind
+@using Microsoft.EntityFrameworkCore
+@using RadzenBlazorDemos.Services
+@using Microsoft.JSInterop
+
+@using System.Text.Json
+@using System.Linq.Dynamic.Core
+@using System.Text.Json.Serialization
+@using System.Web
+
+@inject IJSRuntime JSRuntime
+@inject NavigationManager NavigationManager
+
+@inherits DbContextPage
+
+<p>This example shows how to save/load DataGrid state using Settings property when binding using LoadData event.</p>  
+<p>Settings is converted to json, encoded in base64 and passed to the current URL as a query parameter</p>
+<p>The state includes current page index, page size, groups and columns filter, sort, order, width and visibility.</p>
+<RadzenButton Click="@(args => Settings = null)" Text="Clear saved settings" Style="margin-bottom: 16px" />
+<RadzenButton Click="@(args => NavigationManager.NavigateTo(NavigationManager.Uri, true))" Text="Reload" Style="margin-bottom: 16px" />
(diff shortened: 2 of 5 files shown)


enchev:
Hi @k-Sacr,

Thanks for your pull request however in my opinion the example is too specific to your case and we will unable to merge and support it. Loading also state before initialization of the component will not restore the DataGrid state properly  - for example filtering will not be applied. I'm afraid that we cannot merge this.

k-Sacr:
Hi @enchev thanks for your feedback.

Is the ability to share a link to a page with a datagrid where all the necessary settings (filters / sorting / grouping) are set to other users so specific?
It seems to me that users share links very often.

> Loading also state before initialization of the component will not restore the DataGrid state properly - for example filtering will not be applied. I'm afraid that we cannot merge this.
Are you talking about code changes `else` to `else if (settings == null)` on ReloadOnFirstRender?
Can you please give an example where this is causing problems so that I can investigate further and find a solution?

Because if remove the extra loaddata call, it will **greatly speed up loading**, which will be especially noticeable with complex queries. 
It will also solve the problem of "flickering" when datagrid is shown without settings, and then displayed with settings, this is very noticeable on your demos.

enchev:
Here is what I see with this change in the DataGrid using the same demo with no changes:
image
image
```
09:19:51.55 LoadStateAsync
09:19:51.55 LoadData
09:19:53.58 LoadData
09:19:53.69 LoadStateAsync
09:19:53.77 LoadStateAsync
09:19:53.78 LoadData
09:19:55.64 LoadStateAsync
09:19:55.66 LoadData
```

and without:
image
```
09:25:40.99 LoadStateAsync
09:25:41.09 LoadData
09:25:43.42 LoadData
09:25:43.43 LoadStateAsync
09:25:46.63 LoadStateAsync
09:25:46.65 LoadData
09:25:48.59 LoadStateAsync
09:25:48.59 LoadData
```

k-Sacr:
**Exactly, this change will not affect the current demos/projects in any way.**
Because the call to LoadStateAsync() occurs in OnAfterRenderAsync when the ReloadOnFirstRender method has already been called.

But if the settings are stored in db/redis/url and call LoadStateAsync before FirstRender (for example OnInitialized/OnInitializedAsync/SetParametersAsync), then this will remove an extra critical request in ReloadOnFirstRender (where settings are not taken into account), and then Reload with settings will be called.

And to show how it works, I added a new demo page.

enchev:
> But if the settings are stored in db/redis/url and call LoadStateAsync before FirstRender (for example OnInitialized/OnInitializedAsync/SetParametersAsync), then this will remove an extra critical request in ReloadOnFirstRender (where settings are not taken into account), and then Reload with settings will be called.

Ok. I see your point here and I will push this change in the DataGrid. The demo however for loading/saving settings to a URL query string does not work for me at all and cannot be merged - both Reload button and copy/paste of the url in a new browse tab will not reload any state:
image
image

Furthermore after every DataGrid operation the page will scroll to top which is highly undesired and also when you navigate out to some different demo and return to this one the user will expect state to be loaded similar to the other save/load settings demos however this is not the case here since the demo will simply start from the beginning. 

Maybe an example on how to load/save state with OnInitialized() instead OnAfterRender() using some different technique like global service, etc. will be better.

enchev:
Here is the change as I promised:
https://github.com/radzenhq/radzen-blazor/commit/da5f7e27c4e793d2ca309ed18ed0a715e7d88e6c

k-Sacr:
Thx!

PR #1124, opened 2023-08-22
Add ItemSelected in AutoComplete

sobaka06243:
I need an event that invoked when an item is selected


enchev:
Thanks for you pull request however the same can be accomplished if you do not set TextProperty (use Template instead) and use @bind-Value:
```
@using RadzenBlazorDemos.Data
@using RadzenBlazorDemos.Models.Northwind

@inherits DbContextPage

<RadzenAutoComplete @bind-Value=@CompanyName Data=@customers>
    <Template>
        @context.CompanyName
    </Template>
</RadzenAutoComplete>

@code {
    string companyName;
    public string CompanyName
    {
        get
        {
            return companyName;
        }
        set
        {
            if (companyName != value)
            {
                companyName = value;

                // Add custom code on CompanyName change
            }
        }
    }

    IEnumerable<Customer> customers;

    protected override async Task OnInitializedAsync()
    {
        await base.OnInitializedAsync();

        customers = dbContext.Customers;
    }
}
```

PR #1150, opened 2023-09-11
Can modify groups parameter added

marekm294:
Added CanModifyGroups parameter which is used to disable adding/removing groups but keeping Group UI.
Method EndColumnDropToGroup refactored to avoid nested if.


enchev:
Hey @marekm294,

Thanks for you pull request however you can do the same at the moment using a simple CSS:
```
.rz-group-header
{
   pointer-events: none;
}
```

If you need to target specific DataGrid you can use the id attribute:
```
#[YOUR_DATAGRID_ID] > div.rz-group-header
{
   pointer-events: none;
}
```

PR #1152, opened 2023-09-12
Add Close Tab Button

TimChen44:
Provide a tab closure event. After defining the event, the tab displays a closure button, and the control itself does not perform any logic. The closure function is implemented by the user themselves.
tab

### DIFF
--- a/Radzen.Blazor/RadzenTabs.razor.cs
+++ b/Radzen.Blazor/RadzenTabs.razor.cs
@@ -72,6 +72,13 @@ namespace Radzen.Blazor
         List<RadzenTabsItem> tabs = new List<RadzenTabsItem>();
 
         /// <summary>
+        /// Gets or sets tab close button event 
+        /// </summary>
+        /// <value>The close button callback.</value>
+        [Parameter]
+        public EventCallback<RadzenTabsItem> TabCloseClick { get; set; }
+
+        /// <summary>
         /// Adds the tab.
         /// </summary>
         /// <param name="tab">The tab.</param>
--- a/Radzen.Blazor/RadzenTabs.razor.cs
+++ b/Radzen.Blazor/RadzenTabs.razor.cs
@@ -76,7 +76,13 @@ namespace Radzen.Blazor
         /// </summary>
         /// <value>The close button callback.</value>
         [Parameter]
-        public EventCallback<RadzenTabsItem> TabCloseClick { get; set; }
+        public EventCallback<RadzenTabsItem> TabClose { get; set; }
+
+        /// <summary>
+        /// Gets or sets tab close button display mode 
+        /// </summary>
+        [Parameter]
+        public TabCloseMode CloseMode { get; set; } = TabCloseMode.Selected;
 
         /// <summary>
         /// Adds the tab.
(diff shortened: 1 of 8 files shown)


Radzen.Blazor/RadzenTabs.razor.cs
```diff
@@ -71,6 +71,13 @@ public partial class RadzenTabs : RadzenComponent
 
         List<RadzenTabsItem> tabs = new List<RadzenTabsItem>();
 
+        /// <summary>
+        /// Gets or sets tab close button event 
+        /// </summary>
+        /// <value>The close button callback.</value>
+        [Parameter]
+        public EventCallback<RadzenTabsItem> TabCloseClick { get; set; }
```
akorchev, line 79: Please rename it to just `TabClose`

Radzen.Blazor/RadzenTabsItem.razor.cs
```diff
@@ -138,6 +138,14 @@ async Task OnClick()
             }
         }
 
+        async Task OnTabCloseClick()
+        {
+            if (Tabs.TabCloseClick.HasDelegate)
```
akorchev, line 143: There is no need for the if check here.

Radzen.Blazor/RadzenTabsItem.razor
```diff
@@ -16,6 +16,14 @@
             else
             {
                 <span class="rz-tabview-title">@Text</span>
+                @if (Tabs.TabCloseClick.HasDelegate)
+                {
+                    <span class="rz-ml-05">
+                        <RadzenButton Click="@OnTabCloseClick" class="rz-border-radius" Icon="close" Variant="Variant.Text"
+                                      ButtonStyle="ButtonStyle.Dark" Shade="Shade.Darker" Size="ButtonSize.ExtraSmall"
```
akorchev, line 23: Having both ButtonStyle.Dark and Shade.Darker may look odd in a dark theme. What if those are not set at all?

TimChen44:
Fixed the above issue and added the close button display mode.
The FREE style has been modified. I don't know how to modify the PRO style, so I didn't make any changes.

GIF 2023-9-13 22-43-26

akorchev:
We discussed this internally and decided we should implement something else - a Template property of RadzenTabsItem which would allow adding arbitrary content - buttons, badge etc. The template would display inside the `<a>` element. Would you like to try implementing that?

TimChen44:
The common scenarios here are the close button and badge, but there may be other buttons in some complex scenarios. Supporting templates is a good idea, and I can try to improve it.

akorchev:
It turns out there is a Template property already 🤣 Perhaps we can just make a demo which implements the closing with buttons that are defined in the template. Also the common UX pattern is to display the close button on hover only.

TimChen44:
One problem with existing templates is that users need to create everything themselves, which increases the complexity of user usage for commonly used scenarios.

TimChen44:
Added BadgeText attribute, I feel the effect is quite good.

akorchev:
Hi @TimChen44,

Thank you for your efforts however we decided to not continue with this PR. We prefer to use the Template for such purposes.

PR #1158, opened 2023-09-17
add IsShowArrow

TimChen44:
Sometimes the menu needs a look without a pull arrow, so a switch has been added to hide it.

However, when encountering hidden elements and expanding them on the style, it will cause a change in the menu width. I directly modify the style to worry about affecting other components. If this evaluation is valuable, please help to address the style issue.

GIF 2023-9-17 15-03-12


enchev:
Hey @TimChen44,

Maybe ShowArrow only is better name for such property.

TimChen44:
Renamed ShowArrow

PR #1168, opened 2023-09-26
Add `EditTemplate` to `RadzenDataGrid` 

### LINKED ISSUE #1167: [Feature Request] `RadzenDataGrid` - allow `EditTemplate`
I would like to update `RadzenDataGrid` such that it is possible to flow `EditContext` to an `EditTemplate` for a given row that is currently being edited, similar to the `Template` property.

Purpose: I currently have a row that has too much data to show in a single table row, so I put additional columns into the `<Template/>` property to be displayed when using the expand button. I would like to be able to edit these fields under the row's `EditContext` (using validation, etc.). To me, it would make sense to update RadzenDataGridRow.razor#L138 as follows:
```
@if (Grid.IsRowInEditMode(Item) && Grid.EditTemplate != null)
{
 <CascadingValue Value=@EditContext>
  @Grid.EditTemplate(Item)
 </CascadingValue>
}
else
{
  @Grid.Template(Item)
}
```

I can provide a PR for this if it would help.

viceroypenguin:
This PR adds an `EditTemplate` to the `RadzenDataGrid`, with a cascading `EditContext`, so that the child elements can participate in the same edit context as the parent row. 

NB: commit 7e4d037 is only added to prove that the change works as intended; the UI is not polished. I can remove this commit upon request.

Fixes #1167

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -1125,6 +1125,13 @@ namespace Radzen.Blazor
         public RenderFragment EmptyTemplate { get; set; }
 
         /// <summary>
+        /// Gets or sets the edit template.
+        /// </summary>
+        /// <value>The template.</value>
+        [Parameter]
+        public RenderFragment<TItem> EditTemplate { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


enchev:
See the example in the issue for details on how to implement this without any changes.

PR #1170, opened 2023-09-27
access to SelectedTab

bikotoru:
[#1165
](https://github.com/radzenhq/radzen-blazor/pull/1165)

The object of this change is to be able to select the tab, if you see the example based on the fact that the tab is tengo, you will have other information, so you don't have to load all the information on the page from one principle

```
            <RadzenTabs @ref=@TabsOT TabPosition="TabPosition.Top" RenderMode="TabRenderMode.Client" Change="ChangeTab">
                <Tabs>

                    <RadzenTabsItem Text="Información Vehículo">
</RadzenTabsItem>
</Tabs>
</RadzenTabs>

async Task ChangeTab()
{

  if (TabsOT is null) return;
 
  if (TabsOT.SelectedTab.Text.Equals("Actividades"))
  {
      

      if (listadoSolInsumos is null)
      {
          await SolicitudInsumosLoadData();
      }
      if (planesMantencionList is null || planesMantencionList.Count().Equals(0))
      {
          planesMantencionList = await comgesotdbService.GetPlanDeMantencion(new Query
          {
              Filter = "x=>x.versionid == @0",
              FilterParameters = new object[] { ot.Vehiculo.VersionAuto.id },
              Expand = "VersionAuto.ModeloAuto.MarcaVehiculo",
          });
      }

  }
}
```

In the example presented solo is to create tabs in a foreach, but that is where you hold tabs with distinct information and you can recognize which tab is active.

Is there any reason to access SelectedTab?

### DIFF
--- a/Radzen.Blazor/RadzenTabs.razor.cs
+++ b/Radzen.Blazor/RadzenTabs.razor.cs
@@ -105,7 +105,7 @@ namespace Radzen.Blazor
             }
         }
 
-        RadzenTabsItem SelectedTab
+        public RadzenTabsItem SelectedTab
         {
             get
             {


enchev:
Hey @bikotoru,

Do not open pull requests for already closed subject or we will ban you from our repo. Not to mention that build will fail since there is no comment on public property. This is last warning.

PR #1177, opened 2023-10-02
Added Filters to Query to be able to get filter data via grid.Query.Filters

k-Sacr:
Added to get data `Filters` list Grid and DataGrid through Query.
Need for custom processing of `Filters` list Grid/DataGrid example during export.

devenv_2023-10-02_17-14-32

`object[] FilterParameters` has been removed due to not being used in the code. Maybe it's need returning?

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -518,10 +518,10 @@ namespace Radzen
         /// <value>The filter.</value>
         public string Filter { get; set; }
         /// <summary>
-        /// Gets or sets the filter parameters.
+        /// Gets the filter expression as a collection of filter descriptors.
         /// </summary>
         /// <value>The filter parameters.</value>
-        public object[] FilterParameters { get; set; }
+        public IEnumerable<FilterDescriptor> Filters { get; set; }
         /// <summary>
         /// Gets or sets the order by.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Radzen.Blazor/Common.cs
```diff
@@ -518,10 +518,10 @@ public class Query
         /// <value>The filter.</value>
         public string Filter { get; set; }
         /// <summary>
-        /// Gets or sets the filter parameters.
+        /// Gets the filter expression as a collection of filter descriptors.
         /// </summary>
         /// <value>The filter parameters.</value>
-        public object[] FilterParameters { get; set; }
```
enchev, line 524: Not sure why you've removed FilterParameters however this will brake immediately all apps generated by Radzen IDE/Radzen Blazor Studio.

Radzen.Blazor/RadzenGrid.razor
enchev, line 1: No need to add anything here - this is an obsolete component.

Radzen.Blazor/Common.cs
k-Sacr: Return.
Oh sorry, the visual studio showed that the `FilterParameters` is not used anywhere. I won't delete anything else in future commits.

Radzen.Blazor/RadzenGrid.razor
k-Sacr: Fix

PR #1191, opened 2023-10-11
Added the ability to display smaller segmented pie charts

DanielSugdenPhoenix:
Added the ability to control the starting point and the size of the span for both Pie and Donut charts, functionality that could be useful for people wanting charts that don't span a full 360 degrees.

Tested across a full range of values for both the starting location and the size of the span, including checking the functionality of both the mouseover tooltip and the SeriesClick event of the chart

Screenshots:
image
image
image

### DIFF
--- a/Radzen.Blazor/RadzenDonutSeries.razor
+++ b/Radzen.Blazor/RadzenDonutSeries.razor
@@ -37,19 +37,22 @@
         var radius = CurrentRadius;
         var innerRadius = InnerRadius ?? radius / 2;
 
+        double startAngle = StartAngle.HasValue ? (double)StartAngle - 90 : -90;
+        double totalAngle = TotalAngle.HasValue ? (double)TotalAngle : 360;
+        double endAngle = startAngle + totalAngle;
+
         return
         @<g>
             <g class="@className">
                 @if (PositiveItems.Any())
                 {
                     var sum = PositiveItems.Sum(Value);
-                    double startAngle = -90;
 
                     @foreach (var data in PositiveItems)
                     {
                         var value = Value(data);
-                        var angle = (value / sum) * 360;
-                        var endAngle = startAngle + angle;
+                        var angle = (value / sum) * totalAngle;
+                        endAngle = startAngle + angle;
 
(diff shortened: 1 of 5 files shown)


DanielSugdenPhoenix:
Hi @akorchev , thanks for the quick reply

That is a good idea, the initial implementation was because I was trying to not impact the rest of the code as I didn't know the full scope, however you are right in that removing it can help cleanup the code, changes made

akorchev:
The last thing that bothers me is this comment:

```
      /// <summary>
         /// Gets or sets the start angle of the donut, defaults to -90.
         /// </summary>
         [Parameter]
         public double StartAngle { get; set; } = 0;
```

We say that the initial value is -90 however it is 0. This doesn't seem right.

akorchev:
@DanielSugdenPhoenix I have updated my comment not sure if you saw it - the whole StartAngle is invalid at the moment. It is not -90 and it is not 0 - it is 90 visually.

akorchev:
I think the confusion comes from the way rendering was implemented - clockwise instead of counter clocwkise (which is how degrees increment). I am experimenting with something and will push my suggestions shortly.

DanielSugdenPhoenix:
@akorchev The comment said -90 as a default as that is what I originally had it as, I saw the setting of the startAngle in the PieSeries render fragment being set to -90, however when I saw it was 0 in most other places I decided to just offset the render fragment by -90 instead

At the moment the StartAngle being 0 would result in the starting slice being top center like so: 
image
Is this not an acceptable implementation? A quick google search of pie chart shows that its not uncommon for the starting point of a chart to be here

akorchev:
This is indeed the default start position of a pie chart. However that is not 0 degrees. It is 90 degrees.

akorchev:
I am not sure your fix would cover all cases. Have to test with a few edge cases - empty data and data with one item.

DanielSugdenPhoenix:
I appreciate you taking the time to consider it, as it is quite a big change. I am currently using this fork as a dependency in another project so I will continue to use it to see if there any unintended consequences

PR #1200, opened 2023-10-18
partial validation is supported

marekm294:
Before that, RadzenStep could not be moved to next if Validation failed. This feature allows to validate only part of model in each step.

Limitations - Model is still validated and if the user moves to the next step, he will see validation messages. It would be nice to improve in the future somehow.

### DIFF
--- a/Radzen.Blazor/RadzenSteps.razor.cs
+++ b/Radzen.Blazor/RadzenSteps.razor.cs
@@ -268,27 +268,52 @@ namespace Radzen.Blazor
             return SelectedIndex == index;
         }
 
-        internal async System.Threading.Tasks.Task SelectStep(RadzenStepsItem step, bool raiseChange = false)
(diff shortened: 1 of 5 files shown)


akorchev:
Hi @marekm294,

Thank you for the pull request. Would you mind also contributing a demo that shows how to use this new property?

marekm294:
I try to find time to do that ASAP

marekm294:
I hope something simple as that can be good for now. It can be always improved in the future

akorchev:
Hi @marekm294,

I've just checked the demo however it behaves in the exact same way if the new property is not set. Can you clarify what the purpose of that property is? RadzenSteps renders only the current step at a time so components in other steps that are required do not register for validation at all. Only the components in the current step are validated.

akorchev:
I am closing this request as no further feedback was provided and I couldn't understand the reasoning behind it.

PR #1205, opened 2023-10-18
Adding aria-label and title attributes to pager buttons.

DaveBagler:
This PR adds aria-label and title attribute parameters to the pagers' first, previous, next, and last buttons. It also adds title and aria-label format parameters for the number buttons.


enchev:
Thanks for the pull request @DaveBagler! It will be good if all aria-label properties use Aria in the name. For example: `FirstPageAriaLabel`. This applies to the other pull request as well.

DaveBagler:
@enchev, I've updated this PR and #1204 to use `...AriaLabel` rather than `...Label`.

PR #1208, opened 2023-10-19
Adding title and aria-label attributes to step buttons.

DaveBagler:
This PR adds aria-label and title attribute parameters to the steps' previous and next buttons. It also adds title and aria-label attributes to the tab list for steps.


DaveBagler:
@enchev, this PR uses `AriaLabel` and not `Label`.

Issue #1201, opened 2023-10-18
ContextMenuItem - Add separator item type


RemiCELLARD:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when you have to use RenderFragment instead of `ContextMenuItem` class to create a separator in ContextMenu (example : for DataGrid Cell)

**Describe the solution you'd like**
- Add in class `ContextMenuItem` a new getter/setter (boolean) `IsSeparator`
- When you render this, add separator `<hr/>` in your ContextMenu.

**Describe alternatives you've considered**
You can use RenderFragment (see https://blazor.radzen.com/docs/api/Radzen.ContextMenuOptions.html#Radzen_ContextMenuOptions_ChildContent ) but for my case, it's better to use object.

**Additional context**
No more element

enchev:
You can already place any custom content for the context menu including RadzenMenuItem and a plain `hr` HTML element:
![image](https://github.com/radzenhq/radzen-blazor/assets/5804953/2d9d333d-6685-4ac0-832a-c2e79f00dd0a)
![image](https://github.com/radzenhq/radzen-blazor/assets/5804953/8acb57a9-99d2-4348-9d2e-0289501fde14)
https://blazor.radzen.com/contextmenu

dave-gcv:
Trying to understand why this was closed/not considered ?

Like the OP, I am also look for an "easy" programatic way to add a separator to the context menu.

Having to use the RenderFragment way just isn't convenient enough, is prone to error and inefficient especaiily if you have a large number of items in the menu or the items that appear in the menu are dynamic based on various factors.

Having the "IsSeparator" flag or even support for Class property (applied to the <li> element) just seems logical ?

PR #1214, opened 2023-10-23
Create context menu separators without RenderFragment

gauss-lvs-dev:
I created a ContextMenuSeparator class that extends ContextMenuItem and modified RadzenContextMenu to distinguish between those to types. I feels a litte bit more "natural" to me than the following approaches.

Another approach would be to create a simple "IsSeparator" property which RadzenContextMenu checks against. Yet another approach could be to create the properties "InsertSeparatorAfter" or "InsertSeparatorBefore".

I am happy to contribute any approach, if you want me to.

Someone created an issue 5 days ago (#1201)

### DIFF
--- a/Radzen.Blazor/ContextMenuService.cs
+++ b/Radzen.Blazor/ContextMenuService.cs
@@ -215,4 +215,9 @@ namespace Radzen
         /// <value>The image.</value>
         public string Image { get; set; }
     }
+
+    /// <summary>
+    /// Class ContextMenuSeparator.
+    /// </summary>
+    public class ContextMenuSeparator : ContextMenuItem { }
 }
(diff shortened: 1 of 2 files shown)


enchev:
Thanks for the pull request @gauss-lvs-dev however you can already place any custom content for the context menu including RadzenMenuItem and a plain `hr` HTML element:
image
image
https://blazor.radzen.com/contextmenu

gauss-lvs-dev:
The only downside is that you have to concatenate strings in order to apply more complex rules (e.g. access rights).



Still, thanks for taking the time to read the pull request.

RemiCELLARD:
Hi @gauss-lvs-dev + @enchev ,  

Thanks for your PR and feedback.
It's what i have in mind when i created this Issue ( #1201 ). 
The aim goal is to use objects to create your menu instead of using a RenderFragment (to have an alternative by using only object classes).
Too bad it won't be included in a future update :-(

Thanks for your help.

Issue #1016, opened 2023-06-13
Unable to associate RadzenLabel with RadzenDropDown


ikappas:
**Describe the bug**

RadzenLabel association with RadzenDropDown does not work using "for" and "id" attributes.
Usage of the "for" and "id" attribute was suggested #912

**To Reproduce**

``` razor
<RadzenLabel for="the_drop_down_id" Text="TheDropDownLabel" />
<RadzenDropDown id="the_drop_down_id" TValue="int?" />
```

**Expected behavior**
1. The emitted label should be associated with the input element. 
2. Clicking on the Label should focus and expand the dropdown.

**Additional context**

The "id" attribute, when set on the RadzenDropDown component is emitted on the wrapper div instead of the input element which has no effect and breaks the association. The id attribute should be set on the nested input element.

```html
<label for="the_drop_down_id" class="rz-label" id="3FRAQulmE0" _bl_121="">TheDropDownLabel</label>
<div class="rz-dropdown rz-state-empty" onmousedown="Radzen.activeElement = null" tabindex="0" id="dropdown" _bl_64="">
    <div class="rz-helper-hidden-accessible">
        <input id="the_drop_down_id" aria-haspopup="listbox" readonly="" type="text" tabindex="-1" aria-label="">
    </div>
    <label class="rz-dropdown-label rz-inputtext ">&nbsp;</label>
    <div class="rz-dropdown-trigger  rz-corner-right">
        <span class="rz-dropdown-trigger-icon  rzi rzi-chevron-down"></span>
    </div>
...
</div>
```

ikappas:
So how would someone associate the label with the dropdown?

akorchev:
It is not supported at the moment.

0xor1:
@ikappas all examples of Label and input component usage I have seen dont use `for="name"` as in your example, they look like:
```
            <RadzenLabel Component="IsParallel" Text="Is Parallel"/>
            <RadzenCheckBox Name="IsParallel" @bind-Value="@_model.IsParallel"/>
```
it is `Component` that you should use to associate a lable with an form input component. afaik.

DaveBagler:
@akorchev, I was looking into the `RadzenLabel`'s `Component` parameter and how it fills the `for` attribute of the `label` tag.

[...]

PR #1232, opened 2023-11-02
Issue #1016 - Adds input id parameter for RadzenDropDowns.

DaveBagler:
Adds an `InputId` parameter to set the id attribute of the **RadzenDropDown**'s input element. See #1016.


enchev:
Thanks for the pull request @DaveBagler! I've added InputAttributes similar to other Radzen Blazor form components that can be used to set the id to the input (or any other attribute).

DaveBagler:
Thanks, @enchev. I saw that, and I really like the general solution for input attributes.

PR #1240, opened 2023-11-09
Update RadzenMenuItem.razor.cs

Neoxflo:
Hello Radzen Team,

I tried to implement IsDisabled, but I am not sure if I have done this right, I don't understand the insides very well, but I tried, let me know if this is the right way to do this.

Yours,
Florian


enchev:
Hi @Neoxflo,

Thanks for the pull request however the property should be named Disabled similar to other components and there should be disabled state CSS again similar to other components.

PR #1242, opened 2023-11-13
Menu item disabled property added

enchev:
Close https://github.com/radzenhq/radzen-blazor/pull/1240


Neoxflo:
Hello Enchev,

I tried it also today to get my code working, I was on the completely wrong path, I guess

So I thought getOpenEvents tells the Menu which Icon is clickable and which is not, but now I realize it tells the menu bar before the menus are loaded. So my code can't work because I am checking the current state and therefore onclick is already set to clickable? 

Or is this guess also wrong?

Thanks for adding it, very cool to see how you did it

enchev:
Hi @Neoxflo,

There are two click event related methods. The first one will raise the Blazor component Click event and the second one will handle open/close of the menus with plain JavaScript.

Neoxflo:
Ahhh okay now I see, thanks for implementing this, I am pretty sure I didn't get this working by myself.

PR #1244, opened 2023-11-14
Add PanelTogglerPosition + Demo

msdevcode:
Add PanelTogglerPosition to RadzenPanel

Included a demo:
grafik

Wasnt sure how to deal with the css `rz-panel-titlebar`. I override the `justify-content`, if toggler position `PanelTogglerPosition.Left` is set.

The `<a>` tag of the togglers could probably be a component too. With a component there is no need of copy paste the `<a>` tag as i did. I wasnt able to intgrate them.

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -1826,6 +1826,21 @@ namespace Radzen
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


enchev:
Thanks for the pull request @msdevcode however in my opinion this can be done very easily with simple CSS:
image

PR #1257, opened 2023-11-22
New Component Radzen Date Range Picker

SyedShahriyarAli:
The **Date Range Picker** is an important component mostly used in reports or summaries. I have been using two RadzenDatePickers for date ranges for quite some time but I didn't like it much. Like the image below:

image

So I added a new component RadzenDateRangePicker, it has two calendars one for the Start Date and another for the End Date. It returns a class object **RadzenDateRange** having two attributes _StartDate_ and _EndDate_. Most of the properties are similar to the **RadzenDatePicker** component.

I also added an example in '**_DateRangePickerPage_**' for better understanding.

image

### DIFF
--- a/Radzen.Blazor/RadzenDateRangePicker.razor
+++ b/Radzen.Blazor/RadzenDateRangePicker.razor
@@ -1,0 +1,158 @@
+﻿@using Radzen
+@using Radzen.Blazor.Rendering
+@using Microsoft.AspNetCore.Components.Forms
+@using System.Linq.Expressions
+@using System.Globalization
+@using Microsoft.JSInterop
+
+@inherits RadzenComponent
+@implements IRadzenFormComponent
+
+@if (Visible)
+{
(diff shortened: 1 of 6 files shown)


enchev:
Hi @SyedShahriyarAli,

Thank you very much for your pull request! We finally found time to check carefully your implementation since we will need to support it when merged and we've found that your code differs from our ideas for date range picker. For example we should have some internal component for rendering of the calendar in both current picker and the date range picker instead copy of the entire rendering in every component. The UI also should be different in our opinion, for example:
image
We are also unsure what's the best API for date range picker, is it value with array of dates or two properties for start/end date, etc. and because of these we've decided to not accept this pull request but rather create it in-house at some point.

angerry02:
Waiting to implement this component. I hope you include it in the next updates. Thank you, Radzen team.

franklupo:
Is there a plan for this component?

PR #1268, opened 2023-12-04
Various changes to Radzen Blazor Chart column and bar series

paulo-rico:
Initial requirement: I needed a chart that displayed only profit / loss or increase / decrease. I needed it to show one color for positive and one color for negative as opposed to different colors for each series. I did begin by creating a new series type, but the reality is all I needed was a different way of coloring. So I created a Property and an Enum for selecting a 'ColorMode', and the Render method of each column and bar acts on this appropriately.
However, as much as I would prefer to change one thing a time, it did give rise to other, related items (fixes and features). 
I've created a demo app at [radzen.delaci.co.uk](http://radzen.delaci.co.uk/) to show these changes. Along the top are numbered controls to highlight each item listed below.
(1) Sets the enum of the ColorMode to 'Series' or 'PositiveNegative' to show the different color render modes.
(2) Uses custom Fills as opposed to using the theme.
(3) The way data labels were rendered didn't seem right. Whether the value was positive or negative, it rendered above (column) or to the right (bar) of the rendered bar. It now renders above if positive, below if negative and on the line if 0 (or base value, see next items).
(4) Gives the ability to set the base of the chart to something other than zero.
(5) Sets whether the base label is shown on the chart.
(6) Because of the differences in rendering based on Chart Size, Steps and Font Size, this setting gives 'breathing space' for the base label so adjacent labels can be hidden.

Below the example column and bar series, there is a copy of the "Radzen Blazor Chart column series" from the Radzen demo site to show that, on the face of it, there are no breaking changes.

Hope this is of interest and look forward to your feedback.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/AxisBase.cs
+++ b/Radzen.Blazor/AxisBase.cs
@@ -103,6 +103,13 @@ namespace Radzen.Blazor
         public double ChartBaseValue { get; set; } = 0.0;
 
         /// <summary>
+        /// Gets or sets the buffer distance from the ChartBaseValue to adjacent chart axis labels and lines. Any that fall within will not be displayed. A value of 0.0 (default) hides the base label if not part of the calculated axis grid.
+        /// </summary>
+        /// <value>Buffer distance from the base label to adjacent chart axis labels and lines.</value>
+        [Parameter]
+        public double ChartBaseLabelSpace { get; set; } = 0.0;
+
+        /// <summary>
         /// Gets or sets a value indicating whether this <see cref="AxisBase"/> is visible.
         /// </summary>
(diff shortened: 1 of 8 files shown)


akorchev:
Hi @paulo-rico,

Thank you for the effort! Could you please split this in a smaller PR? I find it hard to review this way.

paulo-rico:
Hi @akorchev 

I'm sure there will be a way to reverse and redo stuff on the same PR, but I'm not that experienced with Git.
I'm going to save my changes local, close this PR and start again with a fresh fork and PR.
I will then make the code changes for each feature/fix and commit as I go along.
Some of the code changes are small per feature, so I think I'll send them first before the bigger items.

See you on another thread.

Paul

PR #1269, opened 2023-12-05
Edit DataFilter; add param to show time in datefilter, separate code from razor component

johanpolak:


enchev:
Hi @johanpolak,

Thanks for your pull request however you can do that even now by declaring DatePicker in FilterTemplate:
```
<FilterTemplate>
    <RadzenDatePicker @bind-Value="@context.FilterValue" Style="width:100%" ShowTime=false />
</FilterTemplate>
```

johanpolak:
Hi @enchev,

You're welcome. I was aware of the possibility to use the <FilterTemplate>, but I didn't know that I could bind the FilterValue directly within the FilterTemplate. So thanks for the advice! That will make the use of this component a lot easier. And thank you guys for all the great work you're doing!

PR #1270, opened 2023-12-05
Change to position of Data Labels for Column and Bar charts

paulo-rico:
This is the first in a series of changes.

Test site - [radzen.delaci.co.uk](http://radzen.delaci.co.uk/)

While using negative and zero data, the labels were always positioned above (column) or to the right (bar) of the bar.

This change places the label above, level or below for column charts and left, level or right for bar charts.


paulo-rico:
Hi @akorchev 
Not sure what's happened with the check failures. None of them seem relate to any changes I've made.

Regards
Paul

enchev:
Hey @paulo-rico,

I've broke the tests however they are now fixed.

Issue #996, opened 2023-05-31
Radzen chart set the color of the chart series based on the value of the current element


kElaadwy:
I am writing to request a feature for the Radzen chart. I would like to be able to set the color of the chart series based on the value of the current element.

For example, I would like to be able to set the color of the chart series to green if the value of the current element is greater than 200, and red if the value of the current element is less than 200.

I believe that this would be a valuable addition to the Radzen chart.

Thank you for your time and consideration.

paulo-rico:
Hi @kElaadwy 
This has now been implemented. See [https://blazor.radzen.com/column-chart](https://blazor.radzen.com/column-chart) for example chart and source code.

PR #1272, opened 2023-12-06
Add new Bar and Column Series property 'ColorMode'

paulo-rico:
This property will enable the chart to color the bars by Series (as it is now) or by PositiveNegative, based on whether value is above or below zero.

Demo - [razden.delaci.co.uk](razden.delaci.co.uk)


paulo-rico:
Hi @akorchev 

It's not as straight forward. When PickColor returns null, it uses the Series position in it's collection to get the color -
` var className = $"rz-column-series rz-series-{Chart.Series.IndexOf(this)}";`
Whereas, PositiveNegative is based on individual values rather than series.

I'll take a look at the code and see if it can be cleaned/reduced. (I was of the same thought, initially, that only the PickColor needed adjusting.

PS - I don't know what happens to the code when pushed. In Visual Studio Git Changes panel, it shows the changes exactly as I would want to see them. In the Render methods, it showed the `if (ColorMode == ColorMode.Series)` as an add, and the accompanying `else` block as an add, and it was straight forward. When pushed and viewed on Git, it's mixing up where the add and deletes occured because of similar code. If you have any suggestions on this (other than dropping Visual Studio :) ), they would be much appreciated.

Regards

Paul

IMMEDIATE EDIT - Leave this with me. In talking it through, I think there may be a solution.

paulo-rico:
Existing check in Render method

`
                // code

                if (ChartSeriesColorMode == ChartSeriesColorMode.Series)
                {
                    fill = PickColor(Items.IndexOf(data), Fills, Fill);
                    stroke = PickColor(Items.IndexOf(data), Strokes, Stroke);
                    className = $"rz-bar-series rz-series-{Chart.Series.IndexOf(this)}";
                }
                else
                {
                    fill = PickColor(val <= 0 ? 0 : 1, Fills, Fill);
                    stroke = PickColor(val <= 0 ? 0 : 1, Strokes, Stroke);
                    className = val <= 0 ? $"rz-bar-series rz-series-0" : $"rz-bar-series rz-series-1";
                }
`

In moving the ChartSeriesColorMode check to PickColor, I would also have to also pass in the 'Value' of the data in order to select the appropriate color if PositiveNegative. OK so far.

But there would still be a need for the check in the Render method in order to get the className correctly based on the ChartSeriesColorMode.

Would you prefer that a new method was implemented to get the ColorClass, passing in ChartSeriesColorMode, BaseClass, Value and Series Index? If so, would you put this in CartesianSeries alongside PickColor?

Regards

Paul

akorchev:
Maybe it would be easier just to extract the index 

```
var index = ChartSeriesColorMode == ChartSeriesColorMode.Series ? Items.IndexOf(data) : val <= 0 ? 0 : 1;
```

Then the rest could be:

```
fill = PickColor(index, Fills, Fill);
stroke = PickColor(index, Strokes, Stroke);
className = $"rz-bar-series rz-series-{index}";
```

akorchev:
There seems to be a major change in the rendering implementation too. All items should be in the same `<g>`. Your change seems to render every item in its own `<g>`. This is a breaking change.

paulo-rico:
Ok, I've changed the svg so it all wraps in a <g> tag as per original code and applied the className to the Path object.

Not sure what Color property you mean for refactoring.
I have remove some duplicate code that sets the PositiveNegative value in the Render and moved it to a variable, i.,e.

```
var sign = val <= 0 ? 0 : 1;
var colorIndex = ChartSeriesColorMode == ChartSeriesColorMode.Series ? Items.IndexOf(data) : sign;
var classIndex = ChartSeriesColorMode == ChartSeriesColorMode.Series ? Chart.Series.IndexOf(this) : sign;

```

Is this what you meant?

akorchev:
> Not sure what Color property you mean for refactoring.

I mean `public override string Color` - it also uses similar code to find the index.

paulo-rico:
Sorry @akorchev Ignore last Push.

akorchev:
Also make sure there are no breaking changes in the rendering. The `<g>` element should have a class name not the `<path>`.

paulo-rico:
Just seen comment. Ignore commit. The class name needs to be set if there are no custom colors (Fills) set. When this is by series, it is OK that it set before looping through the items as it's the same for all. But with PositiveNegative, its by each data item.

What about we set the class in the <g> element as before and have a variable to hold a new className for PositiveNegative mode?

We could then interrogate that variable and either output a Path with the new class (PositiveNegative) or output a Path without any class definition, as before.

paulo-rico:
Just going to throw some ideas out there to add to your thought process.

The thing with the PositiveNegative view is that the color selection isn't really dynamic. You must have two colors. So, do we have two special properties for this (PositiveColor?, NegativeColor?) or do we insist that there are at least two colors defined in the Fills property when ColorMode = ColorModePositiveNegative? This way, we can ignore any class manipulation and not incorporate breaking changes.

I'll give it further thought myself.

Regards and stay well

Paul

paulo-rico:
Inspired by this issue - [https://github.com/radzenhq/radzen-blazor/issues/996](https://github.com/radzenhq/radzen-blazor/issues/996), I think there is better, more generic way to achieve this.

Instead of dealing with PositiveNegative, we instead set a list of color ranges -

```
List<SeriesColorRange> colorrange = new List<SeriesColorRange>()
{
    new SeriesColorRange() {Min = -10000000, Max = 0, Color = "red"},
    new SeriesColorRange() {Min = 0, Max = 255000, Color = "orange"},
    new SeriesColorRange() {Min = 255000, Max = 10000000, Color = "#06cd06"},
};
``` 

This is a parameter on the Series

In the Render methods, we include this in the PickColor call along with Value and ColorMode -

`
var fill = PickColor(Items.IndexOf(data), Fills, Fill, ColorMode, FillRange, val);
`

Then the PickColor method becomes -

```
        protected string PickColor(int index, IEnumerable<string> colors, string defaultValue = null, ChartSeriesColorMode colorMode = ChartSeriesColorMode.Series, IList<SeriesColorRange> colorRange = null, double value = 0.0)
        {
            if (colorMode == ChartSeriesColorMode.Series)
            {
                if (!(colors == null || !colors.Any()))
                {
                    return colors.ElementAt(index % colors.Count());
                }
            }
            else
            {               
                if(colorRange != null)
                {
                    var result = colorRange.Where(r => r.Min <= value && r.Max > value).FirstOrDefault<SeriesColorRange>();
                    return result != null ? result.Color : defaultValue;
                }
            }

            return defaultValue;
        }

```

This accomodates the ability to set the color of a data item to a specific color dependant on it's value. This means all the developer has to do to achieve PositiveNegative is - 

```
    List<SeriesColorRange> posnegrange = new List<SeriesColorRange>()
    {
        new SeriesColorRange() {Min = double.MinValue, Max = 0, Color = "red"},
        new SeriesColorRange() {Min = 0, Max = double.MaxValue, Color = "blue"},
    };

```
Running demo here - [radzen.delaci.co.uk](http://radzen.delaci.co.uk/)

This has reduced the changes to Series Render methods to adding the extra parameters to the PickColor call.

I'll commit the changes now. I don't know if you can just check this commit against the original files. I've obviously made a lot of changes back and forth.

Regards

Paul

akorchev:
I like the idea of color ranges a lot! It seems this makes the ColorMode option obsolete doesn't it? If ColorRanges is set then use it. Otherwise don't. The default value would be `null` anyway. It would also make sense for other chart series at some point. What do you think?

paulo-rico:
Excellent!

I can make the changes necessary to the existing Pull/Commit chain. There have been a great deal of back and forths with this. Are you able to just see the latest changes against the original file? Or would it be easier for you if I created a fresh fork and made the changes on a fresh copy (there aren't that many changes now, so it's as easy for me either way)?

I'll also make the same changes to the StackedBar and StackedColumn series. I can't see that there are any other series that would benefit from FillRange.

Paul

akorchev:
> I can make the changes necessary to the existing Pull/Commit chain.

I think making the changes to the current commit chain would suffice. The diff should always compare with the current master branch.

>  I can't see that there are any other series that would benefit from FillRange.

Probably no indeed. Only series that render a single "item" per value need this change.

Radzen.Blazor/CartesianSeries.cs
```diff
@@ -728,14 +728,24 @@ public virtual IEnumerable<ChartDataLabel> GetDataLabels(double offsetX, double
...
-        protected string PickColor(int index, IEnumerable<string> colors, string defaultValue = null)
+        /// <param name="colorRange">The color range value.</param>
+        /// <param name="value">The value of the item.</param>
+        protected string PickColor(int index, IEnumerable<string> colors, string defaultValue = null, IList<SeriesColorRange> colorRange = null, double value = 0.0)
         {
-            if (colors == null || !colors.Any())
+            if (colorRange != null)
             {
-                return defaultValue;
+                var result = colorRange.Where(r => r.Min <= value && r.Max > value).FirstOrDefault<SeriesColorRange>();
```
akorchev, line 737: This should probably be `r => r.Min <= value && r.Max >= value` to include Max

Radzen.Blazor/CartesianSeries.cs
```diff
@@ -728,14 +728,24 @@ public virtual IEnumerable<ChartDataLabel> GetDataLabels(double offsetX, double
...
+                return result != null ? result.Color : defaultValue;
+            }
+            else
+            {
+                if (!(colors == null || !colors.Any()))
```
akorchev, line 742: I find this harder to grok that the previous code `if (colors == null || !colors.Any())`

Radzen.Blazor/RadzenBarSeries.razor
```diff
@@ -24,16 +24,17 @@
 
         var barHeight = BandHeight;
         var height = barHeight / barSeries.Count() - padding + padding / barSeries.Count();;
-        var className = $"rz-bar-series rz-series-{Chart.Series.IndexOf(this)}";
+        var className = $"rz-column-series rz-series-{Chart.Series.IndexOf(this)}";
```
akorchev, line 27: This should be `rz-bar-series`

Radzen.Blazor/RadzenBarSeries.razor
```diff
@@ -47,10 +48,10 @@
                 {
                     path = $"M {x0.ToInvariantString()} {y.ToInvariantString()} L {(x+radius).ToInvariantString()} {y.ToInvariantString()} A {r} {r} 0 0 0 {x.ToInvariantString()} {(y+radius).ToInvariantString()} L {x.ToInvariantString()} {(y+height-radius).ToInvariantString()} A {r} {r} 0 0 0 {(x+radius).ToInvariantString()} {(y + height).ToInvariantString()} L {x0.ToInvariantString()} {(y+height).ToInvariantString()} Z";
                 }
-                var fill = PickColor(Items.IndexOf(data), Fills, Fill);
-                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke);
+                var fill = PickColor(Items.IndexOf(data), Fills, Fill, FillRange, val);
+                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke, StrokeRange, val);
 
-                <Path @key="@path" D="@path" Stroke="@stroke" StrokeWidth="@StrokeWidth" Fill="@fill" LineType="@LineType" Style="@style" />
+                <Path class="@className" @key="@path" D="@path" Stroke="@stroke" StrokeWidth="@StrokeWidth" Fill="@fill" LineType="@LineType" Style="@style" />
```
akorchev, line 54: There shouldn't be a class here.

Radzen.Blazor/RadzenColumnSeries.razor
```diff
@@ -43,10 +44,10 @@
...
                 }
-                var fill = PickColor(Items.IndexOf(data), Fills, Fill);
-                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke);
-
-                <Path @key="@path" D="@path" Stroke="@stroke" StrokeWidth="@StrokeWidth" Fill="@fill" LineType="@LineType" Style="@style" />
+                var fill = PickColor(Items.IndexOf(data), Fills, Fill, FillRange, val);
+                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke, StrokeRange, val);
+                
+                <Path class="@className" @key="@path" D="@path" Stroke="@stroke" StrokeWidth="@StrokeWidth" Fill="@fill" LineType="@LineType" Style="@style" />
```
akorchev, line 50: Should not set class here as well.

akorchev:
Last things (I promise)!

1. There seems to be trailing whitespace in some of the files (visible as big green rectangles in the diff)
2. Please use `value` instead of `val` (unless it isn't a conflict - if it is use `seriesValue`)

Radzen.Blazor/RadzenBarSeries.razor
```diff
@@ -26,14 +26,15 @@
...
+        return
         @<g class="@className">
             @foreach(var data in Items)
             {
                 var y = category(data) - barHeight / 2 + index * height + index * padding;
                 var x = value(data);
+                var itemValue = Value(data);                
                 var radius = Chart.BarOptions.Radius;
-
+
```
akorchev, line 37: Trailing whitespace

Radzen.Blazor/RadzenColumnSeries.razor
```diff
@@ -43,11 +44,11 @@
                 {
                     path = $"M {x.ToInvariantString()} {y0.ToInvariantString()} L {(x+width).ToInvariantString()} {y0.ToInvariantString()} L {(x+width).ToInvariantString()} {(y-radius).ToInvariantString()} A {radius.ToInvariantString()} {radius.ToInvariantString()} 0 0 1 {(x + width - radius).ToInvariantString()} {y.ToInvariantString()} L {(x + radius).ToInvariantString()} {y.ToInvariantString()} A {radius.ToInvariantString()} {radius.ToInvariantString()} 0 0 1 {x.ToInvariantString()} {(y-radius).ToInvariantString()} L {x.ToInvariantString()} {y0.ToInvariantString()} Z";
                 }
-                var fill = PickColor(Items.IndexOf(data), Fills, Fill);
-                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke);
-
+                var fill = PickColor(Items.IndexOf(data), Fills, Fill, FillRange, itemValue);
+                var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke, StrokeRange, itemValue);
+
```
akorchev, line 49: Trailing whitespace

Radzen.Blazor/RadzenColumnSeries.razor.cs
```diff
@@ -26,7 +26,7 @@ public partial class RadzenColumnSeries<TItem> : CartesianSeries<TItem>, IChartC
         [Parameter]
         public IEnumerable<string> Fills { get; set; }
 
-        /// <summary>
+        /// <summary>
```
akorchev, line 29: Trailing whitespace.

paulo-rico:
Hi @akorchev 

A bit of help please, if you will.

In VS2022 git panel, the diffs only show the differences from my last commit, so I'm not seeing some of these things you mention. Without starting afresh, is there a way that I can view diffs for my current source against the original source as opposed to my last commit?

Cheers

Paul

akorchev:
To be honest I am not sure. IDEs usually show the difference since the last commit which is expected. You would have to constantly squash your commits and compare to master if you want that. I am looking at the "Files changed" tab in Github which shows the final result.

paulo-rico:
I've been looking into this, and it appears there is a way to check current source against original source. 

https://devblogs.microsoft.com/visualstudio/introducing-new-git-features-to-visual-studio-2022/#compare-branches

I'll use this in future if any subsequent updates have taken a few turns/commits. Hopefully save us both some time :)

Thanks for your patience.

Paul

akorchev:
> Thanks for your patience @akorchev.

Don't mention it. Thank you for your cooperation as well! Would you mind creating a demo which shows this feature? You can add it to the [column chart demos](https://github.com/radzenhq/radzen-blazor/pull/1272).

paulo-rico:
Ok, I've added it to the 'Radzen Blazor Chart column series' page in RadzenBlazorDemos project. A cursory glance at RadzenExample component suggests that the source code is read from git. That would be why it ran, but 'Edit Source' showed the existing source.

PR #1274, opened 2023-12-06
Add ability to configure AutoComplete input type

sand-head:
We're using the AutoComplete component as a search box in an internal application, and users have recently requested a clear button. Since setting `type="search"` would be the most straightforward solution for us, we needed a way to be able to configure the underlying input type.


sand-head:
Sure thing! I'll defer to you on property naming, let me get that changed really quick.

PR #1282, opened 2023-12-11
Added support for translation of DisplayAttribute Description properties in RadzenDataGrid Enum filters

decimvs:
In the RadzenDataGrid Enum filters there isn't the possibilty to intercept the render of the text in the options select menu. This causes that the texts in the DisplayAttribute Descriptions don't get translated if you use a different I18N stack than .NET's default. The aim of this feature is to offer the possibility to the developer to define a function that will be called for each text defined in a DisplayAttribute in addition to the Enum.ToString(). Radzen.Blazor is well I18N oriented, but in my opinion was lacking of this kind of translation posibility. 
This principle can be exported to other kinds of components with a similar problematic. 

The principle is simple: we define a property in the RadzenDataGrid component that brings the possibility to define a function `Func<string, string>`. In the RadzenDataGridHeaderCell component the function is passed to the `Extensions.EnumAsKeyValuePair()` function that passes it to the `Extensions.GetDisplayDescription()` that finally will call this function if it is not null. It lets the developer to define a custom function, in my case `_T(string text)` that returns a string containing the translation of the passed `text`.

### DIFF
--- a/Radzen.Blazor/Extensions.cs
+++ b/Radzen.Blazor/Extensions.cs
@@ -18,20 +18,24 @@ namespace Radzen.Blazor
         /// <summary>
         /// Gets enum description.
         /// </summary>
-        public static string GetDisplayDescription(this Enum enumValue)
+        public static string GetDisplayDescription(this Enum enumValue, Func<string, string> translationFunction = null)
         {
             var enumValueAsString = enumValue.ToString();
             var val = enumValue.GetType().GetMember(enumValueAsString).FirstOrDefault();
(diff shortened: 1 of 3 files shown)


enchev:
Hi @decimvs,

Thanks for your pull request! Just to make sure, you cannot use the approach described in this thread?
https://forum.radzen.com/t/blazor-datagrid-filtering-on-enums-translation/13242/7

decimvs:
Hi @enchev,

No, I can't use this approach because I'm not using regular RESX files for I18N.  In general I can translate all the texts in Radzen components except for enums in filters because enum names are treated in internal procedures. The aim is to open a way to bring a method that is able to translate the enum texts on the fly when the filter's combobox is generated.

PR #1283, opened 2023-12-12
Removed code that breaks the logic of dynamic data loading

RTSTALER:
Provided that the Data changes, the entire value field is cleared, ultimately this breaks the possibility of implementing auto completion with dynamic data loading using the MVVC pattern, where it is impossible to load data from the context inside the ViewModel using LoadData


enchev:
Hey @RTSTALER,

Removing this code will brake immediately components like the DropDown since the View will not be re-initialized on Data change. We cannot accept this change.

Issue #1229, opened 2023-10-31
Markers cannot be dynamically turned off for Radzen Line Chart


isak1080:
**Describe the bug**
I have a radzen line chart with multiple line series. Some of these have markers and some don't. I also have buttons in the UI that show/hide these lines at the user's will.

As long as I don't use markers I don't have a problem, but as soon as one of the series have markers the rendering of the charts gets "weird". I'm not 100% sure what happens, but it seems that it applies markers to all line series (even though they don't have them - I've debugged and verified I don't output marker in the razor). 

**To Reproduce**
I was able to modify the sample over at https://blazor.radzen.com/line-chart to show what I think is the same issue. 
Here is the full sample, just paste it to the demo page: [radzen-markers-bug.razor.txt](https://github.com/radzenhq/radzen-blazor/files/13214459/radzen-markers-bug.razor.txt)

It adds a checkbox:
```
<RadzenCheckBox @bind-Value=@showMarkers2019 Name="markers"></RadzenCheckBox>
<RadzenLabel Text="Show 2019 Markers" For="markers" />
```
and uses the boolean to show/hide the markers:
```
<RadzenLineSeries Smooth="@smooth" Data="@revenue2019" CategoryProperty="Date" Title="2019" LineType="LineType.Dashed" ValueProperty="Revenue">                    
    @if (showMarkers2019)
    {
        <RadzenMarkers MarkerType="MarkerType.Square" />
    }
    <RadzenSeriesDataLabels Visible="@showDataLabels" />
</RadzenLineSeries>
```

If the value of `showMarkers2019` is false at start, checking the checkbox will trigger a refresh and markers will appear. They won't go away by unchecking. If the value starts `true`, the markers appear but cannot be turned off.

**Expected behavior**
Markers appear and disappear with the toggle of the checkbox

**Screenshots**
![image](https://github.com/radzenhq/radzen-blazor/assets/105219900/ac00d4a9-ded9-4d9c-8b5c-a132626dac21)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Edge,  [e.g. chrome, safari]
 - Version: All I've tried in the last 6 months

isak1080:
Thanks. This does indeed work around the problem for now.

Are there any performance considerations with this workaround? I have lines with hundreds of data points each, so would this produce invisible markers for all of them?

PR #1287, opened 2023-12-14
New property ShowMarkers on Line Series

paulo-rico:
Resolution for issue https://github.com/radzenhq/radzen-blazor/issues/1229


akorchev:
Hi @paulo-rico,

Markers are controlled by their own tag. This functionality should be implemented as a Visible property of `<RadzenMarkers />`

PR #1306, opened 2024-01-03
Fix null exception when expanding tree node

stlufred:
If:
- Checkboxes enabled
- CheckedValues is null

Fix works for me, but I'm not 100% sure it works for all cases supported by RadzenTree, so don't be shy on improving my fix :) .

### DIFF
--- a/Radzen.Blazor/RadzenTree.razor.cs
+++ b/Radzen.Blazor/RadzenTree.razor.cs
@@ -322,13 +322,16 @@ namespace Radzen.Blazor
 
                 if (AllowCheckBoxes && AllowCheckChildren && args.Children.Data != null)
                 {
-                    if (CheckedValues != null && CheckedValues.Contains(item.Value))
+                    if (CheckedValues != null)
                     {
-                        await SetCheckedValues(CheckedValues.Union(args.Children.Data.Cast<object>().Except(UncheckedValues)));
-                    }
-                    else
-                    {
-                        await SetCheckedValues(CheckedValues.Except(args.Children.Data.Cast<object>()));
+                        if (CheckedValues.Contains(item.Value))
+                        {
+                            await SetCheckedValues(CheckedValues.Union(args.Children.Data.Cast<object>().Except(UncheckedValues)));
+                        }
+                        else
+                        {
+                            await SetCheckedValues(CheckedValues.Except(args.Children.Data.Cast<object>()));
+                        }
                     }
                 }
             }


enchev:
Hi @stlufred,

Thanks for the pull request however I'm not sure how to reproduce the exception. In our demos the variable bound to CheckedValues is null:
image

stlufred:
You also need to use "dynamic" loading using data binding (like in the "File and directory browser" example).  In fact you need to have both samples combined.  In fact it will probably be easier for you to add check boxes to the "File and directory browser" example.  Then you should get the exception when expanding a node and keeping the selection empty.

enchev:
Here is what I did however no exceptions:
image
image

stlufred:
I guess there must be something else in my code.  I'll try to create a sample that reproduce the problem and I'll update the PR with it,

stlufred:
1.  Modify `RadzenBlazorDemos\Pages\TreeFileSystem.razor` to:
```
@using System.IO

<div class="container-fluid">
    <div class="row my-5">
        <div class="col-lg-6 offset-lg-3">
            <RadzenCard>
                <RadzenTree Data=@entries Expand=@LoadFiles Style="width: 100%; height: 300px"
                            AllowCheckBoxes="true" @bind-CheckedValues=@CheckedValues>
                    <RadzenTreeLevel Text=@GetTextForNode Template=@FileOrFolderTemplate />
                </RadzenTree>
            </RadzenCard>
        </div>
    </div>
</div>

@code {
    IEnumerable<string> entries;
    IEnumerable<object> CheckedValues;

    protected override void OnInitialized()
    {
        entries = Directory.GetDirectories(Directory.GetCurrentDirectory())
                           .Where(entry =>
                           {
                               var name = Path.GetFileName(entry);

                               return !name.StartsWith(".") && name != "bin" && name != "obj";
                           });

    }

    void LoadFiles(TreeExpandEventArgs args)
    {
        var directory = args.Value as string;

        args.Children.Data = Directory.EnumerateFileSystemEntries(directory);
        args.Children.Text = GetTextForNode;
        args.Children.HasChildren = (path) => Directory.Exists((string)path);
        args.Children.Template = FileOrFolderTemplate;
    }

    string GetTextForNode(object data)
    {
        return Path.GetFileName((string)data);
    }

    RenderFragment<RadzenTreeItem> FileOrFolderTemplate = (context) => builder =>
    {
        string path = context.Value as string;
        bool isDirectory = Directory.Exists(path);

        builder.OpenComponent<RadzenIcon>(0);
        builder.AddAttribute(1, nameof(RadzenIcon.Icon), isDirectory ? "folder" : "insert_drive_file");
        builder.CloseComponent();
        builder.AddContent(3, context.Text);
    };
}
```
2. Compile
3. Navigate to `https://localhost:44398/tree-file-system`
4. An item of the tree without checking any checkbox beforehand

image
image

Issue #731, opened 2022-12-05
RadzenNumeric has precision issues


sofiageo:
**Describe the bug**
RadzenNumeric has precision issues because of how it tries to convert a double value.

**To Reproduce**
Use a RadzenNumeric with a double value of 266273.2949999999 and a Format of N2 - or C2.

**Expected behavior**
Expect the value to be rounded to 266273.29 - Instead the value will be rounded to 266273.30

**Additional context**
You can see the difference in values when you run the following demo:

```
double num1 = 266273.294999999;
double num2 = 266273.2949999999;
decimal decimal1 = (decimal)Convert.ChangeType(num1, typeof(decimal));
decimal decimal2 = (decimal)Convert.ChangeType(num2, typeof(decimal));
decimal decimal3 = decimal.Parse(num2.ToString("G17"));
Console.WriteLine(decimal1.ToString("N2"));
Console.WriteLine(decimal2.ToString("N2"));
Console.WriteLine(decimal3.ToString("N2"));
```

I've got the idea for using toString("G17") from this [SO question](https://stackoverflow.com/a/46568334)

akorchev:
Not sure if we can do anything in this regard without breaking the existing implementation. We are currently using Convert.ChangeType.

PR #1315, opened 2024-01-10
Support for custom Numeric Types

pianomanjh:
In my project, we have various custom numeric types, like `Money`, `Point`, etc, that we would like to use with RadzenNumeric.
RadzenNumeric's support for these kind of types is problematic, as it uses `Convert.ChangeType` behind the scenes, such that it only works with primitive types.

This adds support for custom numeric types by:
1. Adding `TypeConverter` support to/from decimal for TValue
2. Adding support for `IComparable<decimal>`, in the case the type implements it
3. Update flow of `InternalValueChanged` to remove unnecessary ChangeValue conversion attempts.  This allows limited support for custom numeric types, even if they don't have a TypeConverter (as long as `ConvertValue` is set)

This would really be a great help!

### DIFF
--- a/Radzen.Blazor.Tests/Dollars.cs
+++ b/Radzen.Blazor.Tests/Dollars.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.ComponentModel;
+using System.Globalization;
+
+namespace Radzen.Blazor.Tests;
+
+[TypeConverter(typeof(DollarsTypeConverter))]
+public readonly record struct Dollars(decimal Amount) : IComparable<decimal>
+{
+    public int CompareTo(decimal other)
+    {
+        return Amount.CompareTo(other);
+    }
+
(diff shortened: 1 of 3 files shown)


enchev:
Excellent submission @pianomanjh! Well done & thank you very much!

sofiageo:
By the way this still doesn't fix #731

pianomanjh:
@sofiageo Correct, this does not change existing behavior for primitive types (breaking change).  It does allow you to now use a type that handles conversion to/from decimal explicitly, avoiding the issue with `Convert.ChangeType` (which just casts double to decimal).

sofiageo:
I know, it just seemed a good timing to remind there is still an issue :) According to dotnet runtime developers, Convert.ChangeType is legacy code and a [code smell](https://github.com/dotnet/runtime/issues/73074). Maybe the Radzen implementation can be changed some time in the future.

pianomanjh:
Yeah, it would be great to get it updated, at least for net 7/8 where they introduced a slew of new interfaces for numerics that are fast and extensible.

PR #1317, opened 2024-01-10
Can change added

marekm294:
Image situation: ask user if he really wants to change step with dialog. Or maybe ask user, if he wants to change current step without saving changes made in this step.
This PR allows both

### DIFF
--- a/Radzen.Blazor/RadzenSteps.razor.cs
+++ b/Radzen.Blazor/RadzenSteps.razor.cs
@@ -164,7 +164,7 @@ namespace Radzen.Blazor
 
         /// <summary>
         /// A callback that will be invoked when the user tries to change the step.
-        /// Invoke the <see cref="RadzenStepsCanChangeEventArgs.PreventDefault"/> method to prevent this change.
+        /// Invoke the <see cref="StepsCanChangeEventArgs.PreventDefault"/> method to prevent this change.
         /// </summary>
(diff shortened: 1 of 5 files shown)


akorchev:
Hi @marekm294,

What's the purpose of this PR? It has no description provided about its intent.

marekm294:
@akorchev  - Done :)

marekm294:
demo added

akorchev:
Thank you for the demo. I understand the intent now. Unfortunately this API is quite strange considering the existing events. There also are two events `CanChange` and `CanChangeAsync`. Could you try implementing it with a single `EventCallback` similar to [this](https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenScheduler.razor.cs#L187)? It would then support both async and sync handlers.

```
<RadzenTabs CanChange=@CanChange>
</RadzenTabs>

@code {
    async Task CanChange(RadzenTabsCanChangeEventArgs args)
    {
        if (condition)
        {
            args.PreventDefault();
        }
    }
} 
```

marekm294:
I hope I understood you

akorchev:
That's exactly what I meant! Just please rename`PreventStepChange()` to `PreventDefault()` (for consistency with other components) and `IsStepChangePrevented` to `IsDefaultPrevented`.

PR #1318, opened 2024-01-10
Support for IFormattable

pianomanjh:
- As a follow-up to PR #1315 , to add support for `IFormattable`, for scenarios where custom numeric types would like to handle their own formatting.
- Adjusts the code in `SetParametersAsync` to also support custom numeric types, and provides a fast path to avoid conversion if the  value to be set is already of the correct type.
- Provides demo examples of using custom numeric types with `RadzenNumeric`.

### DIFF
--- a/Radzen.Blazor.Tests/NumericTests.cs
+++ b/Radzen.Blazor.Tests/NumericTests.cs
@@ -479,5 +479,24 @@ namespace Radzen.Blazor.Tests
             Assert.Contains($" value=\"{maxDollars.ToString()}\"", component.Markup);
             Assert.Equal(component.Instance.Value, maxDollars);
         }
+
+        [Fact]
+        public void Numeric_Supports_IFormattable()
+        {
+            using var ctx = new TestContext();
+
+            var valueToTest = new Temperature(60.23m);
+            const string format = "F";
+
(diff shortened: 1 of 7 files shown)


enchev:
Thanks for the pull request @pianomanjh! I've noticed that on the Dollars type demo in some cases the format is not applied:
image

Maybe when the value is outside of the Min/Max range.

pianomanjh:
@enchev thanks for the review.  I was able to reproduce this, by simply entering the same value twice.  On the second render, the formatting is not applied.  (regardless of min/max)  To be clear, the Dollars type in this case only overrides `ToString` (does not set or leverage `Format`).
Though I'm not completely sure of the root cause, it appears that the formatting is applied during the render cycle, because `FormattedValue` is called.  So I believe Blazor thinks that the dom hasn't changed, so does not update. 
My last commit has a fix for this, please let me know if you think it is safe (the commit message has detail).

enchev:
Hi @pianomanjh,

Unfortunately I was still able to break it by entering same value twice:
image
image

pianomanjh:
I am not able to reproduce, do you have any further details?

enchev:
Looks like the solution was not properly rebuilt - I'm unable to reproduce it now and I'll merge the pull request.

Issue #1320, opened 2024-01-14
RadzenDataGrid should cascade the RadzenDataGrid class to the entire component


djatkinsrealis:
I would like to add custom components to the HeaderTemplate (or elsewhere inside RadzenDatagrid), that can reference the RadzenDataGrid to obtain information. e.g. a custom header showing data stats

Using @ref, between disconnected components, is more difficult as they appear to only be set after initial render so are often null.

At the moment the DataGrid is cascaded to the initial Columns collection only. It would be useful if that cascading value was wrapped around the entire DataGrid.

Splitting the CascadingValue to lines 8 (before all HTML) and 435 (just before @code) does not seems to adversely affect the tests or the RadZen DataGrid demos.

While there are alternatives, this seems to be the simplest way to extend RadzenDataGrid without putting custom code in components that contain a RadzenDataGrid.

PR #1325, opened 2024-01-14
#1320 move DataGrid cascading value to surround entire component

djatkinsrealis:
Move the RadzenDataGrid cascading value to surround the entire component, allowing it to be picked up in customer header/footer etc. Splitting the existing CascadingValue to lines 8 (before all HTML) and approx line 435 just before @code.

Only a 2 line change - but the indenting affected looks horrendous :)

Trivial: Also added brackets to avoid missing colon style warning

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -5,387 +5,123 @@
 @typeparam TItem
 @inherits PagedDataBoundComponent<TItem>
 
-@if (Columns != null)
-{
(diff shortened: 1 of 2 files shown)


enchev:
Hi @djatkinsrealis,

Please keep the original formatting since at the moment one can hardly read what exactly is changed.

djatkinsrealis:
@enchev Is there any way to get GitHub to ignore whitespace changes in PRs, like DevOps does? If not, I will have to redo the change and abort this one as the slightest change in VS is triggering large whitespace differences.

Issue #1228, opened 2023-10-31
Performance DataGrid slow loading/large download data


k-Sacr:
_On the project, we decided to replace the old table with Radzen DataGrid connected through the render server. To our surprise, the Radzen DataGrid was loading 7-10 times slower! It was discovered that the old table only loaded 300KB, while the Radzen table loaded 3MB. That's a 10-fold difference for the same visible data._

I conducted several experiments with RadzenBlazorDemos.Server.

1 - Loading the /datagrid-dynamic page with 100 records, without passing data to <Template>, meaning an empty table:

![chrome_2023-10-31_12-23-34](https://github.com/radzenhq/radzen-blazor/assets/18440948/67d8638e-3791-4ec5-836b-ba3a11caec63)

So, to display an **empty table, it loads 7.9MB!** I think this needs to be optimized.

2 - Regular loading of the /datagrid-dynamic page with 100 records:

![devenv_2023-10-30_21-00-00](https://github.com/radzenhq/radzen-blazor/assets/18440948/bd008816-1b43-4268-82a0-f7fb122cc1af)

8.6MB, but in reality, the HTML markup weighs 1.78MB.

I removed the use of RadzenDataGridCell in RadzenDataGridRow, changing it to similar (for this table) HTML code:

![chrome_2023-10-30_20-56-54](https://github.com/radzenhq/radzen-blazor/assets/18440948/229e3976-2bd1-4b68-b19e-19942cbf4784)

As can be seen from the screenshot, the table now weighs 3.5MB, which **is 2.45 times smaller**! Using RadzenDataGridCell costs 5.1MB
This is a huge difference and savings, without any visible changes in the table.

Unfortunately, I don't have enough knowledge of the Radzen code to perform optimization. 
I want to raise this obvious issue and hope that you can dedicate some time to it.

enchev:
Here is what I see on the same demo even without paging when sorting the DataGrid by FirstName:
![image](https://github.com/radzenhq/radzen-blazor/assets/5804953/e3414c8f-4c0a-4208-8a1a-641b4f92ae9f)

k-Sacr:
> Here is what I see on the same demo even without paging 

Set PageSize="100" in RadzenDataGrid. With this value the problem is more obvious.

With RadzenDataGridCell , with PageSize="100":

https://github.com/radzenhq/radzen-blazor/assets/18440948/15cba15f-9acc-4430-89e1-f9e04eab7fec


[...]

PR #1328, opened 2024-01-15
#1320 Cascade RadzenDataGrid class to all descendants

djatkinsrealis:
Move the RadzenDataGrid cascading value to surround the entire component, allowing it to be picked up in customer header/footer etc. Splitting the existing CascadingValue to lines 8 (before all HTML) and approx line 435 just before the @code

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor
+++ b/Radzen.Blazor/RadzenDataGrid.razor
@@ -5,11 +5,10 @@
 @typeparam TItem
 @inherits PagedDataBoundComponent<TItem>
 
+<CascadingValue Value=this>
 @if (Columns != null)
 {
-    <CascadingValue Value=this>
         @Columns
-    </CascadingValue>
 }
 
 @if (Visible)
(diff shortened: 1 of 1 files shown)


enchev:
Before: 2.6 MB binary message
image
image

After change 2.9 MB binary message
image
image

In my opinion this in unacceptable.

djatkinsrealis:
@enchev Can we get more info on how you test the above, so we can do equivalent checks before submissions?

enchev:
You can easily test the size of the Blazor binary message using our Dynamic data binding demo.
Just set PageSize=100 and press Run in the demo code editor - messages can be observed in your browser network tab for the Blazor WebSocket connection.

enchev:
Check this issue also:
https://github.com/radzenhq/radzen-blazor/issues/1228

djatkinsrealis:
@enchev Hi again, I was very puzzled that a cascade could affect message size so much, so I have been testing that 2-line PR change _only_, in release builds, against the same master branch it was branched from. The only difference being the split of the cascading value to wrap the entire component (and the PageSize=100 in both of course). In both cases I am seeing 2.7Mb message size consistently. Screen shots below (master then branch):
image
image

enchev:
I was testing in Debug, good catch! Now after I've tested the same in release I received 2.6 MB for both branches - I proceed and merge your change.

Issue #1338, opened 2024-01-19
Pagination - Pagination should reset to first page when column sorting occurs


djatkinsrealis:
It is standard behavior e.g. such as in eBay or Amazon website dashboards, to reset to the first page when sorting (or filtering) changes. For instance if you sort by a value column, you expect to see the lowest or highest value displayed first, whereas now you see whatever value happens to occur at the top of current page number you were on. I note Telerik's Grid has the same issue.

Currently sorting does not alter pagination (specifically the current page number), leaving you on a page that does not relate to your sorting preference. 

**We propose resetting the current page number to the first page number, when the sorting changes.**

Note: I _attempted_ to implement this change, simply using conditional awaited calls to the 2 pagers _async_ FirstPage() methods, and testing the change within the DataGrid demos, but it resulted in the pages not responding to paging, or sorting, or even the demo menus, after a few selections. Seemed like a deadlock, so instead I made the changes as show below (not await and reset the current page). I could use some advice on the correct way to implement such a change, but will post a PR of this change for reference/testing.

        /// <summary>
        /// Orders the DataGrid by property name.
        /// </summary>
        /// <param name="property">The property name.</param>
        public void OrderBy(string property)
        {
            var p = IsOData() ? property.Replace('.', '/') : PropertyAccess.GetProperty(property);

            var column = allColumns.ToList().Where(c => c.GetSortProperty() == property).FirstOrDefault();

            if (column != null)
            {
                SetColumnSortOrder(column);
                topPager?.FirstPage();
                bottomPager?.FirstPage();
                CurrentPage = 0;
                Sort.InvokeAsync(new DataGridColumnSortEventArgs<TItem>() { Column = column, SortOrder = column.GetSortOrder() });
                SaveSettings();
            }

            if (LoadData.HasDelegate && IsVirtualizationAllowed())
            {
                Data = null;
    #if NET5_0_OR_GREATER
                ResetLoadData();
    #endif
            }

            InvokeAsync(ReloadInternal);
        }

enchev:
Hi @djatkinsrealis,

Changing this behavior will be huge breaking change for a lot of Radzen.Blazor users which is highly undesirable. You can easily achieve the same using the provided API:
```
<RadzenStack Orientation="Orientation.Horizontal" Gap="0.5rem" AlignItems="AlignItems.Center" Class="rz-p-4 rz-border-radius-1" Style="border: var(--rz-grid-cell-border);">
[...]

PR #1340, opened 2024-01-19
#1338 Reset pagination to first page, when column sorting occurs, retaining default behavior.

djatkinsrealis:
Reset pagination to first page, when column sorting occurs.

Based on feedback to prior commit, I have added a `GotoFirstPageOnSort` property to conditionally enable this behavior and retain the current behavior by default.
 
This avoids having to add code to every usage of RadzenDataGrid.


enchev:
Hey @djatkinsrealis,

I'm afraid that this is not the proper place to add such code, should be here:
image

djatkinsrealis:
@enchev Thanks. That is the advice I was looking for. Will update PR shortly 👍

djatkinsrealis:
@enchev Code moved as suggested. Works really well. PR updated. Thanks 👍

Issue #1345, opened 2024-01-25
RadzenDataGrid displays incorrectly ordered data when using async LoadData


egans146:
**Describe the bug**
This issue arises when using asynchronous LoadData callbacks and AllowSorting=true.
In the example below, the data displayed is not ordered according to the selected sort order :

```razor
<RadzenDataGrid @ref=grid AllowVirtualization=true
                LoadData="OnGridLoadData"
                Data="gridItems"
                Count=gridItemCount
                Style="height: 30rem;"
                AllowSorting=true>
    <Columns>
        <RadzenDataGridColumn TItem="Item" Property="@nameof(Item.Value)" Title="Value" />
    </Columns>
</RadzenDataGrid>

<RadzenButton Click="() => grid.RefreshDataAsync()">Refresh</RadzenButton>

@code {
    class Item(int value) { public int Value { get; } = value; };

    RadzenDataGrid<Item> grid;

    int gridItemCount = 1000;

    IReadOnlyList<Item> itemSource;

    protected override void OnInitialized()
    {
        var items = Enumerable.Range(0, gridItemCount).Select(e => new Item(e)).ToArray();
        Random.Shared.Shuffle(items);
        itemSource = items;
    }

    IEnumerable<Item> gridItems;

    async Task OnGridLoadData(LoadDataArgs args)
    {
        await Task.Delay(Random.Shared.Next(60, 110));

        IEnumerable<Item> query = itemSource;

        var sortOrder = args.Sorts?.FirstOrDefault()?.SortOrder;

        if (sortOrder != null)
        {
            if (sortOrder == SortOrder.Ascending)
                query = query.OrderBy(e => e.Value);
            else
                query = query.OrderByDescending(e => e.Value);
        }

        gridItems = query.Skip(args.Skip ?? 0).Take(args.Top ?? 10).ToArray();
        StateHasChanged();
    }
}
```

![Animation](https://github.com/radzenhq/radzen-blazor/assets/6442594/964f546f-3743-4e1c-b5a5-655a6861e0d8)

**Expected behavior**
RadzenDataGrid should display data sorted by the current sort order.

**Desktop:**
 - OS: Windows
 - Browser: Brave
 - Version 4.24.0

PR #1346, opened 2024-01-26
Change RadzenDataGrid OnSort to async

egans146:
Fix for RadzenDataGrid displays incorrectly ordered data when using async LoadData #1345
OnSort function has been changed to an async function similar to OnFilter


enchev:
Thanks @egans146! I completely missed that!

PR #1348, opened 2024-01-29
enforce columnSize min width

nielsNocore:
changed startColumnResize to enforce the min width property of a column in JavaScript

th min-width is working in chromium based browsers, but not in Firefox and safari (possibly other browsers as wel)
this change will enforce the min width (if set) in the JavaScript resize method. to make min width working.

### DIFF
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -1699,14 +1699,14 @@ window.Radzen = {
           },
           mouseMoveHandler: function (e) {
               if (Radzen[el]) {
-                  var widhtFloat = (Radzen[el].width - (Radzen[el].clientX - e.clientX));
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/wwwroot/Radzen.Blazor.js
```diff
@@ -1699,7 +1699,15 @@ window.Radzen = {
           },
           mouseMoveHandler: function (e) {
               if (Radzen[el]) {
-                  var width = (Radzen[el].width - (Radzen[el].clientX - e.clientX)) + 'px';
+                  var widhtFloat = (Radzen[el].width - (Radzen[el].clientX - e.clientX));
```
enchev, line 1702: Just a small typo - should be `width`

nielsNocore: fixed it in my last commit. thanks for the review.

PR #1354, opened 2024-02-04
DataGrid Column Custom Command Filter

paulo-rico:
Rather than skipping over, the Custom Filter Operator type now takes it's value from a new property CustomFilterCommand and adds that value to whereList for processing.

This enables the developer to write whatever dynamic linq filter they require without predefined functions / restrictions.

The changes in Common.cs do not pertain to the Custom Command Filter. They were documentation errors discovered while writing this.

Because of changes when wrapping an `if () {...}` block in Queryable.cs, lines 1010 to 1114, I have not reformatted the code which should, by rights, be tabbed along, as it was showing all these lines as deleted and then added in my git file changes. They made need shifting over when code has been checked so it is formatted correctly.

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -1669,15 +1669,15 @@ namespace Radzen
         /// </summary>
         DoesNotContain,
         /// <summary>
-        /// Satisfied if the current value is null.
+        /// Satisfied if the current value is in the specified value.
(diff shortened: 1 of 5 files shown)


enchev:
Thanks @paulo-rico! Maybe better name for this property will be `CustomFilterExpression` instead `CustomFilterCommand`.

PR #1366, opened 2024-02-08
Scheduler changes to set a start month for year views

paulo-rico:
As per forum query - https://forum.radzen.com/t/scheduler-yearplanner-not-solar-year/16368

Also, in YearPlanner and YearTimeline, the slots that are at the beginning and end of each row are there for aesthetics only, to fill out the rows to the edges. They are not "real" slots. Although we remove the Click event from these slots, they were still reactive to style modifications in the SlotRender event. This has also been resolved.

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -297,6 +297,61 @@ namespace Radzen
     /// <summary>
     /// Html editor mode (Rendered or Raw). Also used for toolbar buttons to enable/disable according to mode.
     /// </summary>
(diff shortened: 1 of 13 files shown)


akorchev:
Hi @paulo-rico,

This property should not be at RadzenScheduler level as it doesn't work for the other views. Move it to the views instead.

akorchev:
I suggest another rename - the enum from ScheduleStartMonth to just Month and the property from ScheduleStartMonth to just StartMonth.

PR #1368, opened 2024-02-08
Create SECURITY.md

polajenko:


enchev:
Hey @polajenko,

We don't have separate workflow for security related issues and we don't need this template.

PR #1371, opened 2024-02-13
Update RadzenAccordion.razor

zanderi:
Adding an href allows the browser to use it's default behavior such as mouse cursor on hover. This helps with Accessibility and indicates to users it is a clickable/interactive item. 

Setting it to javascript:void(0) disables the default navigation behavior.

### DIFF
--- a/Radzen.Blazor/RadzenAccordion.razor
+++ b/Radzen.Blazor/RadzenAccordion.razor
@@ -18,7 +18,7 @@
             continue;
 
         <div @ref="@item.Element" id="@item.GetItemId()" @attributes="item.Attributes" class="@item.GetItemCssClass()" style="@item.Style">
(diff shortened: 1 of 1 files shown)


zanderi:
image

enchev:
Thanks @zanderi however we removed such hrefs from all components exactly because of accessibility issues. Such link is reported by all validators:
https://copyprogramming.com/howto/web-accessibility-redundant-link-adjacent-links-go-to-the-same-url-for-a-tag-a-href-javascript-void-0-gt

zanderi:
@enchev I can see the issue you're referring to in the article. From experience I have had strange behavior in browsers and mobile devices from omittance of the href. Would you be open to possibly using <button> instead of an anchor tag? This would still allow keyboard navigation, screen reader accessibility, and additionally add rollover states.

enchev:
We are working on accessibility and keyboard navigation improvements for all components  - check keyboard-navigation branch for reference.

PR #1375, opened 2024-02-14
chore: performance and readability improvement

marekm294:
There is no point in iterating the list twice.
I also reverted the ifs and used pattern matching.

You can close this PR if you don't find it useful, but I had to debug this code and had to make it more readable for me, so I decided to post it now that I have these changes.


enchev:
Thanks @marekm294 however I’m afraid that we can’t accept pull request with such drastic changes - frankly it’s impossible to tell at the moment where the code is changed and where only the formatting.

PR #1378, opened 2024-02-15
Appointment drag and drop functionality for Scheduler component.

V45370:
I have added drag and drop functionality for DayView, WeekView and MonthView in the Scheduler component. 
There is an issue when building the Radzen.Blazor project for NET5.0 and NETSTANDARD2.1, but couldn't figure out why. However it builds successfully for above NET versions.
Поздрави, Васил :)

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -436,13 +436,27 @@ namespace Radzen.Blazor
                 "rz-listbox-item ";
         }
 
-        bool preventKeyDown = false;
+        bool preventKeyDown = true;
         int focusedIndex = -1;
         int focusedCellIndex = 0;
 
         internal string GridId()
         {
             return GetId();
+        }
+
+        async Task FocusRow(string key)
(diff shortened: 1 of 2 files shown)


akorchev:
Hi @V45370,

Thank you for the pull request!

I have the following requests/comments:

1. The event should not be per view - it should be exposed at RadzenScheduler level.
2. The event arguments should be SchedulerAppointmentMoveEventArgs as the other scheduler related event argument classes.
3. Currently one can do drag and drop even if it is not enabled which looks misleading. Probably the drag and drop attributes should be set only if the user has handled the AppointmentMove event.
4. Does it work in Timeline and Planner views?
5. It doesn't seem to work on a mobile device. I tested on iOS (iPhone and iPad) but drag-and-drop didn't start at all. Should be supported on newer versions as [this demo](https://codepen.io/miclgael/pen/JqWJma) seems to work fine - I can drag after a long tap.
6. All CSS classes used should have a `rz-` prefix.

V45370:
Thanks for the feedback, I will resolve the issues asap.

PR #1380, opened 2024-02-20
Allow the icon in the ProfileMenu to be enabled/disabled

rauschp:
Currently the ProfileMenu component always renders an arrow down icon.  This may not be preferred in certain instances, so I've added the ability to disable it. This will be enabled by default though! 

I haven't noticed any tests pertaining to this component, so I didn't update/extend any.

### DIFF
--- a/Radzen.Blazor/RadzenProfileMenu.razor.cs
+++ b/Radzen.Blazor/RadzenProfileMenu.razor.cs
@@ -42,6 +42,13 @@ namespace Radzen.Blazor
         [Parameter]
         public EventCallback<RadzenProfileMenuItem> Click { get; set; }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/RadzenProfileMenu.razor.cs
```diff
@@ -42,6 +42,13 @@ protected override string GetComponentCssClass()
         [Parameter]
         public EventCallback<RadzenProfileMenuItem> Click { get; set; }
 
+        /// <summary>
+        /// Enable/Disable the "arrow down" icon
```
enchev, line 46: Maybe the correct description should be Show/Hide instead

rauschp: Good call - updated!

PR #1387, opened 2024-02-22
Axis Color Band Ranges

paulo-rico:
As per forum message - https://forum.radzen.com/t/help-with-lineserieschart-needs-highlighted-area-above-and-below/16503

Add ability to create coloured bands on Value Axis.

### DIFF
--- a/Radzen.Blazor/Radzen.Blazor.csproj
+++ b/Radzen.Blazor/Radzen.Blazor.csproj
@@ -8,7 +8,7 @@
     <IsPackable>true</IsPackable>
     <PackageId>Radzen.Blazor</PackageId>
     <Product>Radzen.Blazor</Product>
-    <Version>4.25.4</Version>
+    <Version>4.25.5</Version>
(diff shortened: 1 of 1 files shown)


akorchev:
Hi @paulo-rico,

I think the suggested API is a bit restrictive - it could be more declarative and also support setting both category and value so bands can be horizontally stacked. It can't be set from a property grid as it currently stands - it relies on a collection of objects instead of tags.

I was thinking of something similar to the existing RadzenSeriesValueLine.

```
<RadzenSeriesBand ValueStart="0" ValueEnd="1000" CategoryStart="January" CategoryEnd="February" Fill="red" />
```

It could also be called RadzenChartBand as it applies to the whole chart which doesn't currently support multiple axes.
What do you think?

paulo-rico:
Hi @akorchev 

I see what you mean. I read the forum thread and it appeared simple on the face of it. I wasn't aware of the chart overlays. That method does fit this purpose better. It also looks like a learning curve while I see how it all hangs together, so may take a short while.

Until then

Paul

PR #1395, opened 2024-02-27
Scheduler drag and drop with mobile support

V45370:
All issues should be resolved now. Tested touch support on Android and on Chrome dev tools for mobile.
Had to use [DragDropTouch.js](https://github.com/Bernardo-Castilho/dragdroptouch)  with a little modification described [here](https://github.com/Postlagerkarte/blazor-dragdrop/issues/147#issuecomment-1200308069)
It is currently added as a separate DragDropTouch.js file, but you might want to integrate it inside Radzen.Blazor.js

Поздрави


akorchev:
> Had to use [DragDropTouch.js](https://github.com/Bernardo-Castilho/dragdroptouch) with a little modification described https://github.com/Postlagerkarte/blazor-dragdrop/issues/147#issuecomment-1200308069

I am afraid we can't accept a PR that introduces a third party js dependency. And just including the file in Radzen.Blazor.js isn't an option either - both license wise and size-wise. Radzen already includes drag and drop implementations and we don't want another one - RadzenSlider, RadzenColorPicker and RadzenDialog use drag and drop.

V45370:
In this case, I will investigate further

akorchev:
I think HTML5 drag and drop works out of the box on latest mobile devices. That's what this page claims https://caniuse.com/dragndrop 

Perhaps the implementation you suggested needs a little tweaking to make it work on a mobile device (without a polyfil or a brand new drag and drop implementation).

akorchev:
You can add a method or two if needed to Radzen.Blazor.js. Also check if some of the existing code would help - there is drag and drop in the splitter, datagrid, slider, dialog etc.

akorchev:
I think I've completed the implementation - removed touch events and the polyfill. It seems setting ondragstart as a string was the only thing required to enable drag and drop on mobile devices.

PR #1398, opened 2024-02-28
feat: column support custom filter clearing

marekm294:
ClearFIlters didn't clear custom Filter.

### DIFF
--- a/Radzen.Blazor/RadzenDataGrid.razor.cs
+++ b/Radzen.Blazor/RadzenDataGrid.razor.cs
@@ -935,7 +935,7 @@ namespace Radzen.Blazor
         /// <summary>
         /// Сlear filter on the specified column
         /// </summary>
-        public async Task ClearFilter(RadzenDataGridColumn<TItem> column, bool closePopup = false)
+        public async Task ClearFilter(RadzenDataGridColumn<TItem> column, bool closePopup = false, bool shouldReload = true)
         {
             if (closePopup)
             {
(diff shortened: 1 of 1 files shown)


Radzen.Blazor/RadzenDataGrid.razor.cs
```diff
@@ -1963,9 +1964,10 @@ public void Reset(bool resetColumnState = true, bool resetRowState = false)
 
             if (resetColumnState)
             {
-                allColumns.ToList().ForEach(c => 
+                allColumns.ToList().ForEach(async c =>
```
enchev, line 1967: async voids should not be used

Radzen.Blazor/RadzenDataGridColumn.razor.cs
```diff
@@ -323,6 +323,14 @@ public string ColumnPickerTitle
         [Parameter]
         public string CustomFilterExpression { get; set; }
 
+        /// <summary>
+        /// Gets or sets ClearCustomFilter.
+        /// ClearCustomFilter should reset all data related to filter.
+        /// ClearCustomFilter is invoked in <see cref="ClearCustomFilterAsync"/>.
+        /// </summary>
+        [Parameter]
+        public EventCallback ClearCustomFilter { get; set; }
```
enchev, line 332: There is already FilterCleared event

marekm294: Ok, I did sth different.
Problem is that it won't fix problem of method RadzenDataGridColumn.ClearFilters(). But it solves my problem so it's suitable solution for me.

I would suggest you to add bug to your backlog to fix CustomFilterExpression problem in the future.

PR #1409, opened 2024-03-04
DatePicker add specific css selectors for week number

nielsNocore:
added specific scss selector for week number in the date picker
added unittest for the week number in date picker
unit teste file auto formated. 

Added these selector so it is easier to add specific styling to the week number column (italic for instance)
image

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor
+++ b/Radzen.Blazor/RadzenDatePicker.razor
@@ -79,7 +79,7 @@
                                         <tr>
                                             @if (ShowCalendarWeek)
                                             {
-                                                <td class="rz-datepicker-week-number">@Culture.Calendar.GetWeekOfYear(new DateTime(CurrentDate.Year, CurrentDate.Month, 1).AddDays(i * 7), Culture.DateTimeFormat.CalendarWeekRule, Culture.DateTimeFormat.FirstDayOfWeek)</td>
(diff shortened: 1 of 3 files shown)


Radzen.Blazor/RadzenDatePicker.razor
```diff
@@ -79,7 +79,7 @@
                                         <tr>
                                             @if (ShowCalendarWeek)
                                             {
-                                                <td class="rz-datepicker-other-month">@Culture.Calendar.GetWeekOfYear(new DateTime(CurrentDate.Year, CurrentDate.Month, 1).AddDays(i * 7), Culture.DateTimeFormat.CalendarWeekRule, Culture.DateTimeFormat.FirstDayOfWeek)</td>
+                                                <td class="rz-datepicker-week-number">@Culture.Calendar.GetWeekOfYear(new DateTime(CurrentDate.Year, CurrentDate.Month, 1).AddDays(i * 7), Culture.DateTimeFormat.CalendarWeekRule, Culture.DateTimeFormat.FirstDayOfWeek)</td>
```
enchev, line 82: This will be breaking change for all users with custom themes. Maybe you can leave the old class and add the new one.

nielsNocore:
makes sense, restored the existing class name in the last commit. 
Thankx

PR #1411, opened 2024-03-05
Try to fix exception

DisplayName2023:
foreach could throw exception during rendering when dialogs is changed


enchev:
Any example how to reproduce the exception?

DisplayName2023:
> Any example how to reproduce the exception?

Open or close the dialog in a working thread

PR #1419, opened 2024-03-11
Added rz-state-disabled class to list box item

k1rillka:
ListBox allows to disable items but no classes reflect that item is disabled.
Added disabled class to the list box item.


enchev:
There is no Disabled property for the ListBox items, that's why this class is not applied. Such property however is available for the entire ListBox:
image
image

PR #1426, opened 2024-03-17
Escape or Tab enable parameter so as not to close the dialog

Pstariell:
if I have added an add button and I want to tab from the search input to the search button and finally add button from the keyboard it is not possible. (the dialog is closed)
With the parameter you can eliminate this malfunction
:)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@v3
       with:
-        dotnet-version: 7.0.x
+        dotnet-version: 8.0.x
     - name: Build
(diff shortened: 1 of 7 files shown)


Pstariell (commented):
Fix Comment

enchev:
Hi @Pstariell,

Not sure I understand some of changes you’ve made. Our projects are already .NET 8.

Pstariell:
> Not sure I understand some of changes you’ve made. Our projects are already .NET 8.

Hello, nice to meet you, I hope my change is appreciated, it should simply resolve in case you want to use the "keyboard tab" in the RadzenDropDownDataGrid component not to close the dialog (useful for when you set the Add parameter)
image

PR #1427, opened 2024-03-18
Refactored the inline scripts in the Date Picker

nosvarato:
* Change the inline scripts inside the Date Picker component to take away the need for Script-Sources unsafe-inline in the content security policy which was a penetration test finding.
* Also removed duplicate TValues which one was set to int and one was set to double which caused errors, Remove the TValue=double as the binding value is int


enchev:
Thanks for the pull request, @nosvarato, however I’m afraid that we will not be able to merge it. At the moment the picker is opening using plain JavaScript and switching completely to Blazor will introduce significant lag in case of server Blazor which is highly undesirable. Using also of async voids will cause UI/state synchronisation problems. I’ll keep the request open and will close it with commit that implements the same JavaScript only opening however not part of the rendering.

Issue #1445, opened 2024-03-28
Scheduler week view - Day not aligning to column in mobile


thekim1:
Description:
In the scheduler week view and to some extent day view the header showing the weekday is not aligned to the columns showing the content of the day when in mobile. 
It looks like the "rz-slot-hour-header" is taking up to much space.

Steps to reproduce:
Can be seen in the documentation at https://blazor.radzen.com/scheduler if using mobile device or mobile sized view in web-browser

1. Go to https://blazor.radzen.com/scheduler
2. Click on week view of a scheduler component
3. See the header not aligning to the columns
![image](https://github.com/radzenhq/radzen-blazor/assets/36292640/25afd477-5501-4c98-8cda-1016011db8f8)

Alternatively link your repo with a sample project that can be run.

Expected behavior:
Days in header should align to the columns even in mobile view

Screenshots
Problem in mobile view:
![image](https://github.com/radzenhq/radzen-blazor/assets/36292640/4b234dba-d669-4c1f-b584-f231ae1acdcb)

Tablet view working as expected:
![image](https://github.com/radzenhq/radzen-blazor/assets/36292640/dc466f2a-ab01-4d43-9ea3-73b94a696197)

Desktop
 - OS: windows 11
 - Browser: Chrome
 - Version 123.0.6312.86
 
 Mobile:
- OS: IOS 17.3.1
- Browser: Safari

thekim1:
Made a fork and commited a possible solution. 
Can i make a pullrequest for this or what would be a better way to fix this?

PR #1450, opened 2024-04-03
Added 2.5rem width to class rz-slot-hour-header with max-width: 576px…

thekim1:
Css fix for mobile resolution in Scheduler module to fix with issue #1445


yordanov:
Thanks for the suggested fix, @thekim1! 🙏🏻  The issue however is still reproducible on screen sizes larger that 576px and would require a deeper CSS change. Also setting a fixed width might not work well in different themes / design systems. Closing this PR now and working on a universal fix.

PR #1472, opened 2024-04-19
set ShowTimeForDateTimeFilter and ShowUpDownForNumericFilter using properties in DataGridColumn

nielsNocore:
added parameters in datagridColumn to set filter properties,
not needed to create a custom override any more. **(breaking change)**


Radzen.Blazor/RadzenDataGridColumn.razor.cs
```diff
@@ -1292,22 +1304,6 @@ internal string GetFilterOperatorSymbol(FilterOperator filterOperator)
             }
         }
 
-        /// <summary>
```
enchev, line 1295: Maybe it will be better to keep both methods and return the property value to avoid the breaking change

PR #1502, opened 2024-05-12
Implemented RadzenAsyncLoader component for Blazor, providing asynchronous data loading

itsalfredakku:
With optional loading and data templates. Resembles the functionality of Flutter's FutureBuilder component.

### DIFF
--- a/Radzen.Blazor/RadzenAsyncLoader.razor
+++ b/Radzen.Blazor/RadzenAsyncLoader.razor
@@ -2,9 +2,9 @@
 
 @if (IsLoading && LoadingTemplate != null)
 {
-    @LoadingTemplate()
+    @LoadingTemplate
 }
-@if (Data != null)
+else if (Result != null)
 {
-    @Template(Data)
+    @Template(Result)
 }
--- a/Radzen.Blazor/RadzenAsyncLoader.razor
+++ b/Radzen.Blazor/RadzenAsyncLoader.razor
@@ -1,0 +1,10 @@
+﻿@typeparam TData
+
+@if (IsLoading && LoadingTemplate != null)
+{
+    @LoadingTemplate()
+}
+@if (Data != null)
+{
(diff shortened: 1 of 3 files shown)


enchev:
Thanks for the pull request @itsalfredakku! Please add an example in the demos in order to know how this component can be used.

itsalfredakku:
# AsyncLoader component
This article demonstrates how to use the AsyncLoader component.

# Loading data asynchronously
The AsyncLoader component enables loading data asynchronously in Blazor applications. It offers the flexibility to specify loading and loaded templates, handle errors during data retrieval, and update the UI accordingly.

```
<RadzenAsyncLoader TData="Server.Models.ApplicationUser" DataTask="@Security.GetUserById(context.Id)">
    <LoadingTemplate>
        <RadzenProgressBarCircular ProgressBarStyle="ProgressBarStyle.Primary" Value="100" ShowValue="false" Mode="ProgressBarMode.Indeterminate" Size="ProgressBarCircularSize.Small" />
    </LoadingTemplate>
    <Template Context="ctx">
        <RadzenStack Orientation="Orientation.Horizontal" AlignItems="AlignItems.Center" Gap="0.5rem">
            @foreach (var role in ctx.Roles)
            {
                <RadzenBadge Text="@role.Name" />
            }
        </RadzenStack>
    </Template>
</RadzenAsyncLoader>
```

akorchev:
What is the purpose of `TData` here? It is usually used for data-bound components - ones that have "items" and the template is repeated for each item. Is it only to get the result out of the DataTask? What if I just want to use `Task` (the non-generic version)? Or ValueTask? 

Another issue I see is that the loading content can be shown only once with the current implementation. This is one of the first things people are going to ask for - show the loading content multiple times. I just feel the API should be different - maybe a single `boolean IsLoading` parameter which the user can set accordingly.

PR #1503, opened 2024-05-13
Add a Footer template to RadzenDropZone

MPapst:
This PR adds a Footer Template to the RadzenDropZone.

As the ChildContent RenderFragment is rendered above the Items, I was in need of a way to add items from within the DropZone.

Demo is added:
image


akorchev:
I think the property should be called `Footer` as Template suggest some context is being passed or the item is instantiated for every data item.

PR #1509, opened 2024-05-17
Allow disabling splitter resize client to server invocations

### LINKED ISSUE #1506: RadzenSplitter with lots of content freezes
**Describe the bug**
Since 4.27.0 RadzenSplitter freezes when there's certain complex content on the panes.

**To Reproduce**
1. install 4.27.0 or later
2. create a page:
```razor
@page "/"

<div style="width: 800px; height: 600px">
    <RadzenSplitter>
        <RadzenSplitterPane Style="background: red">
            <RadzenDataGrid TItem=@Item Data=[] Count=0 AllowFiltering>
                <Columns>
                    @for (var i = 0; i < 100; i++)
                    {
                        <RadzenDataGridColumn TItem=@Item Property=@nameof(Item.Property) Title="Property" />
                    }
                </Columns>
            </RadzenDataGrid>
        </RadzenSplitterPane>
        <RadzenSplitterPane Style="background: green">
            <RadzenDataGrid TItem=@Item Data=[] Count=0 AllowFiltering>
                <Columns>
                    @for (var i = 0; i < 100; i++)
                    {
                        <RadzenDataGridColumn TItem=@Item Property=@nameof(Item.Property) Title="Property" />
                    }
                </Columns>
            </RadzenDataGrid>
        </RadzenSplitterPane>
    </RadzenSplitter>
</div>

@code {
    class Item
    {
        public string Property { get; set; }
    }
}
```
3. Move the splitter couple of times -> it will freeze and memory keeps getting allocated more and more

**Expected behavior**
Do the same using 4.26.1 -> everything works smoothly.

**Desktop (please complete the following information):**
 - Browser: Firefox, Edge, Chrome

iotalambda:
Fixes #1506. By setting `ClientToServerInvocationsOnPaneResize = false`, both the freezing problem and infinite memory consumption are gone and splitters resize smoothly, even if the render mode is interactive, there's network lag and there's complex content on the panes.

### DIFF
--- a/Radzen.Blazor/RadzenSplitter.razor.cs
+++ b/Radzen.Blazor/RadzenSplitter.razor.cs
@@ -30,6 +30,13 @@ namespace Radzen.Blazor
         [Parameter]
         public Orientation Orientation { get; set; } = Orientation.Horizontal;
 
+        /// <summary>
+        /// Gets or sets whether server side <see cref="JSInvokableAttribute">JSInvokable</see>s should be invoked by the client on pane resizing events or not. Set to <c>true</c> by default.
+        /// </summary>
+        /// <value><c>true</c> if should be invoked; otherwise, <c>false</c>.</value>
+        [Parameter]
+        public bool ClientToServerInvocationsOnPaneResize { get; set; } = true;
+
         internal List<RadzenSplitterPane> Panes = new List<RadzenSplitterPane>();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


enchev:
Hey @iotalambda,

I'm unable to reproduce such problem:
splitter-resize

The change added to 4.27.0 was the StateHasChanged() invoke on resize which will recreate the panes content and this indeed might have performance hit depending on what you have. I'm not sure however that disabling this invoke is the proper solution and honestly I don't like the name `ClientToServerInvocationsOnPaneResize`. Maybe ChangePaneStateOnResize will be a better name.

iotalambda:
@enchev Hi, instead of this PR, I decided to implement a fully client side version of RadzenSplitter (with styling from the original splitter), so that all clicks and drags are handled in the client-side only, and it seems to solve the problem even better. I'm quite sure the lagginess occurs only if there's complex data on the panes and also some network latency. In case you want to introduce the fully client side alternative to the library, I can probably create a PR for it at some point.

PR #1513, opened 2024-05-17
RadzenTreeItem now gets children from Value rather than Items list

grwomack:
Updated the RadzenTree to provide the level's "ChildrenProperty" value to the RadzenTreeItems.  If that value exists in the tree item, then the tree item will get it's children from the Value property rather than the Items property.

### DIFF
--- a/Radzen.Blazor/RadzenTree.razor.cs
+++ b/Radzen.Blazor/RadzenTree.razor.cs
@@ -233,7 +233,7 @@ namespace Radzen.Blazor
         public EventCallback<IEnumerable<object>> CheckedValuesChanged { get; set; }
 
         void RenderTreeItem(RenderTreeBuilder builder, object data, RenderFragment<RadzenTreeItem> template, Func<object, string> text,
-            Func<object, bool> hasChildren, Func<object, bool> expanded, Func<object, bool> selected, IEnumerable children = null)
+            Func<object, bool> hasChildren, Func<object, bool> expanded, Func<object, bool> selected, IEnumerable children = null, string childrenProperty = null)
         {
             builder.OpenComponent<RadzenTreeItem>(0);
             builder.AddAttribute(1, nameof(RadzenTreeItem.Text), text(data));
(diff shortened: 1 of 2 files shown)


grwomack:
Pull request allows the `RadzenTreeItem` to be aware of the `ChildrenProperty` that is passed through the `RadzenTreeLevel`.  This allows all of the children, regardless of whether or not they are rendered, to be added to the `CheckedValues` list in the `RazdenTree` when selecting parent checkboxes.

To accomplish this, I had to a 1 new parameter to `RazenTreeItem`:
`[Parameter] public string ChildrenProperty { get; set; }`

Next I updated the `RenderTreeItem` method in the `RadzenTree` to pass the level's ChildrenProperty to the RadzenTreeItem that is building.  Only additions are shown below, anything not changed will be represented with ...

```
void RenderTreeItem(..., string childrenProperty = null)
{
    ...
    builder.AddAttribute(7, nameof(RadzenTreeItem.ChildrenProperty), childrenProperty);
    ...
}
```

Now that the `RadzenTreeItem` knows what the `ChildrenProperty` name is, it can get all of it children from the Value property rather than the Items list.

```
internal IEnumerable<object> GetAllChildValues(Func<object, bool> predicate = null) 
{
    IEnumerable<object> children = [];
    if (string.IsNullOrEmpty(ChildrenProperty))
    {
        children = items.Concat(items.SelectManyRecursive(i => i.items)).Select(i => i.Value);
    }
    else
    {
        children = GetChildrenRecusvie(Value);
    }

    return predicate != null ? children.Where(predicate) : children;
}

public IEnumerable<object> GetChildrenRecusvie(object value)
{
    IEnumerable<object> children = PropertyAccess.GetValue(value, ChildrenProperty) as IEnumerable<object>;
    if (children is not null && children.Any())
    {
        foreach (object child in children)
        {
            IEnumerable<object> grandChildren = GetChildrenRecusvie(child);
            if (grandChildren is not null && grandChildren.Any())
            {
                children = children.Concat(grandChildren);
            }
        }
    }

    return children;
}
```



https://github.com/radzenhq/radzen-blazor/assets/16569753/222cae36-afa5-4791-9aaf-e8f881ce351e

akorchev:
I see the following problems with this implementation:
1. It could lead to perfomance issues - e.g. loading a lot of database items. For example `ChildrenProperty` could be a JOIN to a related table. Loading all children when the user checks an itemd would select all related items which could lead to a SELECT+1 problem (unexpected multiple DB queries). 
2. This approach would load all children every time a parent node is checked.
3. It introduces a new parameter ChildrenProperty which doesn't do anything else. It should probably be provided by other means (cascading value?)
4. Doesn't seem to support load on demand mode via NodeExpand.
5. Doesn't seem to support inline definition.

Those reasons are why this isn't provided out of the box - there are a lot of different scenarios that don't allow just populating everything at once for the purpose of checking a node.

PR #1516, opened 2024-05-18
Avoid partially hiding of chart tooltip near chart ends

### LINKED ISSUE #560: LineChart Tooltip near to the browser end doesn't work
If I have a LineChart and a tooltips is shown in the near of the left end of the display/browser, it will only show a part of the tooltip, the rest is cut by the end of the screen

```
<RadzenChart @ref="chart0">
        @foreach(var item in delevererEKDatas!)
        {
            string tile = item.LineName;
                string color = getRandomColor(item.MainDeliverer);
                <RadzenLineSeries Stroke="@color" Smooth="@smooth" Data="@item.EKData" CategoryProperty="Date" Title=@tile LineType="LineType.Solid" ValueProperty="Price">
                    <ChildContent>
                        <RadzenMarkers MarkerType="MarkerType.Auto" Stroke="@color" />
                    </ChildContent>
                    <TooltipTemplate Context="data">
                    <div>
                        @data.DelInfo<br/>
                        <strong>@data.Date.ToString("dddd, dd MMMM yyyy")</strong><br/>
                        <strong>EK-Preis: @data.Price.ToString("N2")</strong><br/>
                        ab Menge @data.MinQuantity.ToString("N0")<br/>
                            @if (@data.AvailableQuantity is not null) 
                            { 
                                <span>verfügbare Menge: @data.AvailableQuantity</span>
                            }
                    </div>
                    </TooltipTemplate>
                </RadzenLineSeries>
            }
    
        <RadzenCategoryAxis  Padding="20" Min="@searchArticleDateDeliverer.StartDate" Max="@searchArticleDateDeliverer.EndDate" Step="31" FormatString="{0:dd.MM.yy}" />
        <RadzenValueAxis Formatter="@FormatAsEURO">
        </RadzenValueAxis>
        @if(delevererEKDatas.Count == 1)
        {
                <RadzenLegend Position="LegendPosition.Bottom" Visible="false" />
        }
        else if(delevererEKDatas.Count > 4)
        {
            <RadzenLegend Position="LegendPosition.Right" Visible="true" />
        }
        else
        {
            <RadzenLegend Position="LegendPosition.Bottom" Visible="true" />
        }
    </RadzenChart> 
```

maxbruecken:
Added new chart tooltip CSS classes to show tooltip without cutting near ends of the chart. This fixes #560.

### DIFF
--- a/Radzen.Blazor/Rendering/ChartTooltip.razor
+++ b/Radzen.Blazor/Rendering/ChartTooltip.razor
@@ -1,4 +1,8 @@
-<div class="rz-chart-tooltip @Class" style="left: @(X.ToInvariantString())px; top: @(Y.ToInvariantString())px">
+@using Microsoft.JSInterop
+
+@inject IJSRuntime JSRuntime
+
+<div @ref="@chartTooltipElement" class="rz-chart-tooltip rz-chart-tooltip-top-center @Class" style="left: @(X.ToInvariantString())px; top: @(Y.ToInvariantString())px">
     <div class="rz-chart-tooltip-content" style="@Style">
     @if (ChildContent != null)
     {
@@ -38,4 +42,20 @@
 
     [Parameter]
     public string Class { get; set; }
+
+    [CascadingParameter]
+    public RadzenChart Chart { get; set; }
+
+    ElementReference? chartTooltipElement;
+
+    /// <inheritdoc />
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        await base.OnAfterRenderAsync(firstRender);
+        
+        if (chartTooltipElement != null)
+        {
+            await JSRuntime.InvokeAsync<string>("Radzen.setChartTooltipCssClass", chartTooltipElement, Chart.Element, X, Y);
+        }
+    }
 }
--- a/Radzen.Blazor/themes/components/blazor/_chart.scss
+++ b/Radzen.Blazor/themes/components/blazor/_chart.scss
@@ -238,10 +238,33 @@ $chart-color-schemes: (
 
 .rz-chart-tooltip {
   position: absolute;
-  transform: translate(-50%, -100%);
   transition: top 0.2s, left 0.2s;
   top: 0;
   left: 0;
+}
+
+.rz-chart-tooltip-top-right {
+  transform: translate(-11px, -100%);
+}
+
+.rz-chart-tooltip-top-center {
+  transform: translate(-50%, -100%);
+}
+
+.rz-chart-tooltip-top-left {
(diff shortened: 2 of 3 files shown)


akorchev:
What problem does this PR solve? The linked issue doesn't help a lot.

maxbruecken:
This PR solves the problem with hiding of part of chart tooltip if the data points are near to the for example top end of chart or window. The fist screenshot shows the problem, the second one shows how the tooltip will be rendered with changes from this PR.

image

image

akorchev:
Thank you for the clarification. While testing I found those issues:


1. Doesn't seem to work in Safari
image. 
1. The tooltip changes direction unexpectedly in some cases. Check this gif:
change-of-direction. There is enough room for the tooltip but it still flips which I think is unexpected.

Probably the issue is the boundary check function - it checks if the tooltip is within the chart wrapper element. I think a better implementation is to check for the screen boundaries. Already done by some components (check the `openPopup` function).

maxbruecken:
Thank you for testing. I will check and try to fix issues you found. Unfortunately only to check screen boundaries was not enough, in some situations a part of the chart is covered by other elements with higher z-order (header, footer etc.) and thus the chart tooltip can be covered by this elements too.

akorchev:
> in some situations a part of the chart is covered by other elements with higher z-order (header, footer etc.) and thus the chart tooltip can be covered by this elements too.

This can't be always solved by flipping the direction alone. A 100% working solution would be to render the chart tooltip as RadzenTooltip which attaches at the `<body>` and overlays everything and has screen boundary detection as a built-in.

maxbruecken:
> > in some situations a part of the chart is covered by other elements with higher z-order (header, footer etc.) and thus the chart tooltip can be covered by this elements too.
> 
> This can't be always solved by flipping the direction alone. A 100% working solution would be to render the chart tooltip as RadzenTooltip which attaches at the `<body>` and overlays everything and has screen boundary detection as a built-in.

Good point, thank you. I would close this PR and try to reimplement the chart tooltip as RadzenTooltip and create a new PR later. 

Should it be configurable whether to use RadzenTooltip instead of chart tooltip or should I simple replace current chart tooltip?

akorchev:
Hi @maxbruecken,

Replacing the chart tooltip with RadzenTooltip would be a big task so be prepared :) 
- The chart tooltip uses styling from the current series (to set the border to be of the same color).
- RadzenTooltip relies on an ElementReference to open. The chart tooltip uses X and Y coordinates instead.

> Should it be configurable whether to use RadzenTooltip instead of chart tooltip or should I simple replace current chart tooltip?

No as it would be very hard to explain why such an option even exists. However the current look and feel and behavior should be preserved as much as possible.

PR #1518, opened 2024-05-18
Dialog - custom container class

Techoneshot:
New parameter ContentCssClass

### DIFF
--- a/Radzen.Blazor/Rendering/DialogContainer.razor
+++ b/Radzen.Blazor/Rendering/DialogContainer.razor
@@ -175,6 +175,15 @@
         }
     }
 
+    string ContentCssClass
+    {
+        get
+        {
+            var baseCss = "rz-dialog-content";
+            return string.IsNullOrEmpty(Dialog.Options.ContentCssClass) ? baseCss : $"{baseCss} {Dialog.Options.ContentCssClass}";
+        }
+    }
+
     string Style
     {
         get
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/Rendering/DialogContainer.razor
```diff
@@ -31,7 +31,7 @@
             }
             else
             {
-                <div class="rz-dialog-titlebar">
```
enchev, line 34: What happened with `rz-dialog-titlebar`?

Techoneshot: Sorry for inattention, now this should be right

Radzen.Blazor/Rendering/DialogContainer.razor
```diff
@@ -175,6 +175,15 @@
         }
     }
 
+    string ContentCssClass
+    {
+        get
+        {
+            var baseCss = "rz-dialog-content";
+            return string.IsNullOrEmpty(Dialog.Options.ContentCssClass) ? baseCss : $"{baseCss} {Dialog.Options.ContentCssClass}";
```
enchev, line 183: In my opinion in the else part of this condition you will get twice "rz-dialog-content" since it's already assigned in the DialogService:
```
ContentCssClass = options != null ? $"rz-dialog-content {options.ContentCssClass}" : "rz-dialog-content",
```

Techoneshot: It's same as others CSS params but I will check that

PR #1539, opened 2024-06-01
Added ChipTemplate to DropDownDataGrid

stlufred:
Chips only had effect if no Template was present (line 41).  And the markup for when Chips was active was trying to use template (line 50).

Fixed the confusion by adding a new template to personalize the rendering of the chips...


enchev:
Thanks for the pull request @stlufred! I think that we should fix this to work like the DropDown component where ValueTemplate and Template are used if defined for chips: https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenDropDown.razor#L44-L55

I'll push the relevant code and I'll close this.

PR #1541, opened 2024-06-03
Add FilterOperator null check so SetParametersAsync doesn't override initial values

nitrouscookies:
FilterOperator initial values set in markup are being overridden since this [commit] from 4.30.0 (https://github.com/radzenhq/radzen-blazor/commit/6fe6b681acbdcd5409735b90aadcdce62187516f).

See this thread for more details
https://forum.radzen.com/t/filteroperator-startswith-not-being-set-on-initial-load/17397/3


nitrouscookies:
Repo was out of date, will re-submit

PR #1549, opened 2024-06-08
Index calculation when adding panels dynamically

AscarGb:
This fixes an issue when dynamically inserting panels into the middle of a collection.

```
    <RadzenSplitter>

        @if (FirstItems is { Count: > 0 })
        {
            @foreach (var c in FirstItems)
            {
                <RadzenSplitterPane>
                    <SomeComponent1>
                    </SomeComponent1>
                </RadzenSplitterPane>
            }
        }

        @if (LastItem is not null)
        {
            <RadzenSplitterPane>
                <SomeComponent2>
                </SomeComponent2>
            </RadzenSplitterPane>
        }

    </RadzenSplitter>
```

If you first add LastItem, and then FirstItems, the index of the latter will not be calculated correctly.

But it will work like this:
```
    <RadzenSplitter UseGetIndex="true">

        @if (FirstItems is { Count: > 0 })
        {
            @foreach (var c in FirstItems)
            {
                <RadzenSplitterPane GetIndex="() => Array.IndexOf(FirstItems.ToArray(), c)">
                    <SomeComponent1>
                    </SomeComponent1>
                </RadzenSplitterPane>
            }
        }

        @if (LastItem is not null)
        {
            <RadzenSplitterPane GetIndex="() => FirstItems.Count">
                <SomeComponent2>
                </SomeComponent2>
            </RadzenSplitterPane>
        }

    </RadzenSplitter>
```

### DIFF
--- a/Radzen.Blazor/RadzenSplitter.razor.cs
+++ b/Radzen.Blazor/RadzenSplitter.razor.cs
@@ -14,8 +14,6 @@ namespace Radzen.Blazor
     /// </summary>
     public partial class RadzenSplitter : RadzenComponent
     {
-        private int _sizeautopanes = 0;
-
         /// <summary>
         /// Gets or sets the child content.
         /// </summary>
@@ -30,7 +28,13 @@ namespace Radzen.Blazor
         [Parameter]
         public Orientation Orientation { get; set; } = Orientation.Horizontal;
 
-        internal List<RadzenSplitterPane> Panes = new List<RadzenSplitterPane>();
+        /// <summary>
+        /// Use GetIndex in panes
+        /// </summary>
+        [Parameter]
+        public bool UseGetIndex { get; set; }
+        
+        internal List<RadzenSplitterPane> Panes = new();
 
         /// <summary>
         /// Adds the pane.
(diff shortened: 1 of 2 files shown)


enchev:
Try to define `@key` - it's necessary when adding Blazor components in a loop.

AscarGb:
> Try to define `@key` - it's necessary when adding Blazor components in a loop.

It won't help. When a new RadzenSplitterPane appears, it is added to the collection with the last index.

RadzenSplitter.razor:
```
    <div @ref="@Element" @attributes="Attributes" class="@GetCssClass()" id="@GetId()" style="@Style">
        <CascadingValue Value=this>
            @ChildContent
        </CascadingValue>
    </div>
```

RadzenSplitterPane.razor:
```
        [CascadingParameter]
        public RadzenSplitter Splitter
        {
            get => _splitter;
            set
            {
                if (_splitter != value)
                {
                    _splitter = value;
                    _splitter.AddPane(this);
                }
            }
        }
```
The AddPane method is called only for new RadzenSplitterPane. And instead of dividing the panels, the separator will appear at the end:

image

enchev:
I'm not sure I understand you case. Here is what I've tried and it worked:
image
image

AscarGb:
> I'm not sure I understand your case. Here is what I've tried and it worked: image image

Your example will work because all panels are added in order once. Instead of the first "true", put a variable, initially at "false". And then change it to "true" during a user action. The panel that appears will be in the correct place, but with the wrong index.

enchev:
I see what you are after - it's a common problem with rendering order of Blazor child components when dynamically plug a new component - for example: https://github.com/dotnet/aspnetcore/issues/54703

Here is one approach that will avoid this - the idea is to force refresh of all child components:
```
<RadzenSwitch @bind-Value=visible Change="@(args => visibleChanged = true)" />

<RadzenSplitter @ref=splitter>
    @RenderChildContent()
</RadzenSplitter>

@code {
    RadzenSplitter splitter;
    bool visible;
    bool visibleChanged;

    RenderFragment RenderChildContent()
    {
        return __builder =>
        {
            if (visibleChanged)
            {
                visibleChanged = false;
                <text>
                </text>
            }
            else
            {
                <text>
                    @if (visible)
                    {
                        <RadzenSplitterPane>First Pane</RadzenSplitterPane>
                    }

                    @foreach (var i in Enumerable.Range(1, 1))
                    {
                        <RadzenSplitterPane>Pane@(i)</RadzenSplitterPane>
                    }

                    @if (true)
                    {
                        <RadzenSplitterPane>Last Pane</RadzenSplitterPane>
                    }
                </text>
            }
        };
    }
}
```
splitter

AscarGb:
Re-drawing what is not needed is a very bad solution for a large page. Blazor specifically monitors this to avoid doing unnecessary work that reduces productivity.
The page can have nested forms and support moving panels. Your solution seems more like a crutch.
image

PR #1571, opened 2024-06-25
Changes aria-live regions for better screen reader

efinder2:
According to the documentation of the [mdn ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) live regions are only prompted if they are present **before** the content changes / is added. 

The live region of the RadzenAlert was moved outside of the `if(visible)`. If notifications are added, they will be announced.

For the same reason the aria-live was moved from the NotificationMessage to the Notification. 

Also the `role=log` was added to the example console so changes were announced there.

Those changes were tested with NVDA.

### DIFF
--- a/Radzen.Blazor/RadzenAlert.razor
+++ b/Radzen.Blazor/RadzenAlert.razor
@@ -2,7 +2,7 @@
 <div aria-live="polite">
 @if (visible)
 {
-    <div @ref="@Element" style="@Style" @attributes="Attributes" class="@GetCssClass()" id="@GetId()>
+    <div @ref="@Element" style="@Style" @attributes="Attributes" class="@GetCssClass()" id="@GetId()">
         <div class="rz-alert-item">
             @if (ShowIcon)
             {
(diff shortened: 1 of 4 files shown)


Radzen.Blazor/RadzenAlert.razor
```diff
@@ -1,7 +1,8 @@
 ﻿@inherits RadzenComponentWithChildren
+<div aria-live="polite">
```
enchev, line 2: Changing the component rendering is a breaking change.

efinder2: How could we solve this? Problem is the live region must be present for the screen reader to read the modification. Is there possiblity to add it to the next major update?

akorchev: You can easily add that div whenever you use RadzenAlert.

efinder2: Yeah, got it. I'll remove it. Would the Changes of the RadzenNotification accepted alone?

akorchev: I think the changes to RadzenNotification are not a breaking change. You can update the pull request.

efinder2: done

PR #1587, opened 2024-07-09
MonthSelect event on all Year Views

paulo-rico:
Instigated from [this forum issue](https://forum.radzen.com/t/scheduler-change-month-from-year-view/16651)

Although the request was to open a month detail view from the Year View, this would not be generic enough for the scheduler component. But the request to react to a month click was valid. So this PR adds a `MonthSelect` event that reacts to -

- Year - The Month title of each month segment
- Timeline - The Month titles down the left hand side of the view
- Planner - The Month titles down either side of the view

Whilst creating this PR, I needed to return all appointments for the month selected. It became apparent that the method `AppointmentsInSlot` would return the appointments for the month with the required Start and End dates. With this in mind, it appears that `AppointmentsInRange` would be a more appropriate name.

What do you think? If that name change seems acceptable, let me know, and I'll make the change to the other views for consistency.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/IScheduler.cs
+++ b/Radzen.Blazor/IScheduler.cs
@@ -66,10 +66,15 @@ namespace Radzen.Blazor
         /// <summary>
         /// Selects the specified slot.
         /// </summary>
-        /// <param name="start">The start.</param>
+        /// <param name="monthStart">The start of the month.</param>
         /// <param name="end">The end.</param>
         /// <param name="appointments">The appointments for this range.</param>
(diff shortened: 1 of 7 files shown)


paulo-rico:
Hi @akorchev 

Thanks for the library! Development would be a damn site harder without it. Thanks also to Michael Washington for pointing me in this direction a few years ago. 👍

I think I'll leave the renaming. Why take the risk?! They are private and it works. Not worth tinkering with it.

Cheers

Paul

PR #1590, opened 2024-07-10
Stopped Radzen Dialog Service unwanted Render Cycles during dragging.

JoeCawthorne:
- I stopped Radzen Dialog Service unwanted Render Cycles during dragging.
- I did this by removing the current method for making dialogs draggable, and replacing it with a JavaScript implementation.
- This stop the unwanted render cycles during dragging because JavaScript can modify the element properities without triggering a render cycle.
- Eliminating these unwanter render cycles means prevents any render logic from being executed, 
- The unwanted render cycles cause render logic to be executed, which can make a dialogs dragging be slow. Eliminating these render cycles fixes this issue.

### DIFF
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -2290,5 +2290,53 @@ window.Radzen = {
         } else {
             start();
         }
+    },
+
+    makeElementDraggable: function (draggable, showTitle, dialogRef) {
+        if (draggable == "True" && showTitle == "True") {
+            var dialogs = document.querySelectorAll('.rz-dialog');
+            if (dialogs.length == 0) return;
+            var lastDialog = dialogs[dialogs.length - 1];
+
+            var initialPositionX = 0
+            var changeInX = 0
+            var initialPositionY = 0
+            var changeInY = 0
+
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/wwwroot/Radzen.Blazor.js
```diff
@@ -2290,5 +2290,53 @@ window.Radzen = {
         } else {
             start();
         }
+    },
+
+    makeElementDraggable: function (draggable, showTitle, dialogRef) {
+        if (draggable == "True" && showTitle == "True") {
```
enchev, line 2296: It will be better to pass these as booleans, not strings

Radzen.Blazor/wwwroot/Radzen.Blazor.js
```diff
@@ -2290,5 +2290,53 @@ window.Radzen = {
...
+            var lastDialog = dialogs[dialogs.length - 1];
+
+            var initialPositionX = 0
+            var changeInX = 0
+            var initialPositionY = 0
+            var changeInY = 0
+
+            var dialogTitles = document.querySelectorAll('.rz-dialog-titlebar');
+            if (dialogTitles.length == 0) return;
+            dialogTitles[dialogTitles.length - 1].addEventListener('mousedown', initiateElementDrag);
```
enchev, line 2308: If there is no drag this event handler will not be disposed.

Issue #1645, opened 2024-08-12
AutoComplete OpenOnFocus does not actually open


bert-algoet:
When using the AutoComplete component with the parameter OpenOnFocus it does not actually open the component on entering.

The issue seems to be that it does seem to open however it is empty, the content is not filled with the data because there is no filter yet.

enchev:
The component will open the popup as expected however since nothing is entered there will be no items.

bertalgoet:
That is indeed my issue, i would expect to see al possible values when focussing on the control with the OpenOnFocus=true, in the current version however nothing is being shown, so there seems no use for the OpenOnFocus parameter since just a bar of 1px is being shown.

bert-algoet:
> The component will open the popup as expected however since nothing is entered there will be no items.

That is indeed my issue, i would expect to see al possible values when focussing on the control with the OpenOnFocus=true, in the current version however nothing is being shown, so there seems no use for the OpenOnFocus parameter since just a bar of 1px is being shown.

PR #1646, opened 2024-08-12
AutoComplete OpenOnFocus did not load data correctly

bert-algoet:
Fix for issue #1645 where the OpenOnFocus opens an empty dropdown instead of showing all data


enchev:
Thanks for the pull request @bert-algoet however we believe that it will be better to use DropDown component for such cases if you want to list initially all items.

bert-algoet:
@enchev the reason i'm not using the dropdown is because you cannot enter new values that do not exist yet. I'll see what i can do without this change.

enchev:
Check this demo:
https://blazor.radzen.com/dropdown?theme=material3#editable-dropdown

PR #1647, opened 2024-08-13
Radzensteps onafterchange

iustin94:
The purpose of this PR is to enable the RadzenSteps component to provide more control over the flow of moving between steps. As well as making more clarity in the logic flow.

Things done:

- Renamed the CanChange event to BeforeChange: This is becase the CanChange event did not actually prevent the user clicking the change buttons, instead it allowed the button to be clicked and then the event handler had to do something to prevent that from happening.

- Added AllowPrevious and AllowNext boolean flags: Using these flags, the steps provide more control of when a user can go forward or backwards. This way, a user can be completelly prevented from triggering the previous/next button until they perform a required action in the wizard.

### DIFF
--- a/Radzen.Blazor/RadzenSteps.razor
+++ b/Radzen.Blazor/RadzenSteps.razor
@@ -60,11 +60,11 @@
     {
     <div class="rz-steps-buttons">
         <a id="@(GetId() + "prev")" title="@PreviousTitle" arial-label="@PreviousAriaLabel" tabindex="@(IsFirstVisibleStep() ? -1 : 0)"
-           class='@($"rz-steps-prev {(IsPreviousDisabled() ?  "rz-state-disabled" : "")}")' style="@(PreviousButtonStyle)"
+           class='@($"rz-steps-prev {(IsPreviousDisabled() ?  "rz-state-disabled" : "")}")'
                    @onkeypress="@(args => OnKeyPress(args, PrevStep()))" @onkeypress:preventDefault=preventKeyPress @onkeypress:stopPropagation
            @onclick="@PrevStep" @onclick:preventDefault="true"><span class="rzi"></span>@PreviousText</a>
         <a id="@(GetId() + "next")" title="@NextTitle" arial-label="@NextAriaLabel" tabindex="@(IsLastVisibleStep() ? -1 : 0)"
-           class='@($"rz-steps-next {(IsNextDisabled() ?  "rz-state-disabled" : "")}")' style="@(NextButtonStyle)"
+           class='@($"rz-steps-next {(IsNextDisabled() ?  "rz-state-disabled" : "")}")'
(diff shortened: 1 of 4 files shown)


iustin94:
One thing that I am not sure here is in the buttons markdown of the steps, the class of the \<a/> tags is set to rz-state-dissabled. The issue here is that is only a visual disablement, and I could still press the element to change the steps. To prevent this, I added the style to the tags also with `"pointer-events: none; cursor: default"` Ideally I would imagine that this style goes in the rz-state-disabled class instead so that other elements behave the same, however I am not sure if it would break any other component behavior.

akorchev:
Hi @iustin94,

Thank you for the pull request. A couple of remarks:

1. Renaming the CanChange event is a breaking change. We won't accept that.
2. You have changed a lot of the formatting in the affected files without actually changing the behavior. It is hard to track the changes that way. Please remove the formatting changes.
3. Using inline styles isn't recommended. You can probably handle this in the code itself.

akorchev:
Instead of using hardcoded inline style (pointer-events: none and opacity) just add a check in the click handler to see if click should be performed or not based on the AllowNext/AllowPrevious values.

iustin94:
@akorchev  Sure but then the link would not change style so for a user they can see and click on it but nothing would happen. Which would lead to confusion, thinking that there is a bug.

iustin94:
Ah right, I see what you mean, that would not be the case as the style would still be applied.

iustin94:
@akorchev I did as you suggested. It is a bit sub optimal since then the function does end up getting called only to return imediatelly. Then the issue that the CanChange event would be expected to allow for steps change to happen, but if the AllowNext or AllowPrevious are set to false that won't have any effect

akorchev:
I checked again and it seems the Steps component would be broken because AllowNext and AllowPrevious are set to false by default. This means the component won't work by default. Can you double check?

iustin94:
@akorchev Changed defaults to true :)

akorchev:
I am now confused how AllowNext and AllowPrevious are supposed to work. Can't you just use the Disabled attribute of RadzenStepsItem? Please describe a concrete scenario that you want to support. Assigning component parameters via reference isn't a good practice and should be avoided. They should only be set via binding (declaratively).

iustin94:
The intent of this change is to make the UX explicit for users when they can and can not move forward.
In this scenario for example, I want the user to first perform a search and not be allowed to click on next before they do that. I could put the search field in a RadzenTemplateForm and give it a RequiredValidator if I must have that value,

billede

From a UX perspective I would have liked to be able to give a user visual feedback that he can move forward, so that when all the fields are filled he can click next because he can see the button become enabled and possibly colored depending on the style.

akorchev:
We have demonstrated a similar implementation in [this demo](https://blazor.radzen.com/steps) - the next step is disabled until the user selects an item. It seems similar to what you are describing and relies only on the Disabled attribute of RadzenStepsItem.

PR #1660, opened 2024-08-19
RadzenTree AllowCheckOnLabelClick parameter

AntoGagne:
This PR add a new parameter to RadzenTree to allow users to click on the label of an item to check its checkbox.

When the parameter is true, the `span` is replaced with a `label` tag. I moved the `for` attribute to this new `label`, as a `span` cannot use it.

I also added a `cursor: inherit;` style to the `.rz-treenode-label` class to overwrite the `cursor: default;` style set by browsers on the `label` tag.

### DIFF
--- a/Radzen.Blazor/RadzenTree.razor.cs
+++ b/Radzen.Blazor/RadzenTree.razor.cs
@@ -183,6 +183,12 @@ namespace Radzen.Blazor
         /// </summary>
         [Parameter]
         public bool AllowCheckParents { get; set; } = true;
+        
+        /// <summary>
+        /// Specifies whether clicking on a item's label will check its checkbox. Set to <c>true</c> by default.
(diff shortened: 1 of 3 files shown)


enchev:
Thanks @AntoGagne! In my opinion it's safe to just change the `span` to `label` without adding new property. Looks like regression made by mistake several months ago by me: https://github.com/radzenhq/radzen-blazor/commit/1023fd90f1cd0cc59c5bf36974701203a5ab5b2c
@yordanov can comment on `cursor: inherit;`

yordanov:
👍🏻  `cursor: inherit;` should be fine.

enchev:
Ah! I've just pushed the change to the master branch directly and just saw your comment, sorry. I think without the property is OK, other cases can be covered with a custom template.

PR #1695, opened 2024-09-16
Correction on how to get scrollTop and scrollLeft on openPopup

antonio272a:
The issue:
When the Dropdown is used after the scroll, (after 100vh), when oppening the dropdown popup, it would consider that the dropdown is at 100vh and would scroll the window up to render the popup above 100vh.

Got this issue on:
- Edge 121.0.2277.128
- OperaGX 113.0.5230.75

To resolve, instead of checking if it's edge to decide from where to get the scrollTop and scrollLeft, it just tries to get from one of them, regardless of the userAgent.

### DIFF
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -1066,13 +1066,8 @@ window.Radzen = {
 
     var parentRect = parent ? parent.getBoundingClientRect() : { top: y || 0, bottom: 0, left: x || 0, right: 0, width: 0, height: 0 };
 
-    if (/Edge/.test(navigator.userAgent)) {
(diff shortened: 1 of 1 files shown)


enchev:
Please provide an example demonstrating the issue. You can use our demos - they are editable.

antonio272a:
The video of the issue happening on edge and correcting it by adding the code of this PR:
https://www.loom.com/share/894f51b5376d454dba09ce00b4f57827?sid=a7bf81bb-cd63-44cd-a195-fb62a7841da5

This is the page, I have this EditorFor component that encapsulates some radzen input components, including RadzenDropdown
image

On this case it renders a RadzenColumn with RadzenFormField and then the RadzenDropdown, I made a demo of it
image

But for some reason this doesn't happen on the Demo.

enchev:
We cannot debug video - we require runnable code.

PR #1696, opened 2024-09-16
Created RadzenDropdown LoadDataOnOpen Parameter

antonio272a:
The issue:
On an app I have a dropdown component that uses RadzenDropdown and searchs on database for the options based on the Value type.
On a specific page of mass creation I have about 100 dropdowns that search on the dabase in generic ways, so it triggers about 100 SQL's in a very short time, what causes the database license to return an error for too many users.

The solution:
In order to not have to handle it manually every time I have this kind of scenario, I added a LoadDataOnOpen parameter to RadzenDropdown that basically does not trigger the LoadData on the dropdown initialization and runs it on the first time that the popup is oppened.

Also created a Demo for test on "/dropdown-load-data-on-open"

### DIFF
--- a/Radzen.Blazor/RadzenDropDown.razor.cs
+++ b/Radzen.Blazor/RadzenDropDown.razor.cs
@@ -79,10 +79,10 @@ namespace Radzen.Blazor
         {
             var disabled = !string.IsNullOrEmpty(DisabledProperty) ? GetItemOrValueFromProperty(item.Item, DisabledProperty) : false;
 
-            var args = new DropDownItemRenderEventArgs<TValue>() 
-            { 
-                DropDown = this, 
-                Item = item.Item, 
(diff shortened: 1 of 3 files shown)


enchev:
Thanks for the pull request @antonio272a however this is breaking change and we cannot accept it - not to mention the changes made to other parts of library not related to this at all.

antonio272a:
I don't see how this is a breaking change, since it's a new parameter that if not used, the behaviour will be exatly the same as it was before.

And it's done only on RadzenDropdown, even if it's used on other components, it'll not change the behaviour if the component do not implement this parameter

enchev:
If you do no load data initially the DropDown will unable to perform lookup and will unable to display what you've set as TextProperty as selected item until you open the popup.

PR #1717, opened 2024-10-01
Added AdditionalContent Parameter to Scheduler

bikotoru:
### 

This PR introduces a new parameter, `AdditionalContent`, to the scheduler component. It allows developers to add custom content next to the year, month, and day buttons, providing more flexibility for adding additional controls or information to the scheduler's header area.

#### Changes:
- Added `AdditionalContent` parameter of type `RenderFragment`.
- Updated documentation comments to describe the parameter's functionality.

This feature improves the customizability of the scheduler and addresses the need for more adaptable UI elements.

Please review the implementation and let me know if there are any changes needed.

### DIFF
--- a/Radzen.Blazor/RadzenScheduler.razor
+++ b/Radzen.Blazor/RadzenScheduler.razor
@@ -20,11 +20,17 @@
         </div>
         <div class="rz-scheduler-nav-title">@SelectedView?.Title</div>
         <div class="rz-scheduler-nav-views">
-        @foreach (var view in Views)
-        {
-            <RadzenButton Click=@(args => OnChangeView(view)) Icon=@view.Icon Text=@view.Text class="@($"{(IsSelected(view)? " rz-state-active" : "")}")" />
-        }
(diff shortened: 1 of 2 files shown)


akorchev:
Hi @bikotoru,

Thank you for the PR. Can we think of a better name than AdditionalContent? It isn't clear where this additional content will go.

bikotoru:
PersonalizedView?
CustomView? 
I'm kind of bad with names haha

akorchev:
What is the purpose of this to begin with? What do you want to add there?

bikotoru:
@akorchev  
For example a Refresh Button

akorchev:
What bothers me is that it renders together with the default navigation (and the name). Somebody would ask for it to render **before** and not **after**. I can propose a different implementation - NavigationTemplate which replaces the whole navigation section - easier to name and provides full control. The existing navigation can be emulated with existing components too. What do you think?

bikotoru:
perfect I have applied the change

PR #1730, opened 2024-10-08
Add same color to datepicker button as text

bikotoru:
I found that when using the InputClass in a datepicker, it only applied to the label text and not to the input text itself. Additionally, there is no way to manipulate the class of the button.

image


enchev:
The InputClass property is applied exactly to the input element:
image

Adding this CSS class to the button will be a breaking change, it's better to add ButtonClass new property that can be used in the button.

bikotoru:
> The InputClass property is applied exactly to the input element: image
> 
> Adding this CSS class to the button will be a breaking change, it's better to add ButtonClass new property that can be used in the button.

@enchev Done, change applied

PR #1745, opened 2024-10-20
Render chart tooltips in the same way as standard tooltips

### LINKED ISSUE #560: LineChart Tooltip near to the browser end doesn't work

maxbruecken:
This PR changes rendering of chart tooltip so, that chart tooltips are shown above all another controls, like the standard tooltips. This PR should fix #560.
Unfortunaltely there are some breaking changes:
- Implementations of IChartSeries and IChartSeriesOverlay in external projects using Radzen must be updated
- RadzenChartTooltip should be included in e.g. MainLayout.razor and ChartTooltipService should bei injected. No changes are necessary if RadzenComponents.razor and .AddRadzenComponents() are already used


akorchev:
Hi @maxbruecken,

Thank you for the PR!

A few questions:

1. Is a tooltip service really needed?
2. If it is needed can't the existing one be used instead of creating a separate one (which seems almost identical). If needed I would suggest extending the existing one with a new method (OpenChartTooltip etc - which could be internal).

maxbruecken:
Hi @akorchev,

thank you for reviewing my PR.
1. yes, the tooltip service ist needed. The chart tooltip component is now outside of the chart, so there should be a service, that can be injected in both components
2. sure, I can move the new method from ChartTooltipService to existing TooltipService and make it internal.
3. I can implement the flipping of the tooltip on the top end of the page, I guess, the method 'openPopup' in Radzen.Blazor.js must be slightly changed, the standard tooltip seems to have the same issue (flipping on the bottom end seems to work):

image

maxbruecken:
Hi @akorchev,

I have moved the OpenChartTooltip method to the TooltipService and removed the ChartTooltipService. The chart tooltip will be shown under the series point now, if there isn't enough space above the series point.

akorchev:
I've made a few things internal. Thanks!

PR #1748, opened 2024-10-22
added autocomplete parameters to RadzenDropDown.

mcgovern-ellsworth:
Autocomplete="off" was set in some circumstances of this control, but not all. Added configurability for both the typing input as well as the filter input.

### DIFF
--- a/Radzen.Blazor/RadzenDropDown.razor
+++ b/Radzen.Blazor/RadzenDropDown.razor
@@ -13,7 +13,7 @@
         <div class="rz-helper-hidden-accessible">
             <input disabled="@Disabled" aria-haspopup="listbox" readonly="" type="text" tabindex="-1"
                    name="@Name" value="@(internalValue != null ? internalValue : "")" id="@Name"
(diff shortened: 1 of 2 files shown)


Radzen.Blazor/RadzenDropDown.razor
```diff
@@ -13,7 +13,7 @@
         <div class="rz-helper-hidden-accessible">
             <input disabled="@Disabled" aria-haspopup="listbox" readonly="" type="text" tabindex="-1"
                    name="@Name" value="@(internalValue != null ? internalValue : "")" id="@Name"
-                   aria-label="@(!Multiple && internalValue != null ? internalValue : EmptyAriaLabel)" @attributes="InputAttributes"/>
+                   aria-label="@(!Multiple && internalValue != null ? internalValue : EmptyAriaLabel)" @attributes="InputAttributes" autocomplete="@InputAutoCompleteType"/>
         </div>
```
enchev, line 17: This input is hidden and the user cannot enter anything.

mcgovern-ellsworth: ah good call. I will remove the InputFilter config hookups.

mcgovern-ellsworth:
Removed.

PR #1749, opened 2024-10-22
Added DropDownTemplate / DropDownTextProperty to specify a different template / property name in the drop down list

Greg-MM:
Added DropDownTemplate to specify a different template in the drop down list vs control
Added DropDownTextProperty to specify a different text property in the drop down list vs control
Uses the normal template / property name if undefined so no breaking changes.

RadzenDropDown_DropDownTemplate_DropDownTextProperty

### DIFF
--- a/Radzen.Blazor/DropDownBase.cs
+++ b/Radzen.Blazor/DropDownBase.cs
@@ -228,11 +228,25 @@ namespace Radzen
         public RenderFragment<dynamic> Template { get; set; }
 
         /// <summary>
+        /// Gets or sets the template in drop down.
+        /// </summary>
+        /// <value>The template in drop down.</value>
+        [Parameter]
+        public RenderFragment<dynamic> DropDownTemplate { get; set; }
+
+        /// <summary>
         /// Gets or sets the value property.
         /// </summary>
         /// <value>The value property.</value>
         [Parameter]
         public string ValueProperty { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value property.
+        /// </summary>
+        /// <value>The value property.</value>
+        [Parameter]
+        public string DropDownTextProperty { get; set; }
 
         /// <summary>
         /// Gets or sets the disabled property.
@@ -449,6 +463,11 @@ namespace Radzen
                     textPropertyGetter = GetGetter(TextProperty, type);
                 }
 
+                if (!string.IsNullOrEmpty(DropDownTextProperty))
+                {
+                    dropDownTextPropertyGetter = GetGetter(DropDownTextProperty, type);
+                }
+
                 if (!string.IsNullOrEmpty(DisabledProperty))
                 {
                     disabledPropertyGetter = GetGetter(DisabledProperty, type);
@@ -471,6 +490,7 @@ namespace Radzen
 
         internal Func<object, object> valuePropertyGetter;
         internal Func<object, object> textPropertyGetter;
+        internal Func<object, object> dropDownTextPropertyGetter;
         internal Func<object, object> disabledPropertyGetter;
 
         /// <summary>
@@ -492,6 +512,10 @@ namespace Radzen
                 if (property == TextProperty)
                 {
                     return textPropertyGetter != null ? textPropertyGetter(item) : PropertyAccess.GetItemOrValueFromProperty(item, property);
+                }
+                else if (property == DropDownTextProperty)
+                {
+                    return dropDownTextPropertyGetter != null ? dropDownTextPropertyGetter(item) : PropertyAccess.GetItemOrValueFromProperty(item, property);
                 }
                 else if (property == ValueProperty)
                 {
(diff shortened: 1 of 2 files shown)


enchev:
I'm afraid that I don't understand how this is different compared to Template/ValueTemplate.

Greg-MM:
When using Template/ValueTemplate I am getting "Operator '==' incompatible with operand types 'String' and 'Zone_VM'" using the below

`<RadzenDropDown @bind-Value="@Location.CurrentZone" Data="@Location.Zones" TValue="Zone_VM" LoadData="@Location.GetZones" FilterCaseSensitivity="FilterCaseSensitivity.CaseInsensitive" AllowFiltering="true">
    <ValueTemplate>
        @((context as Zone_VM).City)
    </ValueTemplate>
    <Template>
       @((context as Zone_VM).ZoneDisplayFriendly)
    </Template>
</RadzenDropDown>`

OR

`<RadzenDropDown @bind-Value="@Location.CurrentZone" Data="@Location.Zones" TValue="Zone_VM" LoadData="@Location.GetZones" ValueProperty="@nameof(Zone_VM.City)" TextProperty="@nameof(Zone_VM.ZoneDisplayFriendly)" FilterCaseSensitivity="FilterCaseSensitivity.CaseInsensitive" AllowFiltering="true">
</RadzenDropDown>`

Location.CurrentZone is an object

however when I use my implementation replacing ValueTemplate with Template and Template with DropDownTemplate it is working, looks like the error is coming from SelectItemFromValue or OnParametersSetAsync in DropDownBase, I'm currently investigating if maybe there is currently a bug , should the above be working with ValueTemplaate / Template?

Greg-MM:
Seems like line 1252 in DropDownBase (SelectItemFromValue) when T is an object is causing the error comparing the object to whatever type ValueProperty is instead of comparing it to the ValueProperty of value (ex Foo.value == Bar instead of Foo.value == Bar.value), changing

`SelectedItem = view.AsQueryable().Where(DynamicLinqCustomTypeProvider.ParsingConfig, $@"{ValueProperty} == @0", value).FirstOrDefault();`

to the below does seem to make Template / ValueTemplate work like this PR for me, let me know if you want me to make a PR for this, I don't think it should break anything, seems like there are some other functions that have this bug as well (UpdateSelectedItems, line 1241), but I haven't done a full search yet

`SelectedItem = view.AsQueryable().Where(DynamicLinqCustomTypeProvider.ParsingConfig, $@"{ValueProperty} == @0", GetItemOrValueFromProperty(value, ValueProperty)).FirstOrDefault();`

enchev:
Template/ValueTemplate gives you the ability to customize what should be displayed in the DropDown or when the value is chosen. TextProperty gives you the same if the templates are not defined and it will be used for filtering. ValueProperty is used when you need to perfect lookup using @bind-Value.

akorchev:
@Greg-MM I think you just need to set TextProperty and your scenario should work without any changes.

Greg-MM:
I tried that, the error comes in when bind-Value is not a string, but an object.
Changing String value to Customer value on the "Define Text & Value properties" demo and hitting run will cause the InvalidCastException upon selecting a value, so I think the above changes in the comment changing
value to GetItemOrValueFromProperty(value, ValueProperty) will allow binding the DropDown value to an object instead of just a string

`@using RadzenBlazorDemos.Models.Northwind

@inherits DbContextPage

<RadzenStack Orientation="Orientation.Horizontal" AlignItems="AlignItems.Center" JustifyContent="JustifyContent.Center" Gap="0.5rem" class="rz-p-sm-12">
    <RadzenLabel Text="Select Value" Component="DropDownTextValueProperties" />
    <RadzenDropDown @bind-Value=@value Data=@customers TextProperty="@nameof(Customer.CompanyName)" ValueProperty="@nameof(Customer.CustomerID)" Style="width: 100%; max-width: 400px;" Name="DropDownTextValueProperties" />
</RadzenStack>

@code {
    Customer value;
    IEnumerable<Customer> customers;

    protected override async Task OnInitializedAsync()
    {
        await base.OnInitializedAsync();

        customers = dbContext.Customers;
    }
}`

enchev:
When binding to the entire object you should not set ValueProperty:
```
@using RadzenBlazorDemos.Models.Northwind

@inherits DbContextPage

<RadzenDropDown @bind-Value=@value Data=@customers TextProperty="@nameof(Customer.CompanyName)"  />

@code {
    Customer value;
    IEnumerable<Customer> customers;

    protected override async Task OnInitializedAsync()
    {
        await base.OnInitializedAsync();

        customers = dbContext.Customers;
    }
}
```

Greg-MM:
Awesome, thanks a lot its working now, would you be interested in a PR with code that throws an exception when ValueProperty is defined and T is not String, or maybe just null out the property, was trying to think of a way of detecting at compile time, best bet would probably be some sort of Roslyn analyzer, but I haven't done any of that and don't have time to go down that rabbit hole at the moment.

Also looking to implement this same functionality of binding to objects, different Template and ValueTemplate to the AutoComplete control by making it a generic class as it more fits the free form entry that I am going for, is this something that you would want a PR for, I'm really only testing in WASM so it would need some decent testing on Server / Hybrid environments as it will likely be much more significant changes to that class?

enchev:
We prefer not to throw such exception, also the DropDown component is already enough complex and we would like to avoid more coupling with other components - the AutoComplete component also don’t have ValueProperty. I’m closing this PR - use our forum in the future to avoid unnecessary coding.

PR #1768, opened 2024-11-02
Dialog context help button

Techoneshot:
Implements optional parameter that shows help button on title bar which opens passed url in new tab


enchev:
You can use TitleContent to add such button:
image
image

PR #1791, opened 2024-11-15
Update Radzen.Blazor.js

Ezekiel768:
Fix startup exception


akorchev:
We won't accept this pull request as it adds unnecessary delay and defeats the purpose of preloading the icons. We will add a check for document.body instead for people that insist not to follow our instructions where to include the script.

PR #1805, opened 2024-11-21
Modify tooltip positioning and tooltip triangle pointer positioning

OleFosdal:
- Determine `tooltip-content` css-class solely in javascript method `openPopup()`. 
- Added support for left/right alligned triangle pointer for top/bottom tooltip.
- Added support for top/bottom alligned triangle pointer for left/right tooltip.
- Recalculate left position for bottom/top tooltips if they extend right side of screen.
- Recalculate top position for left/right tooltips if they extend bottom side of screen.

I made this solution before Commit a0f1545 ("Tooltip position updated when forced to change"), so unfortunately i have not implemented that current fix/solution. However i dont see the need for that with this current code.

Have tested DateGrid/Chart popups too and have tried to avoid configuring theese as much as possible. 
The "reset" logic of removing all css-classes before adding new might be able to be done more cleanly.


enchev:
Have you tested the fix I've pushed? Your changes are might affect drastically all components that are using popups - DataGrid, ContextMenu, DropDown/DropDownDataGrid, DatePicker, AutoComplete, etc.

OleFosdal:
Yea i feared that as well. 
So even though it might not merge because of that, its a showcase for what i wished the tooltip might be able to do. Maybe in the future where there is more time to adjust. 

The fix you showed still has some css problems with the triangle-pointer (hence why i try to modify the logic behind the css too). 
image

In general i would have hoped that the "smartPosition" would have worked in a way where no position is given to the method, and it will auto-calculate where to position the tooltip with some priority maybe (first try bottom, then top etc.). 
But when i define that it should be positioned bottom i would have hoped it stayed at the bottom and knowing the potential consequenses of it.

enchev:
Here are several issues I've found so far with this pull request:
image
image
image

In my opinion at the moment these changes will not solve the problems reported by you on the forum and will cause other problems with multiple components.

OleFosdal:
Did not know how to build for Material-base3. So try it in default theme or any free theme i assume.

OleFosdal:
So far i have tested all the popups i know of and they seem to be working now. They work like they use to (still smartPositions on both y- and x-axis when needed to. Aand only tooltips behave in a way of the new code. 

But then again i know it is changing some fundamental code for all popups and there might be unforseen bugs. So understand why this might not be a viable solution.

enchev:
I've verified once again the case you've reported in [this forum thread](https://forum.radzen.com/t/tooltip-smartposition/19104/6) to confirm that the current state in the master branch will fix the problem and since the code in openPopup function is used by various components and might cause multiple potential problems, I've decided that we should not risk to merge it.

OleFosdal:
It does not fix the css showed in [this comment](https://github.com/radzenhq/radzen-blazor/pull/1805#issuecomment-2491556134) however. 

Anyway, i will try use this modified code instead. Feel free to come back and ask me if ive encounterd additional problems. 
I know it might be a pain for me to keep code updated with any changes you make to the openPopup method.

PR #1811, opened 2024-11-25
Add FilterSortOrder to RadzenDataFilterProperty

tale80:
Introduced a FilterSortOrder parameter to control the sorting order of filters in RadzenDataFilterProperty. Updated relevant components to utilize and display the correct property sort order. This change ensures that filters are organized as specified by their sort order values.

FilterSortOrder start at 100 to allow adding priority by just adding a small value to some filters and not enforce to set the order on all

### DIFF
--- a/Radzen.Blazor/RadzenDataFilter.razor.cs
+++ b/Radzen.Blazor/RadzenDataFilter.razor.cs
@@ -294,15 +294,16 @@ namespace Radzen.Blazor
         /// Gets the properties collection.
         /// </summary>
         /// <value>The properties collection.</value>
-        public IList<RadzenDataFilterProperty<TItem>> PropertiesCollection
+        public List<RadzenDataFilterProperty<TItem>> PropertiesCollection
         {
             get
             {
-                return properties;
+                return properties.OrderBy(r => r.FilterSortOrder).ToList();
             }
         }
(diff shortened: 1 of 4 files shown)


enchev:
Thanks for the pull request @tale80 however I'm not sure what problem resolves this, can you clarify? Furthermore if you want to sort in a specific way filter properties you can always  create them in a for loop with specific order.

tale80:
@enchev : as you wish...
For a real use case, where you set a filter for a complex VM, with property to filter and not, with nested VM, and sub component, your solution will not work.

If you setsomehting like that
``` 
<RadzenDataFilter Data="@orders">
    <Properties>
        <RadzenDataFilterProperty Property="@item.Property1" Title="@item.Title" />
        <RadzenDataFilterProperty Property="@item.Property2" Title="@item.Title2" />
        <CustomDataFilterComponent Property="@item.Property3" Title="@item.Title3" />
        <RadzenDataFilterProperty Property="@item.Property4" Title="@item.Title4" />
    </Properties>
</RadzenDataFilter>
```

Your drop down will end with
Title
Title2
Title4
Title3

The custom component is added after all "direct" properties.

Too bad, I will override again a lot of components to get order in drop down

PR #1812, opened 2024-11-26
Handle nullable date type in queryable extension.

tale80:
Added a check to handle properties of nullable date types by accessing their `Value` property when filtering. This ensures that the filtering logic works correctly with nullable dates.

This PR correct an incorrect request built by the component if the filter need to be applied on a Nullable or Nullable in database

### DIFF
--- a/Radzen.Blazor/QueryableExtension.cs
+++ b/Radzen.Blazor/QueryableExtension.cs
@@ -434,6 +434,11 @@ namespace Radzen
             }
             else if (PropertyAccess.IsDate(columnType))
             {
+                if (columnType.GetGenericTypeDefinition() == typeof(Nullable<>))
+                {
+                    property = $"{property}.Value";
+                }
+                
                 var v = column.FilterValue;
                 if (v != null)
                 {


enchev:
OrderDate in [this demo](https://blazor.radzen.com/datafilter?theme=material3) is nullable, can you clarify what is fixing this request?
image

tale80:
@enchev  it's because you're currently executing it in IQueryable only in the view.
It can work only if you have a small dataset and provide all data to the filter component. In real world with more data, the request must be executed on server side, and probably on the DB directly for most developers.

If you get your request with ToFilterString, send it to the server and request in DB (for example with DynamicLinq), you will get this error on nullable DateTime/DateTimeOffset (maybe true for others nullable type, I haven't check on all field type for now) : 
```
 System.InvalidOperationException: The binary operator Equal is not defined for the types 'System.Nullable`1[System.DateTimeOffset]' and 'System.DateTime'.
```

Same for all others operator (NotEquals, GreaterThan, LessThan, GreaterThanOrEquals, LessThanOrEquals)

enchev:
I'm afraid that without runnable code demonstrating the problem we will unable to test and merge your pull request.

tale80:
Weird... I will test without the fix on a nullable DateTime?, I have the trouble on a nullable DateTimeOffset, maybe the behavior is different on both

PR #1826, opened 2024-12-02
 implement missing dateonly check for parsing value of DateOnly

simonluebker:
In the filter string concatenation for OData  was the DateOnly type missing. For the nullable DateOnly it worked properly.


simonluebker:
Hey there,
why is this closed without comment? Please give feedback

best  regards,
Simon

Issue #1675, opened 2024-09-03
Rounded corners in stacked column chart do not display correctly


RichardPoes:
**Describe the bug**
When supplying data to the stacked bar chart, where one of the top-most bars has a length of 0, no rounding is applied, as show in the screenshot

**To Reproduce**
I have created a gist with a minimum working example: https://gist.github.com/RichardPoes/1662357aa582454198b08ce04e1cf51e
This code can be pasted inside the code tab of the Radzen Blazor documentation page of the stacked column chart https://blazor.radzen.com/stacked-column-chart

**Expected behavior**
I would expect the topmost part of the bar to be always rounded, regardless of whether there being other 0-valued categories stacked upon it.

**Screenshots**
![image](https://github.com/user-attachments/assets/97ab96f0-4f19-4b30-b9c4-d603535fd34b)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Google Chrome
 - Version: 128.0.6613.114 (Official Build) (64-bit)

**Additional context**

enochwhitaker3:
I think I've got this fixed and I will be making a PR when I get the code to be a little cleaner

PR #1829, opened 2024-12-02
Fixing issue #1675: Rounded corners in stacked column chart do not display correctly 

enochwhitaker3:
I fixed the issue where the radius would not be applied to a column if the top columns value = 0, making the radius appearing not to be applied


akorchev:
Hi @enochwhitaker3,

Thank you for the pull request!

Could you please keep the original formatting as much as possible? A lot of unrelated code has changed and we don't want most of the file to blame to you.

akorchev:
Just fixed it as it also had to be addressed for stacked bar charts as well. Thanks once again!

PR #1837, opened 2024-12-05
add AllowMoveAllItems  to picklist

nielsNocore:
implemented: add property AllowMoveAllItems to picklist. to be able to optionally show/hide the move all buttons.


enchev:
Thanks for the idea @nielsNocore! I've extended it with two more properties to allow separate setting for each button as well:
https://github.com/radzenhq/radzen-blazor/commit/da43d91b5a0275abf11f0545d52aae5eff2b6ed2

Issue #1842, opened 2024-12-08
RadzenDataFilter Numeric Input Issue when switching between And - Or operators


kerajel:
Proposed resolution: https://github.com/radzenhq/radzen-blazor/pull/1843

Another issue related to the 'Blazor server lag' with numeric inputs

When you input something in the numeric input (other input types work fine) and then switch between And - Or operators of the RadzenDataFilter without explicitly triggering the 'onchange' event, the input is lost:

![Recording 2024-12-09 030522](https://github.com/user-attachments/assets/2a64849e-c9f1-43cd-9d60-d79f6106f896)

Radzen.Blazor Version=5.6.12

Issue #1818, opened 2024-11-28
RadzenDataFilter Numeric Input Issue


kerajel:
RadzenDataFilter numeric input suffers the same issues as RadzenDataGrid's AdvancedFilter on numeric columns as described in https://github.com/radzenhq/radzen-blazor/issues/1809 (that is, the issue is reproduced only on the hosted environments, influenced by Blazor server lag)

**To reproduce**

Negative case
Define a RadzenDataFilter with numeric column
Input a value and filter right away

**Expected result**
Input value is registered in the RadzenDataFilter<>.ToFilterString();

**Actual result**
Input value is NOT registered in the RadzenDataFilter<>.ToFilterString();

Positive case
Define a RadzenDataFilter with numeric column
Input a value and DO NOT filter right away
Focus something on the page
Do filter

**Expected result** = **Actual Result**
Input value is registered in the RadzenDataFilter<>.ToFilterString();

kerajel:
The following script helped to resolve the issue:

```js
function forceUpdateNumericInputs() {
    document.querySelectorAll('.rz-numeric-input input').forEach(input => {
        input.blur();
        input.focus();
    });
}
```

```c#
async Task Submit() // whatever action that leads to reading RadzenDataFilter<>.ToFilterString();
{
    await JSRuntime.InvokeVoidAsync("forceUpdateNumericInputs");
    string filterString = @dataFilter.RadzenDataFilter<>.ToFilterString(); // returns current input
}
```

enchev:
Fixed in https://github.com/radzenhq/radzen-blazor/commit/a0cfc5d26736c5d7f7397356818ceef6059ba152

PR #1843, opened 2024-12-08
RadzenDataFilter numeric input fixes

kerajel:
I wish there was a better way to fix these issues, but since the numeric input relies on the 'onchange' event and not on the 'oninput' event it's hard to come up with something. If there is a better way to force an input to 'commit' its changes into the 'onchange' event on demand, it'd be best to use it, but I am not aware of other solutions.

Resolving https://github.com/radzenhq/radzen-blazor/issues/1818
blurAllNumericInputs - this functions should be called from client's code right before you want to read RadzenDataFilter<>.ToFilterString();

Resolving https://github.com/radzenhq/radzen-blazor/issues/1842
Reference the new OnFilterOperatorChanged method (its logic was inlined before). The loss of values occurs at the ```await InvokeAsync(ChangeState);``` because the 'onchange' event of the corresponding numeric input has not committed and server-side DOM is not aware of user's input. Temporarily stashing the current input values and then restoring them resolves the issue.  Other blurring techniques did not work in this case.

### DIFF
--- a/Radzen.Blazor/RadzenDataFilter.razor
+++ b/Radzen.Blazor/RadzenDataFilter.razor
@@ -12,7 +12,7 @@
 @if (Visible)
 {
     <div @ref="@Element" style="@Style" @attributes="Attributes" class="@GetCssClass()" id="@GetId()">
-        <RadzenSelectBar @bind-Value="LogicalFilterOperator" Change="EventCallback.Factory.Create<LogicalFilterOperator>(this, OnFilterOperatorChanged)" Size="ButtonSize.Small" class="rz-datafilter-operator-bar">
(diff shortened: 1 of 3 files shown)


enchev:
I don't think that we should merge this code  - the DataGrid filtering should not mess with all input components in the current page in first place. I still believe that handling internally oninput is the only reliable way to avoid such problems with Blazor server lag:
https://github.com/radzenhq/radzen-blazor/commit/124033d122ba981f554498c3b8cefdd9a0bbe9e7

kerajel:
@enchev, but how would I handle the 'oninput' if the numeric input uses 'onchange'?

enchev:
Check the code I’ve linked. Both can coexist.

enchev:
Here is what I'm refering:
https://github.com/radzenhq/radzen-blazor/commit/1a40da99aa6d2bae88e2810421b71e8f8aa5d9c2

PR #1852, opened 2024-12-12
Scheduler New View - MultiDay

paulo-rico:
Working in the same way as `RadzenWeekView` with these exceptions -

1. Will use actual `StartDate `as opposed to `StartDate.StartOfWeek`
2. New property for this view `NumberOfDays` (defaulted to 2) that dictates how many days to show.  `EndDate` calculated using this property. Also used for Previous / Next calculation.

Added a new demo section in the Scheduler demo page to highlight this.

Regards
Paul

### DIFF
--- a/Radzen.Blazor/RadzenMultiDayView.razor
+++ b/Radzen.Blazor/RadzenMultiDayView.razor
@@ -1,0 +1,15 @@
+@using Radzen.Blazor
+@using Radzen.Blazor.Rendering
+
+@inherits SchedulerViewBase
+
+@code {
+    public override RenderFragment Render()
+    {
+        var appointments = Scheduler.GetAppointmentsInRange(StartDate, EndDate).ToList();
+
+        return @<CascadingValue Value=@Scheduler>
+                <WeekView HeaderFormat=@HeaderFormat MinutesPerSlot=@MinutesPerSlot StartDate=@StartDate EndDate=@EndDate StartTime=@StartTime EndTime=@EndTime Appointments=@appointments TimeFormat="@TimeFormat" AppointmentMove=OnAppointmentMove />
+            </CascadingValue>;
+    }
+}
--- a/Radzen.Blazor/RadzenMultiDayView.razor.cs
+++ b/Radzen.Blazor/RadzenMultiDayView.razor.cs
@@ -1,0 +1,106 @@
+using Microsoft.AspNetCore.Components;
+using Radzen.Blazor.Rendering;
+using System;
+
+namespace Radzen.Blazor
+{
+    /// <summary>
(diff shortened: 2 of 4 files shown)


akorchev:
Hi @paulo-rico,

Thank you for the pull request. I have seen people requesting this before. However they wanted to set a different start day of the week. For example from Tuesday to Friday. Can this be supported at the moment?

paulo-rico:
Hi @akorchev 
Can't see why not. It would mean that the developer would have to work out the `StartDate` from the given date and just set the `NumberOfDays` to seven.

akorchev:
I think this would be the better API - StartDay and EndDay instead of StartDate calculation shenanigans + NumberOfDays. This would even allow "exotic" views such as Sunday to Wednesday.

akorchev:
Responding to the edits.

Maybe startday and endday should be weekview's properties. I am not so sure about the need for a multi-day view on its own if we do that. Are there any scenarios where users would need that?

paulo-rico:
TBH, when I was putting this together, it occurred to me that, with a couple of properties on the `WeekView`, we could make it way more flexible, accommodating scenarios that you mention and the scenario that prompted me to make this. The defaults for these properties rendering the `WeekView` that is already in place.

My major concern was that the name `WeekView` didn't really fit anymore. It would be more of a `MultiDay` view.

Personally, I would prefer not to have the additional view and to make the `WeekView` more flexible.

What do you think?

akorchev:
>  I need to be able to view just "today" and the next two or three days.

This is indeed a better fit for a new view - multi-day view.

paulo-rico:
I've been giving this some thought. Here's what I've come up with. 

The `MultiDayView` is valid in it's own right, but I'm not happy with the code duplication. The `MultiDayView` is basically a copy of the `WeekView` with the shackles off. I still think that `WeekView` is a valid component, even though you can accomplish the same with the `MultiDayView` with some ectra work from the developer.

My proposal is this -

1. We add in the `MultiDayView` initially, and ensure that all is fine. 
2. At a later date, when happy that this view is behaving, we change the implementation of the `WeekView` so that the markup is just the `MultiDayView` with the shackles on, i.e., the `StartDate = Scheduler.CurrentDate.Date.StartOfWeek()` and `NumberOfDays = 7`. In theory, this should work and will resolve the code duplication.

What do you think?

akorchev:
Sounds OK to me. Does this mean there is no more work to do in this PR?

paulo-rico:
It does. I've just Pushed again (clean up), removing a comment marker that's not required

paulo-rico:
Sorry about this @akorchev 

I have overlooked one thing. I haven't changed the icon for this view. It's the same as `MonthView`

This is what I was going to put there -

`        public override string Icon => "width_normal";
`

Are you able to change it your end, or shall I open another PR?

PR #1861, opened 2024-12-17
Added the AutoClose property to the Accordion. 

marcelaopimenta:
Added the AutoClose property to the Accordion. By default, it is true. If the user clicks on an already open item, it closes. If it is false, if the user clicks on an already open item, it remains open.

### DIFF
--- a/Radzen.Blazor/RadzenAccordion.razor.cs
+++ b/Radzen.Blazor/RadzenAccordion.razor.cs
@@ -68,6 +68,13 @@ namespace Radzen.Blazor
         public EventCallback<int> Collapse { get; set; }
 
         /// <summary>
+        /// Gets or sets a property to prevent collapse expanded item.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


enchev:
Thanks for the pull request @marcelaopimenta however I don't remember any other users needing such feature. Can you provide a bit more details? You don't want already expanded item to be collapsed?

marcelaopimenta:
Hi Vladimir, how are you!?

In my case I need one item of the accordion to always be open. Here is an
example. In a payment form the user must choose one. Credit Card or PIX
(Braziian Payment)

[image: image.png]
Or

[image: image.png]

Another user needed this, but for another business rule.
https://forum.radzen.com/t/prevent-accordian-component-from-closing/3662

Best Regards

Marcelo Fernandes
======================================
Gtalk ou Email: ***@***.***
Skype: marcfsp

PR #1894, opened 2025-01-04
Support for inserting after specific row in DataGrid

ENeMy235:
Added method in DataGrid to allow to insert after specific row: 
`InsertAfterRow(TItem itemToInsert, TItem rowItem)`

Code adjusted, to keep normal insert intact (at top of the grid).

Updated demo page **DataGridInLineEdit**, to show the option of inserting after specific row.


Kebechet:
This would really help us in our use-case as well.

PR #1904, opened 2025-01-11
Appointment Resize

paulo-rico:
Hi Team

This is a PR that addresses resizing appointments in the Scheduler.

I've only implemented this for `DayView`, `MultidayView` and `WeekView`. The other views show partial appointments as occupying full slots. It's not really applicable as is stands.

I've tried to comment all aspects of the changes so you can follow the logic.

Apart from the standard summary / documentation, I've not added any comments to `Common.cs`, `IScheduler.cs` and `RadzenScheduler.razor.cs` as the changes are straightforward boilerplate to service Events and such.

The bulk of the changes lie in `Appointment.razor` and `Radzen.Blazor.js`, which I have heavily commented.

You will notice that I've included a `<style>` block within `Appointment.razor`. I couldn't find how to add this properly, so I must leave it with you to transfer to the appropriate place (or advise how I do that).

The first demo on the Scheduler page is the one that implements `AppointmentResize`.

Stay well and regards
Paul

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -739,6 +739,27 @@ namespace Radzen
     }
 
     /// <summary>
+    /// Supplies information about a <see cref="DropableViewBase.AppointmentMove" /> event that is being raised.
+    /// </summary>
+    public class SchedulerAppointmentResizeEventArgs
+    {
+        /// <summary>
+        /// Gets or sets the appointment data.
+        /// </summary>
+        public AppointmentData Appointment { get; set; }
+
+        /// <summary>
+        /// Gets or sets the start DateTime.
+        /// </summary>
+        public DateTime Start { get; set; }
+
+        /// <summary>
+        /// Gets or sets the end DateTime.
+        /// </summary>
+        public DateTime End { get; set; }
+    }
+
+    /// <summary>
     /// Supplies information about a <see cref="RadzenMenu.Click" /> event that is being raised.
     /// </summary>
     public class MenuItemEventArgs : MouseEventArgs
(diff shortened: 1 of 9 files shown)


akorchev:
Hi @paulo-rico,

I wish you had discussed this with us before starting work on it. We prefer to implement it ourselves when we have the chance.

paulo-rico:
Hi @akorchev 

I must say, with the changes required to carry out this PR, I was dubious about whether you would accept. It probably is a bridge too far and I fully understand why you would develop it in-house.

But I never see these things as a waste of time with regards personal development. I've learnt more about the Radzen library and I treat programming as a hobby, receiving great pleasure when I complete a task that is outside my usual comfort zone.

Is it you or I that Closes the PR?

While I have your attention, I have an unrelated change I would like to add.

In the same way we added the `OnMonthClick` event for the Year Views, I would like to add an `OnDayClick` to the `MonthView` (on the Day number) and the `Week` / `Multiday` (on the Day header). Does that sound acceptable?

If I have any more suggestions for PR's in the future, where's the best place to raise them, as a GitHub Issue or on the Radzen Forum?

Cheers and stay well, Atanas

Regards
Paul

akorchev:
> In the same way we added the OnMonthClick event for the Year Views, I would like to add an OnDayClick to the MonthView (on the Day number) and the Week / Multiday (on the Day header). Does that sound acceptable?

Yes, this sounds reasonable. Just have in mind that we don't add an "On" prefix for events. 

> If I have any more suggestions for PR's in the future, where's the best place to raise them, as a GitHub Issue or on the Radzen Forum?

Github issues is better IMHO.

PR #1910, opened 2025-01-15
Add DaySelect event to Day, Week and Month Views

paulo-rico:
Similar to the `MonthSelect` on the various Year Views, this will raise a `SelectDay` event on a click of the Day Header (Day and Week View) or the Day Number (Month View).

Additional

Also in this PR, I had to change a line in the `OnKeyPress` event in `DayView.razor`. 

    `currentAppointments = AppointmentsInSlot(CurrentDate, CurrentDate.AddDays(1));`

This was taking the `CurrentDate`, which includes the `StartTime` and setting the range to `CurrentDate.AddDays(1)`. This resulted in (using default `StartTime` value) from 08:00 one day to to 08:00 the next day. All other range calculations are based on actual start of day to end of day.

### DIFF
--- a/Radzen.Blazor/RadzenScheduler.razor.cs
+++ b/Radzen.Blazor/RadzenScheduler.razor.cs
@@ -206,6 +206,24 @@ namespace Radzen.Blazor
         public EventCallback<SchedulerMonthSelectEventArgs> MonthSelect { get; set; }
 
         /// <summary>
+        /// A callback that will be invoked when the user clicks a month header button.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// &lt;RadzenScheduler Data=@appointments MonthSelect=@OnMonthSelect&gt;
+        /// &lt;/RadzenScheduler&gt;
+        /// @code {
+        /// void OnMonthSelect(SchedulerTodaySelectEventArgs args)
+        /// {
+        ///     args.Month = DateTime.Month.AddMonth(1);
+        /// }
+        /// }
+        /// </code>
+        /// </example>
+        [Parameter]
+        public EventCallback<SchedulerDaySelectEventArgs> DaySelect { get; set; }
+
+        /// <summary>
         /// A callback that will be invoked when the user clicks an appointment in the current view. Commonly used to edit existing appointments.
         /// </summary>
         /// <example>
@@ -392,6 +410,12 @@ namespace Radzen.Blazor
         public async Task SelectMonth(DateTime monthStart, IEnumerable<AppointmentData> appointments)
         {
             await MonthSelect.InvokeAsync(new SchedulerMonthSelectEventArgs { MonthStart = monthStart, Appointments = appointments, View = SelectedView });
+        }
+
+        /// <inheritdoc />
+        public async Task SelectDay(DateTime day, IEnumerable<AppointmentData> appointments)
+        {
+            await DaySelect.InvokeAsync(new SchedulerDaySelectEventArgs { Day = day, Appointments = appointments, View = SelectedView });
         }
 
         /// <inheritdoc />
--- a/Radzen.Blazor/Rendering/DayView.razor
+++ b/Radzen.Blazor/Rendering/DayView.razor
@@ -2,7 +2,7 @@
 @inherits DropableViewBase
 
 <div class="rz-view rz-day-view">
-    <div class="rz-view-header">
+    <div class="rz-view-header" style="cursor: pointer;" @onclick=@(args => OnDayClick(@StartDate))>
         <div class="rz-slot-hour-header"></div>
         <div class="rz-slot-header">
             @StartDate.ToString("ddd", Scheduler.Culture)
@@ -64,6 +64,11 @@
         await Scheduler.SelectSlot(date, date.AddMinutes(MinutesPerSlot), AppointmentsInSlot(date, date.AddMinutes(MinutesPerSlot)));
     }
 
+    async Task OnDayClick(DateTime day)
+    {
+        await Scheduler.SelectDay(day, AppointmentsInSlot(day.StartOfDay(), day.EndOfDay()));
+    }
+
     IDictionary<string, object> Attributes(DateTime date)
     {
         var attributes = Scheduler.GetSlotAttributes(date, date.AddMinutes(MinutesPerSlot));
@@ -84,7 +89,7 @@
         var updown = key == "ArrowUp" || key == "ArrowDown";
         var leftright = key == "ArrowLeft" || key == "ArrowRight";
 
-        currentAppointments = AppointmentsInSlot(CurrentDate, CurrentDate.AddDays(1));
+        currentAppointments = AppointmentsInSlot(CurrentDate.StartOfDay(), CurrentDate.EndOfDay());
 
         if (arrow && currentAppointments.Any())
         {
--- a/Radzen.Blazor/Rendering/MonthView.razor
+++ b/Radzen.Blazor/Rendering/MonthView.razor
@@ -82,7 +82,7 @@
             {
                 var dayOfWeek = StartDate.AddDays(days++);
                 <div @onclick="@(args => OnSlotClick(dayOfWeek))" ondragover="event.preventDefault();" @ondrop=@(args => @OnDrop(dayOfWeek)) @attributes=@Attributes(dayOfWeek)>
-                    <div class="rz-slot-title @(dayOfWeek == CurrentDate ? " rz-state-focused" : "")">
+                    <div class="rz-slot-title @(dayOfWeek == CurrentDate ? " rz-state-focused" : "")" style="cursor: pointer;" @onclick=@(args => OnDayClick(dayOfWeek)) @onclick:stopPropagation>
                         @dayOfWeek.Day
                     </div>
                 </div>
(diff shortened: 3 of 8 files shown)


Radzen.Blazor/RadzenScheduler.razor.cs
```diff
@@ -205,6 +205,24 @@ public partial class RadzenScheduler<TItem> : RadzenComponent, IScheduler
...
+        /// void OnMonthSelect(SchedulerTodaySelectEventArgs args)
+        /// {
+        ///     args.Month = DateTime.Month.AddMonth(1);
+        /// }
+        /// }
+        /// </code>
+        /// </example>
+        [Parameter]
+        public EventCallback<SchedulerDaySelectEventArgs> DaySelect { get; set; }
+
```
akorchev, line 225: The code example seems to be incorrect - refers to MonthSelect.

Radzen.Blazor/Rendering/DayView.razor
```diff
@@ -2,7 +2,7 @@
 @inherits DropableViewBase
 
 <div class="rz-view rz-day-view">
-    <div class="rz-view-header">
+    <div class="rz-view-header" style="cursor: pointer;" @onclick=@(args => OnDayClick(@StartDate))>
```
akorchev, line 5: Using `style="cursor: pointer"` should be avoided. Either update the rz-view-header CSS class or leave it as it is.

Radzen.Blazor/Rendering/DayView.razor
```diff
@@ -84,7 +89,7 @@
         var updown = key == "ArrowUp" || key == "ArrowDown";
         var leftright = key == "ArrowLeft" || key == "ArrowRight";
 
-        currentAppointments = AppointmentsInSlot(CurrentDate, CurrentDate.AddDays(1));
+        currentAppointments = AppointmentsInSlot(CurrentDate.StartOfDay(), CurrentDate.EndOfDay());
```
akorchev, line 92: This looks as a breaking change as EndOfDay method is different than AddDays(1). Why is this change required?

Radzen.Blazor/Rendering/MonthView.razor
```diff
@@ -82,7 +82,7 @@
             {
                 var dayOfWeek = StartDate.AddDays(days++);
                 <div @onclick="@(args => OnSlotClick(dayOfWeek))" ondragover="event.preventDefault();" @ondrop=@(args => @OnDrop(dayOfWeek)) @attributes=@Attributes(dayOfWeek)>
-                    <div class="rz-slot-title @(dayOfWeek == CurrentDate ? " rz-state-focused" : "")">
+                    <div class="rz-slot-title @(dayOfWeek == CurrentDate ? " rz-state-focused" : "")" style="cursor: pointer;" @onclick=@(args => OnDayClick(dayOfWeek)) @onclick:stopPropagation>
```
akorchev, line 85: Same as before - `style="cursor: pointer"` should be avoided.

Radzen.Blazor/Rendering/WeekView.razor
```diff
@@ -8,7 +8,8 @@
         <div class="rz-slot-hour-header"></div>
     @for (var day = StartDate; day < EndDate; day = day.AddDays(1))
     {
-        <div class="rz-slot-header">
+        var loopDay = day;
+        <div class="rz-slot-header" style="cursor: pointer;" @onclick=@(args => OnDayClick(@loopDay))>
```
akorchev, line 12: Avoid `style="cursor: pointer"`.

RadzenBlazorDemos/Pages/SchedulerConfig.razor
```diff
@@ -34,6 +34,12 @@
         new Appointment { Start = DateTime.Today.AddDays(1), End = DateTime.Today.AddDays(12), Text = "Vacation" },
     };
 
+    async Task OnDaySelect(SchedulerDaySelectEventArgs args)
```
akorchev, line 37: Just use `void` instead of `async Task` - would make the code simpler and avoid the need of "await Task.CompletedTask.

RadzenBlazorDemos/Pages/SchedulerConfig.razor
```diff
@@ -34,6 +34,12 @@
...
+    {
+        console.Log($"DaySelect: Day={args.Day} AppointmentCount={args.Appointments.Count()}");
+        await Task.CompletedTask;
```
akorchev, line 40: The same - if not using async code just use `void` for the event handler.

akorchev (changes requested):

Radzen.Blazor/RadzenScheduler.razor.cs
paulo-rico: Resolved

RadzenBlazorDemos/Pages/SchedulerConfig.razor
paulo-rico: Resolved

RadzenBlazorDemos/Pages/SchedulerConfig.razor
paulo-rico: Resolved

Radzen.Blazor/Rendering/DayView.razor
paulo-rico: Resolved - I've removed the style setting. I don't actually know how to add it to the `rz-view-header` CSS class. That facility seems to be locked out.  The css directory being set to git 'Ignored'.

Radzen.Blazor/Rendering/MonthView.razor
paulo-rico: Resolved

Radzen.Blazor/Rendering/WeekView.razor
paulo-rico: Resolved

Radzen.Blazor/Rendering/DayView.razor
akorchev: The styles are defined in the SASS files - for example `_scheduler.scss`.

Radzen.Blazor/Rendering/DayView.razor
paulo-rico: This has highlighted an error in my code (this, and some of the other changes), so I've remedied them. But I still think there is an error with this line - 

`currentAppointments = AppointmentsInSlot(CurrentDate, CurrentDate.AddDays(1));`

Whenever we get `Appointments`, we use the `StartDate` and `EndDate` which are built from the `Scheduler.CurrentDate.Date` and adding on `StartTime` and `EndTime` respectively.

I now think it should be -

```     
        var dayStart = CurrentDate.Date.Add(StartTime);
        var dayEnd = CurrentDate.Date.Add(EndTime);
        currentAppointments = AppointmentsInSlot(dayStart, dayEnd);
```

so that we get Appointments using the same range as elsewhere in the code. 

From what I see, what will happen is that, for gathering Appointments to render, we use `AppointmentsInSlot` with the `StartDate` and `EndDate`, but for getting Appointments in `OnKeyPress`, we use `StartDate` and `StartDate.AddDays(1)`, which is different and could potentially fetch in more appointments than are rendered.
So you could be hitting a key to move through appointments that haven't been rendered (I believe this is the only issue with the current code and not a major item).

What do you think?

Radzen.Blazor/Rendering/DayView.razor
paulo-rico: Found them! Didn't know they were now hidden (Excluded from Project).

Radzen.Blazor/Rendering/DayView.razor
akorchev: Indeed ideally the code that renders the appointments and the keyboard navigation should be the same.

Radzen.Blazor/Rendering/DayView.razor
akorchev: I still prefer avoiding inline styles unless absolutely necessary. Alternatively you can just not apply the cursor styling.

paulo-rico: I'll remove it

Radzen.Blazor/SchedulerDaySelectEventArgs.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+using Radzen.Blazor;
+
+namespace Radzen
+{
+    /// <summary>
+    /// Supplies information about a <see cref="RadzenScheduler{TItem}.MonthSelect" /> event that is being raised.
```
akorchev, line 8: Refers to MonthSelect.

Radzen.Blazor/RadzenScheduler.razor.cs
akorchev: The code still seems to refer to MonthSelect.

Radzen.Blazor/Rendering/MonthView.razor
```diff
@@ -125,6 +125,11 @@
         await Scheduler.SelectSlot(date, date.AddDays(1), AppointmentsInSlot(date, date.AddDays(1)));
     }
 
+    async Task OnDayClick(DateTime day)
+    {
+        await Scheduler.SelectDay(day, AppointmentsInSlot(day.Date, day.Date.AddHours(23).AddMinutes(59).AddSeconds(59)));
```
akorchev, line 130: This still feels wrong. Why are you doing it like this?

paulo-rico: You're right. It's should be as simple as

`await Scheduler.SelectDay(day, AppointmentsInSlot(day, day.AddDays(1)));`

Is this better?

PR #1911, opened 2025-01-15
Add is valid decimal value validation

ivan-rosales-rieloff:
Add is valid decimal value validation on ConvertToDecimal to avoid errors when decimal format is invalid, i.e., double comma, empty strings


enchev:
I've reverted this since one of the Numeric test failed.

PR #1912, opened 2025-01-16
Add empty string support, culture based cast

ivan-rosales-rieloff:
Add empty string support and culture base type conversion, update test suite to use culture and validate string to number conversion and empty string to zero value conversion.

Use component culture parameter to cast value to decimal type.

### DIFF
--- a/Radzen.Blazor.Tests/NumericTests.cs
+++ b/Radzen.Blazor.Tests/NumericTests.cs
@@ -463,10 +463,11 @@ namespace Radzen.Blazor.Tests
             ctx.JSInterop.SetupModule("_content/Radzen.Blazor/Radzen.Blazor.js");
 
             var value = new Dollars(11m);
-            Dollars? ConvertFunc(string s) => decimal.TryParse(s, out var val) ? new Dollars(val) : null;
+            Dollars? ConvertFunc(string s) => decimal.TryParse(s, System.Globalization.CultureInfo.InvariantCulture, out var val) ? new Dollars(val) : null;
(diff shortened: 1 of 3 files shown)


Radzen.Blazor.Tests/NumericTests.cs
```diff
@@ -489,14 +490,91 @@ public void Numeric_Supports_TypeConverter()
 
             var component = ctx.RenderComponent<RadzenNumeric<Dollars>>(
                 ComponentParameter.CreateParameter(nameof(RadzenNumeric<Dollars>.Format), format),
-                ComponentParameter.CreateParameter(nameof(RadzenNumeric<Dollars>.Value), valueToTest)
+                ComponentParameter.CreateParameter(nameof(RadzenNumeric<Dollars>.Value), valueToTest),
+                ComponentParameter.CreateParameter(nameof(RadzenNumeric<Dollars>.Culture), System.Globalization.CultureInfo.InvariantCulture)
             );
```
enchev, line 495: Why you need to specify Culture parameter here? This property is not mandatory, please make sure that everything works as expected without specified Culture.

ivan-rosales-rieloff: I add a test without culture, in my machine the culture is es-CL and the test fail if I don't specify culture, to ensure the string conversion I use

```
Assert.Contains($" value=\"{valueToTest.ToString(format, System.Globalization.CultureInfo.CurrentCulture)}\"", component.Markup);
```
to validate the render result, also add a new test using a specific culture to test both cases with and without culture set

PR #1915, opened 2025-01-19
Text shifting fix for RadzenDropDown

dmitrienko-andrey:
Used the **rz-helper-hidden** class instead of **rz-helper-hidden-accessible** for the dropdown input wrapper. The fix needs to be tested for all possible uses of this control.

### DIFF
--- a/Radzen.Blazor/RadzenDropDown.razor
+++ b/Radzen.Blazor/RadzenDropDown.razor
@@ -10,7 +10,7 @@
 {
     <div @ref="@Element" @attributes="Attributes" class="@GetCssClass()" onmousedown="Radzen.activeElement = null" @onclick="@(args => OpenPopup("ArrowDown", false, true))" @onclick:preventDefault @onclick:stopPropagation style="@Style" tabindex="@(Disabled ? "-1" : $"{TabIndex}")"
          @onkeydown="@((args) => OnKeyPress(args))" id="@GetId()"  @onfocus="@((args) => OnFocus(args))">
-        <div class="rz-helper-hidden">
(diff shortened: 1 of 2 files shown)


yordanov:
Hi @dmitrienko-andrey,

What is it that you want to fix with this change? A screenshot before/after could help. 

There is a fundamental difference on how the two classes render, `.rz-helper-hidden` uses display: none;

```CSS
.rz-helper-hidden-accessible {
  opacity: 0;
  height: 0;
    
  input {
    height: 0;
    padding: 0;
    margin: 0;
    border: 0;
  }
}

.rz-helper-hidden {
  display: none;
}
```

dmitrienko-andrey:
Hi @yordanov!
Can you tell me what specific functionality will be lost or stop working if this fix is applied? For my current project I am using this crutch and so far everything is working as expected:
```
.rz-helper-hidden-accessible {
    display: none;
}
```

dmitrienko-andrey:
One more option to fix it:
```
.rz-helper-hidden-accessible {
  opacity: 0;
  height: 0;
    
  input {
    height: 0;
    padding: 0;
    margin: 0;
    border: 0;
    width: 0;
  }
}
```

yordanov:
One of the differences is that the input's `aria-label=""` won't be visible to screen readers. Here are the other differences:

image

[Source: stackoverflow](https://stackoverflow.com/questions/272360/does-opacity0-have-exactly-the-same-effect-as-visibilityhidden/273076#273076)

yordanov:
@dmitrienko-andrey, thanks! That should do the job.

PR #1924, opened 2025-01-24
Strong type Value on DataBoundFormComponent<T>

pianomanjh:
This updates the Value property of `DataBoundFormComponent<T>` to be of type `T`, for more canonical use, consistency, and ease of extension.
We have several custom controls that benefit from inheriting from this component, to better fit in the Radzen ecosystem.  Needing to cast T when binding in these scenarios causes various difficulties, and is unnecessary as much as I can tell.

This should not result in any breaking changes, as users who are already casting this value to T will simply no longer need to.

Please let me know if I'm missing something as to the need for `Value` to be `object` vs `T`.

### DIFF
--- a/Radzen.Blazor.Tests/DropDownTests.cs
+++ b/Radzen.Blazor.Tests/DropDownTests.cs
@@ -123,7 +123,7 @@ namespace Radzen.Blazor.Tests
 
             List<DataItem> boundCollection = [new() { Text = "Item 2" }];
 
-            var component = DropDown<string>(ctx, parameters => {
+            var component = DropDown<List<DataItem>>(ctx, parameters => {
                 parameters.Add(p => p.ItemComparer, new DataItemComparer());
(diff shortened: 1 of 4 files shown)


akorchev:
This **is** a breaking change as some components support multiple selection and single selection. The value could be either T or IEnumerable<T> depending on selection mode. I don't think we can accept this pull request.

pianomanjh:
Can you share an example with me?  I've tested this across the various multi/single select component demos.  (ListBox/DropDown/DropDownDataGrid/etc)

enchev:
DataGrid CheckBoxList filtering by first column
image

pianomanjh:
thanks for the example, not sure how I missed this.  This particular situation is probably solvable with some code in `RadzenDataGridHeaderCell`, but likely not worth the risk as I understand it.  
For my purposes I imagine I believe I can use `FormComponent<T>` instead of `DataBoundFormComponent<T>`, closing.

PR #1932, opened 2025-01-31
Added a parameter called "Size" to the "RadzenGravatar" component. 

jamesalainheffer:
This will allow for users to adjust the size of the gravatar image coming in, so that they are not stuck with just a 32px image.

### DIFF
--- a/Radzen.Blazor/RadzenGravatar.razor.cs
+++ b/Radzen.Blazor/RadzenGravatar.razor.cs
@@ -29,12 +29,11 @@ namespace Radzen.Blazor
         public string AlternateText { get; set; } = "gravatar";
 
         /// <summary>
-        /// Gets or sets the size. Defaulted to 32 (pixels).
-        /// 
+        /// Gets or sets the size. Defaulted to 36 (pixels). 
         /// <value>The size.</value>
         /// </summary>
         [Parameter]
-        public int Size { get; set; } = 32;
+        public int Size { get; set; } = 36;
 
         /// <summary>
         /// Gets gravatar URL.
@@ -46,9 +45,8 @@ namespace Radzen.Blazor
                 var md5Email = MD5.Calculate(System.Text.Encoding.ASCII.GetBytes(Email != null ? Email : ""));
 
                 var style = "retro";
-                var width = Size;
 
-                return $"https://secure.gravatar.com/avatar/{md5Email}?d={style}&s={width}";
+                return $"https://secure.gravatar.com/avatar/{md5Email}?d={style}&s={Size}";
             }
         }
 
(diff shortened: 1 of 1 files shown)


Radzen.Blazor/RadzenGravatar.razor.cs
```diff
@@ -28,6 +28,14 @@ public partial class RadzenGravatar : RadzenComponent
         [Parameter]
         public string AlternateText { get; set; } = "gravatar";
 
+        /// <summary>
+        /// Gets or sets the size. Defaulted to 32 (pixels).
+        /// 
+        /// <value>The size.</value>
+        /// </summary>
+        [Parameter]
+        public int Size { get; set; } = 32;
```
akorchev, line 37: The current default value is `36`. Please use it instead of `32`.

Radzen.Blazor/RadzenGravatar.razor.cs
```diff
@@ -38,7 +46,7 @@ protected string Url
                 var md5Email = MD5.Calculate(System.Text.Encoding.ASCII.GetBytes(Email != null ? Email : ""));
 
                 var style = "retro";
-                var width = "36";
+                var width = Size;
```
akorchev, line 49: You can delete the `width` variable and use `Size` directly:
```
return $"https://secure.gravatar.com/avatar/{md5Email}?d={style}&s={Size}";
```

akorchev (changes requested):

Radzen.Blazor/RadzenGravatar.razor.cs
jamesalainheffer: Amazing! I didn't want to change too much of what was already there and working. I just assumed swapping out the hard coded value for that of a parametrized variable would have achieved the same result.

But, that's cool! Hopefully we will be able to have bigger, non-blurry profile images now 🚀

Radzen.Blazor/RadzenGravatar.razor.cs
jamesalainheffer: Oops 😅

Radzen.Blazor/RadzenGravatar.razor.cs
```diff
@@ -28,6 +28,13 @@ public partial class RadzenGravatar : RadzenComponent
         [Parameter]
         public string AlternateText { get; set; } = "gravatar";
 
+        /// <summary>
+        /// Gets or sets the size. Defaulted to 36 (pixels). 
+        /// <value>The size.</value>
```
akorchev, line 33: The `<value>` tag should be outside of `<summary>`.

akorchev (changes requested):

Radzen.Blazor/RadzenGravatar.razor.cs
jamesalainheffer: Oops, my bad 🤣 has been fixed, apologies

PR #1933, opened 2025-01-31
RadzenDatePicker floating label and RadzenFormField support

### LINKED ISSUE #1929: Support floating label for RadzenDatePicker

It seems like RadzenDatePicker is currently not intented to be used inside a RadzenFormField I could not find any examples for this in the docs. Thus the floating label is not working if you use it like this anyway.

It would be great if RadzenDatePicker would have full support for use inside RadzenFormField because date fields are an essential part for a lot of forms and should behave consistently.

From my inspection it is basically 3 things that prevent this from working currently.
1. `rz-state-empty` css class is not set for empty date picker values
2. Focused RadzenLabel styles are only considered for intented form fields (RadzenTextBox, RadzenTextArea etc.)
3. Inline variant of the date picker will obviously not work in this concept

Thoughts about addressing those three issues:

1. This could be a simple one. Add the `rz-state-empty` css class based on the `HasValue` property.
2. As RadzenFormField has a special `rz-state-focused` css class set when focused, the focused label styles could be defined based on this.
   https://github.com/radzenhq/radzen-blazor/blob/716d1a6cba131b2bab82b0612cac9a1e39a652a6/Radzen.Blazor/themes/components/blazor/_form-field.scss#L346-L388
   Maybe add `.rz-form-field.rz-state-focused .rz-form-field-label` respectively `.rz-form-field:not(.rz-variant-outlined).rz-state-focused .rz-form-field-label` to the selectors?
   
   This could possibly also enable us to create custom form field components that behave as expected.

3. This could be the toughest one. Only solution that comes to my mind is to separate the components. Do a RadzenCalenderPicker or something like that as the inline variant. Make use of it in the date picker field popup....

- Fixed variant by a wrapping component that adds the `rz-state-empty` css class and adds the focused label styles. (Tough one as you have to pass a lot of parameters through the wrapper, at least the ones that are used).
- Live with the always floating label for now ;)

**Additional context**

See it in action.

Image

rklfss:
This PR enables the RadzenDatePicker to be used inside a RadzenFormField including floating label support.

- Therefore the CSS class `rz-state-empty` is added to the component if there is no value.
- Stylesheets are adjusted so that the focused state of RadzenFormField is considered by the floating label.

resolves #1929

### DIFF
--- a/Radzen.Blazor.Tests/DatePickerTests.cs
+++ b/Radzen.Blazor.Tests/DatePickerTests.cs
@@ -175,6 +175,34 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void DatePicker_Renders_EmptyCssClass_WhenValueIsEmpty()
+        {
+            using var ctx = new TestContext();
+            ctx.JSInterop.Mode = JSRuntimeMode.Loose;
+            ctx.JSInterop.SetupModule("_content/Radzen.Blazor/Radzen.Blazor.js");
+
+            var component = ctx.RenderComponent<RadzenDatePicker<DateTime>>();
+
+            component.SetParametersAndRender(parameters => parameters.Add(p => p.Value, null));
+
+            Assert.Contains(@$"rz-state-empty", component.Markup);
+        }
+
+        [Fact]
+        public void DatePicker_DoesNotRender_EmptyCssClass_WhenValueIsNotEmpty()
+        {
+            using var ctx = new TestContext();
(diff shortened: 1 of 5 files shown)


rklfss:
@enchev Got the hint ;) But no idea how to adjust the demo stylesheets. Guess I'm not able to do this myself.

enchev:
You can use some of the non premium themes (like Material) to test - we will update premium themes before the release.

PR #1945, opened 2025-02-05
added parameter to set "AllowSelectAll" in de radzenPicklist

nielsNocore:
added parameter to set "AllowSelectAll" in de radzenPicklist

### DIFF
--- a/Radzen.Blazor/RadzenPickList.razor.cs
+++ b/Radzen.Blazor/RadzenPickList.razor.cs
@@ -43,6 +43,13 @@ namespace Radzen.Blazor
         public bool Multiple { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether selecting all items is allowed.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


Radzen.Blazor/RadzenPickList.razor.cs
```diff
@@ -42,6 +42,13 @@ public partial class RadzenPickList<TItem> : RadzenComponent
         [Parameter]
         public bool Multiple { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether selecting all items is allowed.
+        /// </summary>
+        /// <value><c>true</c> if selecting all items is allowed; otherwise, <c>false</c>.</value>
+        [Parameter]
+        public bool AllowSelectAll { get; set; }
```
enchev, line 50: Maybe the default value should be true since the default value of DropDownBase AllowSelectAll is true as well

nielsNocore: true, updated the pr

PR #1958, opened 2025-02-11
Form field: styling fixes and tweaks

Cosmatevs:
- ~if a form field is a container element that has its own variant, this variant no longer affects the form field style, e.g. breaking placement of labels inside empty text areas:~ Moved to https://github.com/radzenhq/radzen-blazor/pull/2002
misplaced textarea label
- icons inside drop downs, drop down data grids, and date pickers behave more consistently for flat and filled variants:
  - if the field is empty, they're centered vertically towards the whole field,
  - otherwise, they are centered verticaly towards the value, effectively going down.
Current setup:
Current form field icons
New setup:
More consistent form field icons

I hope the new setup will make it easier to implement *clear* and other action buttons in other field components.

### DIFF
--- a/Radzen.Blazor.Tests/DatePickerTests.cs
+++ b/Radzen.Blazor.Tests/DatePickerTests.cs
@@ -155,7 +155,7 @@ namespace Radzen.Blazor.Tests
                 parameters.Add<bool>(p => p.AllowClear, true);
             });
 
-            Assert.Contains(@$"<i class=""notranslate rz-dropdown-clear-icon rzi rzi-times""", component.Markup);
+            Assert.Contains(@$"<i class=""notranslate rz-clear-field-icon rzi rzi-times""", component.Markup);
         }
 
         [Fact]
--- a/Radzen.Blazor/RadzenDatePicker.razor
+++ b/Radzen.Blazor/RadzenDatePicker.razor
@@ -15,7 +15,7 @@
             <div class="rz-field-icon-wrapper">
                 @if (AllowClear && !Disabled && HasValue)
                 {
-                    <i class="notranslate rz-clear-icon rzi rzi-times" @onclick="@Clear" @onclick:stopPropagation="true"></i>
+                    <i class="notranslate rz-clear-field-icon rzi rzi-times" @onclick="@Clear" @onclick:stopPropagation="true"></i>
                 }
                 @if (ShowButton)
                 {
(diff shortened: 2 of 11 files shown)


yordanov:
Hi @Cosmatevs,

Thanks for fixing these glitches. Maybe it's me, but I cannot find any changes addressing the first bullet point "if a form field is a container element that has its own variant, this variant no longer affects the form field style"?

We'll need more time to review the rest. At first glance,  some of the styles could be considered breaking changes and might require bumping the major version of the components.

Cosmatevs:
@yordanov The first issue was addressed by changing css selectors from e.g. `.rz-variant-filled` to `.rz-form-field.rz-variant-filled`.
On GitHub it looks like I've rewritten the whole formfield.scss file 😐 That's because I made the whitespace consistent with other scss files. It doesn't look so scary once you switch the *hide whitespace* setting.

yordanov:
Hi @Cosmatevs,

Thanks for isolating the "css selectors for variants" fix in a separate PR. I am closing this PR as it is too complex - CSS classes renamed, components' layouts changed etc. Let's try to incrementally resolve these.

Some feedback:
- The jumping trigger icons do not feel right. From a UX perspective it is too much happening on a single interaction. Also, how would a multi-select with chips look like? Here is how it looks now:
image

- I like the clear-icon utility variables! Let's try to add these without renaming the elements' CSS classes.

Cosmatevs:
@yordanov Understandable. However, I need to clarify a few things. I should have done this earlier, as the changes weren't just to improve the appearance of the form field component.
1. The only CSS class name change was from `rz-dropdown-clear-icon` to `rz-clear-field-icon`. I decided to do this because this class isn't only used in dropdowns.
2. There's one new class: `rz-field-icon-wrapper`. It's intended to:
    - group field icons,
    - make them behave similarly across components,
    - make it easier to implement more than one icon without having to do heavy styling to make it look right.
  In the current setup, you can see the problem in the DatePicker component, which has two icons. They behave differently in a form field, and the clear icon doesn't respect the `ShowButton="false"` setting, leaving a large gap after the clear button:
  Clearable DatePicker without a trigger button
  My solution makes the icons automatically adjust to the end of a block. And it also makes them behave the same way in a form field.
  Styling for appropriate field padding depending on the buttons shown was still needed, though.
3. The classes `rz-dropdown-trigger` and `rz-multiselect-trigger` are no longer used as they were wrappers for `rz-dropdown-trigger-icon` and `rz-dropdown-multiselect-icon`.
4. Multi-select fields with chips work visually almost the same as before. The only difference is that the height of the label is taken into account when centering the icons in the flat or filled variants. *Jumping* only occurs when a field is empty and you make it focused or active. Just like other components with icons.

PR #1977, opened 2025-02-18
Fix OnSideClose being called on open

### LINKED ISSUE #1976: Dialog OnSideClose gets called when opening and closing
**Describe the bug**
Dialog OnSideClose gets called when opening and closing.

**To Reproduce**
Add a handler to OnSideClose for a dialog, and it will be called whenever the side dialog opens and closes. See these lines of code:
https://github.com/radzenhq/radzen-blazor/blob/a2591f7f7725543493bbea92afe7ed1cfe08069b/Radzen.Blazor/DialogService.cs#L171
https://github.com/radzenhq/radzen-blazor/blob/a2591f7f7725543493bbea92afe7ed1cfe08069b/Radzen.Blazor/DialogService.cs#L193

**Expected behavior**
OnSideClose should only be invoked when the side dialog is actually closed.

**Additional context**
I'm guessing this code is here to make sure any existing side dialogs are closed first. But this should be done in a way that doesn't also call the OnSideClose event. Otherwise, there is no way to actually be notified correctly.

sethcalebweeks:
- Move invocation of OnSideClose to inside the if block that checks if the task is complete.
- Add Console to example to demonstrate event handler

Closes #1976

### DIFF
--- a/Radzen.Blazor/DialogService.cs
+++ b/Radzen.Blazor/DialogService.cs
@@ -210,9 +210,9 @@ namespace Radzen
             if (_sideDialogTask?.Task.IsCompleted == false)
             {
                 _sideDialogTask.TrySetResult(result);
+                OnSideClose?.Invoke(result);
             }
 
-            OnSideClose?.Invoke(result);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


enchev:
This change was reverted since it broke closing of side dialogs opened with OpenSide() instead OpenSideAsync()

Horea-Marc:
This issue is still persisting in version 6.2.6

enchev:
Hey @Horea-Marc,

You’ve missed to read my post? This fix was reverted.

Horea-Marc:
So, #1976 is still an open issue?

enchev:
Yes, feel free to propose fix that will not break anything else.

PR #1980, opened 2025-02-19
Update RadzenDatePicker.razor.cs

AndradeAlbuquerque:
In some cultures, such as Portuguese (Portugal), the name of the day of the week appears in full, overlapping one another. This change forces the abbreviation of the name of the day of the week.

Before:
Screenshot 2025-02-19 123202

After:
Screenshot 2025-02-19 123048

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor.cs
+++ b/Radzen.Blazor/RadzenDatePicker.razor.cs
@@ -522,7 +522,7 @@ namespace Radzen.Blazor
             {
                 for (int current = (int)Culture.DateTimeFormat.FirstDayOfWeek, to = current + 7; current < to; current++)
                 {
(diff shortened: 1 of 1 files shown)


enchev:
This code crashed on several different cultures and the commit was reverted.

AndradeAlbuquerque:
> This code crashed on several different cultures and the commit was reverted.

I suspected this might happen. Works perfectly in Portuguese from Portugal (pt-PT). 
I will do more tests with other cultures to see what happens.

akorchev:
The code fails when the length of the string is less than 3 characters. Happens with `de-DE` culture for example. Also calling ToUpper() is unnecessary.

AndradeAlbuquerque:
I will correct the code and test for other cultures.

PR #1981, opened 2025-02-19
Scheduler Resource Grouping

paulo-rico:
Hi @akorchev 

Following on from my last PR, this is a revised edition.

This initial PR is just to highlight the method by which I have achieved the grouping, i.e. using recursion (we never know how many resources are present!!). 

The key files are `RadzenResourceDayView` and `ResourceDayView` (and their `WeekView` counterparts).

I am fully aware that the code can be integrated with existing code, i.e. I don't think we need to have separate `DropableViewBase` and `DropableResourceViewBase`. The reason for the change is that, because of recursion, there are individual renders based on `DropableViewBase`, so the two variables `draggedAppointment` and `dragStarted` need to be referenced from a "Parent". This will also work for non-resource views, so we could combine this (this is the reason for the Cascading Parameter of type `ScheduleViewBase` in the new resource views).

I also think that the `RadzenResourceDayView` could easily be integrated with `RadzenDayView`. If you look at the former, I have the component rendering `ResourceDayView` if there are any `Resources` present, and `DayView` if there aren't. We could keep this as one control (the existing `RadzenDayView`). 

I also copied the `DayView` rendering code for `ResourceDayView` with minor changes. This should be the same, with `ResourceDayView` using `DayView` to output the "Diary" portion instead of the copied code. Minor non-breaking changes would be required to DayView to achieve this.

Hope this is an improvement on the last PR and look forward to discussing.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/DialogService.cs
+++ b/Radzen.Blazor/DialogService.cs
@@ -210,9 +210,9 @@ namespace Radzen
             if (_sideDialogTask?.Task.IsCompleted == false)
             {
                 _sideDialogTask.TrySetResult(result);
-                OnSideClose?.Invoke(result);
             }
 
+            OnSideClose?.Invoke(result);
         }
 
         /// <summary>
--- a/RadzenBlazorDemos/Pages/DialogSide.razor
+++ b/RadzenBlazorDemos/Pages/DialogSide.razor
@@ -5,7 +5,7 @@
         <RadzenStack Orientation="Orientation.Horizontal" Gap="0.5rem" AlignItems="AlignItems.Center">
             <RadzenLabel Text="Position:" Component="Position" />
             <RadzenSelectBar @bind-Value="@position" TextProperty="Text" ValueProperty="Value" Name="Position"
-            Data="@(Enum.GetValues(typeof(DialogPosition)).Cast<DialogPosition>().Select(t => new { Text = $"{t}", Value = t }))" Size="ButtonSize.Small" />
+                            Data="@(Enum.GetValues(typeof(DialogPosition)).Cast<DialogPosition>().Select(t => new { Text = $"{t}", Value = t }))" Size="ButtonSize.Small" />
         </RadzenStack>
         <RadzenStack Orientation="Orientation.Horizontal" Gap="0.5rem" AlignItems="AlignItems.Center">
             <RadzenLabel Text="Show mask:" Component="Mask" />
@@ -17,28 +17,15 @@
         </RadzenStack>
     </RadzenStack>
     <RadzenButton Text="Dialog on Side" ButtonStyle="ButtonStyle.Secondary" Click="@OpenSideDialog" />
-    <EventConsole @ref=@console />
 </div>
 
 @code {
-    EventConsole console;
     DialogPosition position;
     bool closeDialogOnOverlayClick;
     bool showMask;
 
-    protected override async Task OnInitializedAsync()
-    {
-        await base.OnInitializedAsync();
-        DialogService.OnSideClose += OnSideClose;
-    }
-
     async Task OpenSideDialog()
     {
         await DialogService.OpenSideAsync<DialogSideContent>("Side Panel", options: new SideDialogOptions { CloseDialogOnOverlayClick = closeDialogOnOverlayClick, Position = position, ShowMask = showMask });
-    }
-
-    async void OnSideClose(dynamic _)
-    {
-        console.Log("Side dialog closed");
     }
 }


paulo-rico:
I should probable add, the whole thing is controlled by a linked list in the form of a `Dictionary<ISchedulerResource, ISchedulerResource>`.
The Key is the "Current" resource, and the Value is the next `Resource` in the list. If there is a "next resource", it will render "itself" again with a pointer to next resource. If next is `null`, it then renders the actual diary.

The way it filters is that each rendered component has a Property and Value that is used to generate a `FilterDescriptor`.  When the diary is eventually rendered, it gets all the FilterDescripters from the parents of each component.

akorchev:
Hi @paulo-rico,

I took a skim over the implementation. It looks a lot better from API point of view and I quite like it. I have some comments about the implementation though

- I prefer the view rendering code to not be duplicated. I saw a lot of similar code which is not ideal as it leads to maintenance nightmare along the road. In short we better add resource support to the existing DayView, WeeView etc. than reimplementing them.
- I dislike having properties such as MinViewWidth, MaxViewHeight and similar. All other Radzen.Blazor components use the more flexible "style" approach. This is why I think  those should come from the style of the view itself e.g. `<RadzenResourceWeekView style="min-width: 850px; max-height: 500px" />`
- I didn't quite understand the need for the linked list. The resource header could be rendered separately from the view and the view needs access to all resources anyway. Maybe I am missing something though.

paulo-rico:
Hi @akorchev 

_Note - for "diary", I refer to the actual rendering of the diary portion, not the navigation bar or group headers._

> I prefer the view rendering code to not be duplicated. I saw a lot of similar code which is not ideal as it leads to maintenance nightmare along the road. In short we better add resource support to the existing DayView, WeeView etc. than reimplementing them.

I did mention this in my initial comment. I was trying to keep the implementation of the resource stuff separate in order not to cloud how it all hung together. It was my intention to amalgamate the code to irradicate any duplication when the methodology had been "signed off".

> I didn't quite understand the need for the linked list. The resource header could be rendered separately from the view and the view needs access to all resources anyway. Maybe I am missing something though.

I can't see any other way. It calls for the classic recursion method. The linked list is just an alternative to passing the index of the next resource record for recursion. I can change it to that if prefer, but they are in effect the same thing.

> I dislike having properties such as MinViewWidth, MaxViewHeight and similar. All other Radzen.Blazor components use the more flexible "style" approach. This is why I think those should come from the style of the view itself e.g. <RadzenResourceWeekView style="min-width: 850px; max-height: 500px" />

I may leave this to discuss while the bones of the PR have been resolved. The width of the diary is currently independent of the resource component, i.e. if there is only one resource record with only two values, the diary displays fine, but if there are multiple resources with multiple values, it will shrink to fit on screen. When the width of the diary is set, the parents (headers [Resource Items]) all expand to accommodate this width. So the width property was put there for that. There's probably an alternative. I'll give it some thought.

Would be easier if I reversed this PR and put in the finished code without duplication before we progress any further?

akorchev:
> I can't see any other way. It calls for the classic recursion method. The linked list is just an alternative to passing the index of the next resource record for recursion. I can change it to that if prefer, but they are in effect the same thing.

There is a lot of recursion in the Radzen.Blazor components but never does it use a linked list. Please show me the code that requires the linked list. In most cases a simple foreach over a collection does the job. For example rendering the nested column headers of RadzenDataGrid doesn't need a linked list.

>  The width of the diary is currently independent of the resource component, i.e. if there is only one resource record with only two values, the diary displays fine,

What is a diary? Do you mean a view? Having such properties has a few limitations mainly related to responsiveness and overriding them with CSS. How does it look like without any size?

> Would be easier if I reversed this PR and put in the finished code without duplication before we progress any further?

The less modifications to the original the better for me to understand the whole PR. In its current state I see a lot of duplication of properties, events, XML API reference comments etc which confuses me.

paulo-rico:
Is this what you mean by nested columns [Composite Columns](https://blazor.radzen.com/datagrid-composite-columns?theme=material3)?

paulo-rico:
I discarded my changes and am currently putting together a concise version. Not sure if PR will reopen after a further commit. If not, I'll open a new one.

Regarding the linked list - The reason I need to do this and cannot use a `foreach` loop is that there are no relationships between the resources entered. With tree view and column gouping e.t.c., the parent / child relationship is defined by the actual markup, i.e. 
```
        <RadzenDataGridColumn Title="Title" Property="@nameof(Employee.TitleOfCourtesy)" Filterable=@allowCompositeDataCells Sortable=@allowCompositeDataCells Width="820px">
            <Columns>
                <RadzenDataGridColumn Property="@nameof(Employee.FirstName)" Title="FirstName" Width="480px" Filterable=@allowCompositeDataCells Sortable=@allowCompositeDataCells>
                    <Columns>
                        <RadzenDataGridColumn Property="@nameof(Employee.Title)" Title="Job Title" Width="200px" />
                        <RadzenDataGridColumn Property="@nameof(Employee.BirthDate)" Title="Birth Date" FormatString="{0:d}" Width="140px" />.....
 
```

The linked list just gives an easy way of moving through the resources until there are none left and the actual "diary" can be rendered.

akorchev:
Why don't you just traverse the resource collection then? It should be in the exact same order as the linked list.

paulo-rico:
Because there are a couple of properties on the resource that can be set to change that. There's a Visible property (thinking of changing to Active) that will include or ignore a resource in grouping and there's an Order property to change the group order.

akorchev:
You can filter and order the collection:

```
foreach (var resource in Resources.Where(r => r.Visible).OrderBy(r => r.Order))
{
}
```

paulo-rico:
But recursion won't be possible with that. Because it is just a list of resources with no relationship, the recursion needs to happen at the top level rather than from any child(ren). I'm not far with this update. It should become apparent when you see the code.

PR #1990, opened 2025-02-20
Update RadzenDatePicker.razor.cs for abbreviated day names

AndradeAlbuquerque:
The names of the days of the week are not abbreviated in some cultures such as Portuguese (Portugal) or Portuguese (Brazil). 
I propose again this change that I have already tested in several different cultures.

Before:
Screenshot 2025-02-19 123202

After:
Screenshot 2025-02-19 123048

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor.cs
+++ b/Radzen.Blazor/RadzenDatePicker.razor.cs
@@ -522,7 +522,8 @@ namespace Radzen.Blazor
             {
                 for (int current = (int)Culture.DateTimeFormat.FirstDayOfWeek, to = current + 7; current < to; current++)
                 {
-                    yield return Culture.DateTimeFormat.AbbreviatedDayNames[current % 7];
+                    string weekday = Culture.DateTimeFormat.AbbreviatedDayNames[current % 7];
+                    yield return weekday.Length <= 3 ? weekday : weekday[..3];
                 }
             }
         }


akorchev:
I am afraid we can't accept this pull request.

This is a .NET issue that applies to the pt-PT culture only: https://github.com/dotnet/runtime/issues/61088

For example using "pt-BR" or "pt" as culture seems to work fine.
```
<RadzenDatePicker Culture="@(new System.Globalization.CultureInfo("pt-BR"))" @bind-Value=@value />
```

image

If we accept that pull request it would introduce a breaking change for all cultures as it would strip abbreviated names to 3 chars always. You can try using this as a workaround:

AndradeAlbuquerque:
Yes, it is a dotnet issue. I understand your concerns. It could be done for 'pt-PT' only as a temporary fix.
Anyway, the workaround Setting 'pt' (instead of 'pt-PT') may not resolve this for many Portuguese (Portugal) apps where Culture pt-PT is in place.
I'll have a private version of Radzen Components until dotnet is not fixed. I will not make any other pull request regarding this.
Thanks.

AndradeAlbuquerque:
> I am afraid we can't accept this pull request.
> 
> This is a .NET issue that applies to the pt-PT culture only: [dotnet/runtime#61088](https://github.com/dotnet/runtime/issues/61088)
> 
> For example using "pt-BR" or "pt" as culture seems to work fine.
> 
> ```
> <RadzenDatePicker Culture="@(new System.Globalization.CultureInfo("pt-BR"))" @bind-Value=@value />
> ```
> 
> image
> If we accept that pull request it would introduce a breaking change for all cultures as it would strip abbreviated names to 3 chars always. You can try using this as a workaround:

Thanks, but "pt" is not an option to consider in existing apps where pt-PT must be (or is) in place. DotNet is broken. We will work on a private copy of these components.

akorchev:
You can also use CSS to clip the text:

```
    .rz-calendar-view th span {
        width: 30px;
        display: inline-block;
        overflow: hidden;
    }
```

AndradeAlbuquerque:
> You can also use CSS to clip the text:
> 
> ```
>     .rz-calendar-view th span {
>         width: 30px;
>         display: inline-block;
>         overflow: hidden;
>     }
> ```

Thanks for the tip.

PR #1994, opened 2025-02-21
RadzenScheduler - Resource Grouping

paulo-rico:
Hi all

Easiest way to follow through this is to begin at `RadzenDayView.razor` (week view and multi day work the same).

If there are no resources or you have elected to not show grouping (`ShowResourceGrouping`) it will output the same as it did before. The slight difference there is that, along with `CascadingParameter` for `Scheduler`, it also passes down itself (`SchedulerViewBase`) for the `DropableViewBase`. It uses this as a root for holding both `DragStarted` and `DraggedAppointment` variables. They had to be moved to a "parent" component in order to move appointments between multiple "diaries" (the bit at the bottom of each header that shows the slots and appointments).

The recursion loop then begins until there is no "child" resource, and then it renders the "diary" portion. So the rendering order would be something like this, where the integer is the resource and the decimal is the resource data (assuming two data items per resource)

1.1
2.1
3.1
Diary
1.1
2.1
3.2
Diary
1.1
2.2
3.1
Diary
e.t.c.

At each level, we have access to `Property` and `Value` so we can traverse back up the tree to build a `FilterDescription[]` in order to pass these `filteredAppointments` on to the "diary".

Although there are a great deal of file changes, the majority of these are to accommodate `ResourceFilterList` in the Scheduler events (which is defaulted to avoid breaking change). This is an `IEnumerable<string Property, string Value>` to pass back to identify which resource(s) are relavent to the "Click" or "Move", e.t.c.

Hope this is moving in the right direction.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/RadzenStackedBarSeries.razor
+++ b/Radzen.Blazor/RadzenStackedBarSeries.razor
@@ -31,12 +31,22 @@
                 var x = GetBarRight(data, barIndex, category, barSeries);
                 var itemValue = Value(data);
                 var radius = Chart.BarOptions.Radius;
-
                 var x0 = GetBarLeft(data, barIndex, category, barSeries);
+
+                if (itemValue < 0)
+                {
+                    (x, x0) = (x0, x);
+                }
 
                 if (radius > 0)
                 {
-                    var items = barSeries.SelectMany(series => series.ItemsForCategory(category(data))).OfType<TItem>().ToList();
+                    var items = barSeries.SelectMany(series => series.ItemsForCategory(category(data))).OfType<TItem>()
+                        .Where(item =>
+                        {
+                            var value = Value(item);
+                            return itemValue >= 0 ? value >= 0 : value < 0;
+                        }).ToList();
+
                     var index = items.IndexOf(data);
                     var width = Math.Abs(x0 - x);
 
@@ -56,6 +66,7 @@
                 {
                     path = $"M {x0.ToInvariantString()} {y.ToInvariantString()} L {(x+radius).ToInvariantString()} {y.ToInvariantString()} A {r} {r} 0 0 0 {x.ToInvariantString()} {(y+radius).ToInvariantString()} L {x.ToInvariantString()} {(y+height-radius).ToInvariantString()} A {r} {r} 0 0 0 {(x+radius).ToInvariantString()} {(y + height).ToInvariantString()} L {x0.ToInvariantString()} {(y+height).ToInvariantString()} Z";
                 }
+
                 var fill = PickColor(Items.IndexOf(data), Fills, Fill, FillRange, itemValue);
                 var stroke = PickColor(Items.IndexOf(data), Strokes, Stroke, StrokeRange, itemValue);
 
--- a/Radzen.Blazor/RadzenStackedBarSeries.razor.cs
+++ b/Radzen.Blazor/RadzenStackedBarSeries.razor.cs
@@ -84,11 +84,15 @@ namespace Radzen.Blazor
             {
                 var stackedBarSeries = BarSeries.Cast<IChartStackedBarSeries>();
                 var count = stackedBarSeries.Max(series => series.Count);
-                var sums = Enumerable.Range(0, count).Select(i => stackedBarSeries.Sum(series => series.ValueAt(i)));
-                var max = sums.Max();
-                var min = Items.Min(Value);
+                var categories = Enumerable.Range(0, count);
 
-                scale.Input.MergeWidth(new ScaleRange { Start = min, End = max });
+                var positiveSums = categories.Select(i => stackedBarSeries.Sum(series => Math.Max(0, series.ValueAt(i))));
+                var negativeSums = categories.Select(i => stackedBarSeries.Sum(series => Math.Min(0, series.ValueAt(i))));
+
+                var maxPositive = positiveSums.Max();
+                var minNegative = negativeSums.Min();
+
+                scale.Input.MergeWidth(new ScaleRange { Start = minNegative, End = maxPositive });
             }
 
             return scale;
@@ -162,29 +166,47 @@ namespace Radzen.Blazor
             return category(item) - BarHeight / 2;
         }
 
-        private static double Sum(int barIndex, IEnumerable<IChartStackedBarSeries> stackedBarSeries, double category)
+        private static (double positiveSum, double negativeSum) Sum(int barIndex, IEnumerable<IChartStackedBarSeries> stackedBarSeries, double category)
         {
-            return stackedBarSeries.Take(barIndex).SelectMany(series => series.ValuesForCategory(category)).DefaultIfEmpty(0).Sum();
+            var values = stackedBarSeries.Take(barIndex).SelectMany(series => series.ValuesForCategory(category)).DefaultIfEmpty(0).ToList();
+            var positiveSum = values.Where(v => v >= 0).DefaultIfEmpty(0).Sum();
+            var negativeSum = values.Where(v => v < 0).DefaultIfEmpty(0).Sum();
+            return (positiveSum, negativeSum);
         }
-
         private double GetBarRight(TItem item, int barIndex, Func<TItem, double> category, IEnumerable<IChartStackedBarSeries> stackedBarSeries)
         {
-            var count = stackedBarSeries.Max(series => series.Count);
+            var value = Value(item);
+            var (positiveSum, negativeSum) = Sum(barIndex, stackedBarSeries, category(item));
 
-            var sum = Sum(barIndex, stackedBarSeries, category(item));
-
-            var y = Chart.CategoryScale.Scale(Value(item) + sum);
-
-            return y;
+            if (value >= 0)
+            {
+                var y = Chart.CategoryScale.Scale(value + positiveSum);
+                return y;
+            }
+            else
+            {
+                var y = Chart.CategoryScale.Scale(negativeSum);
+                return y;
+            }
         }
 
         private double GetBarLeft(TItem item, int barIndex, Func<TItem, double> category, IEnumerable<IChartStackedBarSeries> stackedBarSeries)
         {
-            var ticks = Chart.CategoryScale.Ticks(Chart.ValueAxis.TickDistance);
+            var value = Value(item);
+            var (positiveSum, negativeSum) = Sum(barIndex, stackedBarSeries, category(item));
 
-            var sum = Sum(barIndex, stackedBarSeries, category(item));
-
-            return Chart.CategoryScale.Scale(Math.Max(ticks.Start, sum));
+            if (value >= 0)
+            {
+                var ticks = Chart.CategoryScale.Ticks(Chart.ValueAxis.TickDistance);
+                var sum = Math.Max(ticks.Start, positiveSum);
+                return Chart.CategoryScale.Scale(sum);
+            }
+            else
+            {
+                var ticks = Chart.CategoryScale.Ticks(Chart.ValueAxis.TickDistance);
+                var sum = Math.Max(ticks.Start, negativeSum + value);
+                return Chart.CategoryScale.Scale(sum);
+            }
         }
 
         int IChartBarSeries.Count
--- a/Radzen.Blazor/RadzenStackedColumnSeries.razor
+++ b/Radzen.Blazor/RadzenStackedColumnSeries.razor
@@ -29,11 +29,21 @@
                 var radius = Chart.ColumnOptions.Radius;
                 var y0 = GetColumnBottom(data, columnIndex, category, stackedColumnSeries);
 
+                if (itemValue < 0)
+                {
+                    (y, y0) = (y0, y);
+                }
+
                 if (radius > 0)
                 {
                     var height = Math.Abs(y0 - y);
 
-                    var items = stackedColumnSeries.SelectMany(series => series.ItemsForCategory(category(data))).OfType<TItem>().ToList();
+                    var items = stackedColumnSeries.SelectMany(series => series.ItemsForCategory(category(data))).OfType<TItem>()
+                        .Where(item =>
+                        {
+                            var value = Value(item);
+                            return itemValue >= 0 ? value >= 0 : value < 0;
+                        }).ToList();
 
                     var index = items.IndexOf(data);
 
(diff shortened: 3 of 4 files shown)


Radzen.Blazor/RadzenDayView.razor
```diff
@@ -7,13 +7,54 @@
...
+        var resources = Scheduler.ActiveResourceList;
 
-        return
-    @<CascadingValue Value=@Scheduler>
-        <DayView StartDate=@StartDate EndDate=@EndDate StartTime=@StartTime EndTime=@EndTime Appointments=@appointments
-            TimeFormat=@TimeFormat MinutesPerSlot=@MinutesPerSlot
-            AppointmentMove=OnAppointmentMove />
-    </CascadingValue>
-    ;
+        if (resources.Count > 0 && ShowResourceGrouping)
```
akorchev, line 12: Is there any use case for using resources and having ShowResourceGrouping as `false`? I don't think of any.

Radzen.Blazor/RadzenDayView.razor
```diff
@@ -7,13 +7,54 @@
...
+                else
+                {
+                    resourceDictionary[resource] = null;
+                }
+                next++;
+            }
+            var currentResource = resourceDictionary.FirstOrDefault().Key;
+
+            return
+                @<div class="rz-resource-scheduler">
```
akorchev, line 33: Why is RadzenDayView rendering the resource header? I think RadzeResourceDayView should do that and RadzenDayView should stay mostly untouched. The same goes for the other views.

For example I think RadzenResourceDayView should:

1. Render the resource headers
2. Render RadzenDayView for every resource.

akorchev (changes requested):

akorchev:
Hi @paulo-rico,

I think there may have been a misunderstanding judging by your latest commits. I still think we should have new Resource based views - RadzenResourceDayView, RadzenResourceWeekView etc. Those should render the resource headers and then render the children views passing them appointments filtered by data.

Radzen.Blazor/themes/components/blazor/_scheduler.scss
```diff
@@ -798,3 +798,67 @@ $scheduler-highlight-background-color: rgba(255,220,40,.2) !default;
...
+  border-block-start: 1px;
+}
+
+.rz-resource-scheduler-group-header-title {
+  background-color: var(--rz-scheduler-header-background-color);
+  height: 1.5rem;
+  font-weight: 600;
+}
+
+.rz-resource-scheduler-diary {
```
akorchev, line 846: Still don't know what "diary" is. We call them "views"?

RadzenBlazorDemos/Pages/SchedulerResourceConfig.razor
```diff
@@ -0,0 +1,207 @@
...
+
+    async Task OnAppointmentMove(SchedulerAppointmentMoveEventArgs args)
+    {
+        var draggedAppointment = appointments.FirstOrDefault(x => x == args.Appointment.Data);
+
+        if (draggedAppointment != null)
+        {
+            draggedAppointment.Start = draggedAppointment.Start + args.TimeSpan;
+            draggedAppointment.End = draggedAppointment.End + args.TimeSpan;
+            if(args.ResourceFilters!=null)
```
akorchev, line 198: This doesn't look as a good public API. I don't understand what it is doing.

Radzen.Blazor/SchedulerViewBase.cs
```diff
@@ -99,5 +106,15 @@ public async Task OnAppointmentMove(SchedulerAppointmentMoveEventArgs data)
         {
             await Scheduler.AppointmentMove.InvokeAsync(data);
         }
+        /// <summary>
+        /// Mark that a drag has started.
+        /// </summary>
+        /// <value>Has a drag started?</value>
+        public bool DragStarted { get; set; }
```
akorchev, line 113: Why are those needed?

Radzen.Blazor/Rendering/ResourceHeaderBase.razor
```diff
@@ -0,0 +1,96 @@
...
+                    filters.Add(current.Parent.Filter);
+                    current = current.Parent;
+                }
+            }
+
+            return Scheduler.GetAppointmentsInRange(StartDate, EndDate, filters.ToArray()).ToList();
+        }
+    }
+
+    public IList<(string Field, string Value)> resourceFilterList
```
akorchev, line 77: Properties must follow PascalCase naming convention.

Radzen.Blazor/RadzenDayView.razor
paulo-rico: This disappears with having separate Views. When it was one View, it gave the option to ignore any resources and render as normal.

Radzen.Blazor/RadzenDayView.razor
```diff
@@ -7,13 +7,54 @@
     public override RenderFragment Render()
     {
         var appointments = Scheduler.GetAppointmentsInRange(StartDate, EndDate).ToList();
+        var resources = Scheduler.ActiveResourceList;
```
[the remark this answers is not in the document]
paulo-rico: Gone. Related to combined Views.

Radzen.Blazor/ISchedulerResource.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Gets the property of the appointment data that corespondes to this resource
+        /// </summary>
+        /// <value>The property.</value>
+        string Property { get; set; }
+        /// <summary>
+        /// Gets the order of grouping
+        /// </summary>
+        /// <value>The order.</value>
+        int Order { get; set; }
```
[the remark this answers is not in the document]
paulo-rico: Gone

Radzen.Blazor/Common.cs
```diff
@@ -736,6 +736,11 @@ public class SchedulerAppointmentMoveEventArgs
         /// Gets or sets the time span.
         /// </summary>
         public TimeSpan TimeSpan { get; set; }
+
+        /// <summary>
+        /// List of resource filters.
+        /// </summary>
+        public IList<(string Property, string Value)> ResourceFilters { get; set; }
```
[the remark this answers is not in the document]
paulo-rico: Now `SchedulerResourceFilter`

Radzen.Blazor/themes/components/blazor/_scheduler.scss
paulo-rico: I've changed it to "view".

I used "diary" in order to differentiate between the view and the "diary" portion (for want of a better word). Originally, all we had was a `Scheduler` that contained a `View`. With the resource views, We have a `Scheduler` that contains a `View` that comprises a series of "Headers" and a series of "diaries".

Radzen.Blazor/SchedulerViewBase.cs
paulo-rico: This is to do with tracking the Started and Appointment variables when we're using multiple views in one. As it was, these variables were stored in the `DropableViewBase`, so you couldn't drag between the views. They would all have there own local variables. I've moved those tracking variables up the hierarchy so that all the views share these and we can drag-drop appointments between them.

paulo-rico:
> We should probably split replace Property with TextProperty and ValueProperty to make it more real-life:
<RadzenSchedulerResource Title="Location" ValueProperty="BuildingId" TextProperty="Name" Data="@([new Building { BuildingId = 1, Name = "Building One" }])">

I'm assuming this means we store the reference (Id) in the Appointment record as opposed to the Text.
I'm positive it is, but want to be 100% sure before making the changes :)

paulo-rico:
The nesting gave me some thoughts. Not to actually nest the resources, but to link them in some way.

My thoughts are along these lines. 

As it stands, we are outputting all headers for all "parent" headers, i.e. basically, nested `foreach` without any filtering (on the output of the resource headers).

But, as a real world example, this isn't always the best approach. Take a hotel chain, for example.

A chain has many hotels. A hotel has many rooms, these rooms have bookings. As it stands with the current grouping method, it will output "Hotel 1" and then output all the rooms in the database under this header. It will then render "Hotel 2" and all the rooms again. Sometimes, like this scenario, there needs to be a link to "filter" child group headers (rooms) for it to make sense in real life. Not output rooms for a hotel that don't exist.

Also in the above scenario, Appointment records in a database may not (or shouldn't, if following standard design practices) have the hotel reference as part of it. It would gather the hotel from the hotel reference of the room record, if need be. So this is making me think that not all resource records used for grouping will take part in the filter process of the appointment records, as it does now.

An initial idea is as follows. We have the following properties as part of each resource record -

- `Title` - the Title of the resource
- `ValueProperty` - The Id property of the `Resource` record. Example `RoomId` in a "Room" `Resource` record
- `TextProperty` - The text property of the `Resource` record, used primarily for the header title. Example `RoomTitle` in a "Room" `Resource` record
- `ResourceFilterProperty` - The property of the `Resource` record that links to the `ValueProperty` of the `Parent` resource. Example `HotelRef` in a "Room" `Resource`. If left blank or `null` all resource values will be output as it does now. Otherwise, filter the resource data records based on the parent resource and just output those headers (Rooms).
- `AppointmentFilterProperty` - The property of the `Appointment` record that holds the reference to this `Resource` data's record Id (`ValueProperty`). Example `RoomRef` in the `Appointment` record. If this is left blank or `null`, this `Resource` is not used to filter `Appointment` records.

Example Resource and Appointment components / data -

```
<RadzenSchedulerResource Title="Hotel" ValueProperty="HotelId" TextProperty="HotelName" 
        Data="@([new Hotel { HotelId = 1, HotelName = "The Dorchester" }, 
                 new Hotel { HotelId = 2, HotelName = "The Grand" }])" />
<RadzenSchedulerResource Title="Room" ValueProperty="RoomId" TextProperty="RoomName" 
        ResourceFilterProperty="HotelRef" 
        AppointmentFilterProperty = "RoomReference" 
        Data="@([new Room { RoomId = 1, RoomName = "101", HotelRef = 1 }, 
                         new Room { RoomId = 2, RoomName = "102", HotelRef = 1 }, 
                         new Room { RoomId = 3, RoomName = "103", HotelRef = 1 }, 
                         new Room { RoomId = 4, RoomName = "Presidential", HotelRef = 2 }, 
                         new Room { RoomId = 5, RoomName = "Penthouse", HotelRef = 2 }, 
                         new Room { RoomId = 6, RoomName = "Executive", HotelRef = 2 }])" />

```

```
    {
        new Appointment { Start = DateTime.Today, End = DateTime.Today, Text = "Mr Smith", RoomReference = 4 },
        new Appointment { Start = DateTime.Today, End = DateTime.Today, Text = "Mrs Jones", RoomReference = 5 },
        new Appointment { Start = DateTime.Today, End = DateTime.Today, Text = "Mrs Beech", RoomReference = 1 },
        new Appointment { Start = DateTime.Today, End = DateTime.Today, Text = "Mrs Chapman", RoomReference = 2 },
        new Appointment { Start = DateTime.Today, End = DateTime.Today, Text = "Mr Hunter", RoomReference = 6 }
    };

```
So, in the above example, it will output all the Hotels as header. When it comes to rendering the Rooms, it will see that `ResourceFilterProperty` as a value so will only output Room headers where the `ResourceFilterProperty` is equal to the Parent `ValueProperty`.

We're still not nesting. The "Parent" is the preceding `Resource.`

I'm sure this offers the flexibility required for this type of view.

What do you think? Would this work in the GUI IDE?

akorchev:
Found an example that seems flexible enough: https://demos.telerik.com/kendo-ui/scheduler/resources-grouping-hierarchical

1. Resources are independent from each other. Rooms and attendees don't have a relationship.
2. The "hierarchy" is determined by the "group" setting - it specifies the order of resources to group by.
3. Resources may repeat - as attendees do. Also an appointment can have multiple resources from the same type (multiple people can attend the same meeting).

paulo-rico:
> We may be overcomplicating things. Having to set three or four properties per resource seems to be too much.

> Resources are independent from each other. Rooms and attendees don't have a relationship.

Looking at the Kendo example supplied, it looks like this is what it does do. It must have a relationship so all resources are not output for all resources (for example, all second row resources for all first row resources).

Their example code -

```
    .Resources(resource =>
    {
        resource.Add(m => m.RoomID)
            .Title("Room")
            .Name("Rooms")
            .DataTextField("Text")
            .DataValueField("Value")
            .DataColorField("Color")
            .BindTo(new[] {
                    new { Text = "Meeting Room 101", Value = 1, Color = "#6eb3fa" },
                    new { Text = "Meeting Room 201", Value = 2, Color = "#f58a8a" }
            });
        resource.Add(m => m.Attendees)
            .Title("Attendees")
            .Name("Attendees")
            .Multiple(true)
            .DataTextField("Text")
            .DataValueField("Value")
            .DataColorField("Color")
            .DataParentValueField("Parent")
            .BindTo(new List<SchedulerResourceModel>() {
                    new SchedulerResourceModel(){ Text = "Alex", Color="red", Value = 1},
                    new SchedulerResourceModel(){ Text = "Bob", Color="green",  Value = 2, Parent = 1 } ,
                    new SchedulerResourceModel(){ Text = "Charlie",Color="yellow",  Value = 3, Parent = 2 }
            });
    })
```

This has a parent field, and, judging by the code, works in much the same way as my proposal whereby, if the value is null, outputs to all "parent" resources. In this example "Bob" is output for "Room 101", "Charlie" is output for "Room 102" and "Alex", whose parent value is null (not set), is output to both).

I have identified a few (some major) errors in my initial interpretation of what a "Resource "Scheduler" is, which may have been the source of confusion in some of our correspondence. The Kendo stuff has highlighted this and, I believe, put my mind on the right track.

As has been noted on previous occasions (other PR's), my ability to transfer code and system logic into plain English leaves a lot to be desired. If you think the example you provided is roughly the way to go, I think, from my point of view, the best way forward would be for me to re-engineer this update and send it on with detailed comments, then modify it from there. Even if not completely correct, I think it will be closer to what is required.

What do you think?

akorchev:
This all sounds perfect. I am also to blame for not explaining everything in the best possible way.

paulo-rico:
Looking at the bottom, I can reopen this PR. I'm assuming that will be available when there's something to commit / merge. I'll see when I actually commit something.

paulo-rico:
Final notes before draft commit - Had to add `Order` property back in for `ScheduleResource`. If a `ScheduleResource` is hidden and then shown again, will not be in the order of `ScheduleResource` components. See [Blazor Question: Order/Index of ChildComponents rendered via ChildContent/RenderFragment #54703](https://github.com/dotnet/aspnetcore/issues/54703)

paulo-rico:
Added `HeaderFormat` property and `@StartDate.ToString(HeaderFormat, Scheduler.Culture)` to `DayView` to bring it in line with `WeekView`. Mainly needed for a consistent interface when rendering view from header.

PR #1996, opened 2025-02-21
Fixed datagrid sort issues due to duplicate sortDescriptor lists. 

nielsNocore:
refactor datagrid sorting. to only have one ObservableCollection<SortDescriptor> instead of 2


enchev:
Thanks @nielsNocore however there are so many changes related to code formatting that one can hardly check what's changed really. Can you restore the origin formatting?

nielsNocore:
i will update the PR somewhere today.

nielsNocore:
updated PR, added some additional unit test
'DataGrid_Renders_LoadDataWithOrdering' is a test that will now fail on the main branch, in the load data the orderby property is not cleared if the ordering is cleared. ( this is the bug i tried to fix with this PR)

PR #2006, opened 2025-02-27
ADD:  RadzenDropZone - Allow Virtualization

maras420:
Hi all,

I'm newcomer here. I tried add simple support for enabling virtualization in RadzenDropZone


enchev:
Thanks @maras420! Please add also a demo demonstrating the feature.

PR #2008, opened 2025-02-28
Added “YearFormatter” property in RadzenDatePicker to allow custom formatting of Year

n4213-ndensan:
This enhancement aims to enable immediate response to changes in the era name without relying on .NET’s built-in implementation.

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor.cs
+++ b/Radzen.Blazor/RadzenDatePicker.razor.cs
@@ -285,7 +285,7 @@ namespace Radzen.Blazor
             YearTo = max.HasValue ? max.Value.Year : int.Parse(YearRange.Split(':').Last());
             months = Enumerable.Range(1, 12).Select(i => new NameValue() { Name = Culture.DateTimeFormat.GetMonthName(i), Value = i }).ToList();
             years = Enumerable.Range(YearFrom, YearTo - YearFrom + 1)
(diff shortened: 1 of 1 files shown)


akorchev:
Hi @n4213-ndensan,

Could elaborate of the necessity of this PR? What case does it solve which cannot be addressed by YearFormat?

n4213-ndensan:
Hi @akorchev ,

The reason for this new pull request is that our custom system is highly sensitive to the addition of new era names. When a new era name is introduced, the response speed of the built-in .NET code is not sufficient. Therefore, it is necessary to implement custom code to ensure a prompt response.

akorchev:
I see.

Can't we just set YearFormatter to FormatYear by default and avoid the null check?

```
public Func<int, string> YearFormatter { get; set; } = FormatYear;
```

n4213-ndensan:
sure you can. It seems more better code.

n4213-ndensan:
Due to the fact that the FormatYear method is an instance method, it cannot be used for property initialization. Would it be acceptable to initialize the YearFormatter in the constructor instead?

```
/// <summary>
/// Constructs the date picker.
/// </summary>
public RadzenDatePicker()
{
    YearFormatter = FormatYear;
}
```

n4213-ndensan:
I have modified the code to initialize the YearFormatter property in the constructor and removed the null check.

PR #2017, opened 2025-03-05
Select() by string will use source.ElementType if the Provider is EntityQueryProvider

XorZy:
Select() by string will use source.ElementType if the Provider is EntityQueryProvider

### DIFF
--- a/Radzen.Blazor/DynamicExtensions.cs
+++ b/Radzen.Blazor/DynamicExtensions.cs
@@ -83,6 +83,13 @@ namespace System.Linq.Dynamic.Core
         /// </summary>
         public static IQueryable Select<T>(this IQueryable<T> source, string selector, object[] parameters = null)
         {
+            
+            if (source.Provider.ToString() == "Microsoft.EntityFrameworkCore.Query.Internal.EntityQueryProvider")
+            {
+                return source.Cast(source.ElementType).Select(selector,
+                    expression => ExpressionParser.ParseLambda(expression, source.ElementType));
+            }
+            
             if (source?.ElementType == typeof(object))
             {
                 var type = source.FirstOrDefault()?.GetType() ?? typeof(object);


enchev:
This code will not work with your change:
```
var data = dbContext.Orders.Cast<object>().Select("x => x.OrderDetails");
```

XorZy:
> This code will not work with your change:
> 
> ```
> var data = dbContext.Orders.Cast<object>().Select("x => x.OrderDetails");
> ```

Yes indeed, in my project I actually retrieve the DbSet from a type using this method
 ```
  private static IQueryable<object> Set(DbContext context, Type t)
    {
        return (IQueryable<object>)context.GetType().GetMethod("Set", []).MakeGenericMethod(t).Invoke(context, null);
    }
```

so although the returned object is an ` IQueryable<object>` the underlying IQueryable preserves the EntityType.
I think I was incorrect in assuming that a DbSet preserves the original EntityType even when using Cast<>.

The code in your earlier [commit](https://github.com/radzenhq/radzen-blazor/commit/d52d91852a2a2d90338515ee7777e3251c6f8af5) will indeed work with this use case, but not with mine.
(I think it also makes additional DbRequests just to fetch the type, which is probably not desirable)
So how about this instead?

```csharp
if (source.ElementType != typeof(T))
{
    return source.Cast(source.ElementType).Select(selector,
        expression => ExpressionParser.ParseLambda(expression, source.ElementType));
}
```

enchev:
Not sure why you casting to object when you have already the type. This code:
```
private static IQueryable<object> Set(DbContext context, Type t)
{
   return (IQueryable<object>)context.GetType().GetMethod("Set", []).MakeGenericMethod(t).Invoke(context, null);
}
```

can be:
```
private static IQueryable Set(DbContext context, Type t)
{
    return (IQueryable)context.GetType()
        .GetMethod("Set", Type.EmptyTypes)!
        .MakeGenericMethod(t)
        .Invoke(context, null)!;
}
```

XorZy:
> Not sure why you casting to object when you have already the type. This code:
> 
> ```
> private static IQueryable<object> Set(DbContext context, Type t)
> {
>    return (IQueryable<object>)context.GetType().GetMethod("Set", []).MakeGenericMethod(t).Invoke(context, null);
> }
> ```
> 
> can be:
> 
> ```
> private static IQueryable Set(DbContext context, Type t)
> {
>     return (IQueryable)context.GetType()
>         .GetMethod("Set", Type.EmptyTypes)!
>         .MakeGenericMethod(t)
>         .Invoke(context, null)!;
> }
> ```

You are correct that the cast is unnecessary, quite frankly this is a piece of code I inherited and didn't even know was there until it started throwing exceptions on Radzen 6.x.
But I think the real problem is that Radzen 'hijacks' Dynamic Linq without offerring 1:1 compatibility.
The current use case is arguably an edge case, but it's apparently a common enough oversight that it affected me.

enchev:
We cannot offer 1:1 compatibility with Dynamic Linq  nor we plan to do that. I'm closing this pull request.

XorZy:
Well, in that case could you at least provide additional guidance on how to override the Radzen Extension methods?
In the repro I linked,  there is a reference to`<PackageReference Include="System.Linq.Dynamic.Core" Version="*" />` as suggested in your [blog post,](https://forum.radzen.com/t/just-released-our-new-major-version-of-radzen-blazor-6-0-0-without-reference-to-system-linq-dynamic-core-nuget-package/19748?u=enchev) yet the Radzen Extension Methods are still being used instead.

enchev:
I've updated the post with additional instructions. Keep in mind that you can always go back to 5.x - nothing force you to use 6.x version. The DataGrid in 6.x will output only valid C# expressions that in some cases cannot be parsed by Dynamic LINQ. Additionally since this library parser is too permissive you will might be exposed to vulnerability which is the core problem behind our change.

PR #2025, opened 2025-03-10
Added SameSite and Secure options for theme cookie.

FrankNWB:
Added SameSite and Secure options for theme cookie. 
This allows you to set the theme cookie with a SameSite attribute and thus hide the warning that is displayed in Chrome, for example. In addition, the new IsSecure option can be used to set the “Secure” flag of the cookie (only HTTPS requests are then possible).

Please note that I have created a new enum "CookieThemeServiceOptionsSameSiteMode" instead of using the built-in "SameSiteMode" enum defined in Microsoft.AspNetCore.Http because it is not yet referenced in the project and I didn't want to change that.

Regards
Frank

### DIFF
--- a/Radzen.Blazor/CookieThemeService.cs
+++ b/Radzen.Blazor/CookieThemeService.cs
@@ -8,6 +8,25 @@ using Microsoft.JSInterop;
 namespace Radzen
 {
     /// <summary>
+    /// Specifies the SameSite attribute for the cookie.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


akorchev:
Hi @FrankNWB,

Thank you for the pull request. Could you please rename `CookieThemeServiceOptionsSameSiteMode` to something shorter e.g. `CookieSameSiteMode`?

FrankNWB:
HI @akorchev,
I've renamed it :)

PR #2037, opened 2025-03-15
New component: time span picker

Cosmatevs:
Time span picker
This pull request contains:
- a new component, similar to the date picker in appearance, behavior, and configuration,
- tests for the new component,
- component style tweaks for free themes,
- demos for the new component, covering most configuration properties,
- an update of the form field component demo to include the new component,
- ~a new helper component: `PopupOrInline`, which makes it much easier to handle displaying a part of the component as a popup or just inline,~
- a new enum: `TimeSpanUnit` used to indicate time unit fields that should be shown in the picker
- a fix for the clear button position in the date picker inside a form field – because the implementation of that clear button is the same as in the new component
- additional 'time' tag for the date picker in the demo list – to make it easier to find it if you want just a time picker.

Things I couldn't do:
- component style tweaks for premium themes,
- documentation.

To take into consideration:
- implementing a time span filter for the data grid component.
- implementing some solutions from my closed pull request #1958, especially my comment from 2025-02-28,
- reimplementing popup/inline handling in the date picker ~using the `PopupOrInline` component~ in a similar way. The current solution requires a lot of micromanaging (mostly js calls) to make it work and seems prone to breaking.


Radzen.Blazor/Rendering/Popup.razor
```diff
@@ -30,12 +28,14 @@
     [Parameter]
     public EventCallback Close { get; set; }
 
+    public bool IsOpen { get; private set; }
```
enchev, line 31: Not sure why this change is needed. Usually properties should be parameters - it will be better to leave it as field.

Radzen.Blazor/Rendering/PopupOrInline.razor
```diff
@@ -0,0 +1,54 @@
+﻿@inherits RadzenComponent
```
enchev, line 1: In my opinion this component is not needed. You can close the popup if Inline is changed in the main component.

Radzen.Blazor/Rendering/Popup.razor
Cosmatevs: It was used by the `InlineOrPopup` component, and also by the TimeSpanPicker itself in the past. But now this change is no longer needed so I reverted it.

Radzen.Blazor/Rendering/PopupOrInline.razor
Cosmatevs: I forgot that `RenderFragment` exists when I created this component. Removed it now.

PR #2054, opened 2025-03-25
#2011 Fix: Allow Space for Selection in DropDown

### LINKED ISSUE #2011: RadzenDropDown selection by spacebar
RadzenDropDown, the ability to select an option with a spacebar.

**To Reproduce:**
1. Set Multiple and AllowFiltering to true.
2. Try to search some.
3. When a result appears move down by keyboard (arrow-down).  

Image

4. Try to choose an option by spacebar.
5. Space will be added to the search keyword and the result will be empty.

Image

**Expected:**
Select an option by spacebar and if you type a character or digit (etc) it corrects the search word.

Here is an [example](https://blazor.syncfusion.com/demos/multiselect-dropdown/checkbox) of the new behavior I am trying to describe:

srielango:
This PR allows using the Spacebar to select items in multi-select RadzenDropDown. It also prevents page scrolling when pressing Space.

Backwards-compatible – falls back if Radzen.Blazor.js isn’t updated due to web browser caching.
Enter key behavior remains unchanged for form submissions.

Closes #2011.

### DIFF
--- a/Radzen.Blazor/DropDownBase.cs
+++ b/Radzen.Blazor/DropDownBase.cs
@@ -678,7 +678,32 @@ namespace Radzen
                     //
                 }
             }
-            else if (key == "Enter" || key == "NumpadEnter" || key == "Space")
+            else if(key == "Space")
+            {
+                preventKeydown = true;
+
+                if (selectedIndex >= 0 && selectedIndex <= items.Count() - 1)
+                {
+                    var itemToSelect = items.ElementAtOrDefault(selectedIndex);
+
+                    await JSRuntime.InvokeAsync<string>("Radzen.setInputValue", search, $"{searchText}".Trim());
+
+                    if (itemToSelect != null)
+                    {
+                        await OnSelectItem(itemToSelect, true);
+                    }
+                    var popupOpened = await JSRuntime.InvokeAsync<bool>("Radzen.popupOpened", PopupID);
+
+                    if (popupOpened)
+                    {
+                        if (!Multiple)
+                        {
+                            await ClosePopup(key);
+                        }
+                    }
+                }
+            }
+            else if (key == "Enter" || key == "NumpadEnter")
             {
                 preventKeydown = true;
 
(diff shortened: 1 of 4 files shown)


enchev:
Thanks for the pull request! Please verify if the space can be used during filtering.

srielango:
@enchev 

Thanks for the review! I tested it with filtering enabled. The Spacebar behaves as expected:

While typing in the filter, it adds a space normally.
When navigating with the arrow keys, pressing Space selects/deselects the highlighted item.

Let me know if you need any additional changes!

enchev:
Here is what I received when attempted to run DropDown demos in this branch:
image

srielango:
@enchev ,

Thanks for bringing this to my attention.

I’ve implemented the fix following the existing pattern used in preventArrows. The JavaScript now correctly prevents space selection using keyCode === 32, and the C# call to JSRuntime.InvokeVoidAsync is wrapped in a try-catch to prevent errors.

I initially missed testing the scenario where the JavaScript code doesn’t exist, but I’ve now verified it and ensured it handles that case gracefully. I’ve tested this in Chrome and Edge, and it’s working as expected.

Could you please pull the latest changes and verify on your end? If the issue persists, let me know your browser version and environment details, and I'll investigate further.

Thanks!

srielango:
I’ve updated the implementation to ensure that pressing Space in the DropDown filter never closes the popup. Here’s the new behavior:

Pressing Space in the filter input refines the results as expected.
Using the arrow keys to select an item and pressing Space will now select the item but keep the dropdown open.
Pressing Enter, however, will still select the item and close the dropdown.

This keeps the filtering behavior consistent while maintaining flexibility in item selection. Let me know if this works as expected on your end or if you see any potential improvements!

enchev:
Thanks for the update however now SPACE will never close the popup after select which is breaking change compared to current official

enchev:
I've added the requested functionality and I'm closing this:
https://github.com/radzenhq/radzen-blazor/commit/e81ea15bb056d638c7eb1efb7af1aee82193ca7f

PR #2059, opened 2025-03-27
Radzen date time picker

bc3:
possible workaround


enchev:
This is a breaking change and will affect all DatePicker users that are using time part of the date. I'm afraid that we cannot accept this pull request.

PR #2085, opened 2025-04-10
Added ShowInput property to RadzenDatePicker

leon42:
Added a ShowDateBox property to RadzenDatePicker to allow only the button to be shown. See the demo for an example.
Added demos for ShowButton and ShowDateBox options.


enchev:
Thanks @leon42! In my opinion ShowInput will be better since there is already AllowInput property.

leon42:
I renamed it to ShowInput

PR #2115, opened 2025-04-22
Added DisplayToolTipFor Method to RadzenChart.razor.cs. 

Quasmo:
Added DisplayToolTipFor Method to RadzenChart.razor.cs.  Made Series property public.

### DIFF
--- a/Radzen.Blazor/RadzenChart.razor.cs
+++ b/Radzen.Blazor/RadzenChart.razor.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using System.Linq;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace Radzen.Blazor
 {
--- a/Radzen.Blazor/RadzenChart.razor.cs
+++ b/Radzen.Blazor/RadzenChart.razor.cs
@@ -95,7 +95,7 @@ namespace Radzen.Blazor
 
         internal ScaleBase CategoryScale { get; set; } = new LinearScale();
         internal ScaleBase ValueScale { get; set; } = new LinearScale();
-        internal IList<IChartSeries> Series { get; set; } = new List<IChartSeries>();
+        public IList<IChartSeries> Series { get; set; } = new List<IChartSeries>();
         internal RadzenColumnOptions ColumnOptions { get; set; } = new RadzenColumnOptions();
         internal RadzenBarOptions BarOptions { get; set; } = new RadzenBarOptions();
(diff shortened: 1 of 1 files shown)


Radzen.Blazor/RadzenChart.razor.cs
```diff
@@ -95,7 +95,7 @@ public partial class RadzenChart : RadzenComponent
 
         internal ScaleBase CategoryScale { get; set; } = new LinearScale();
         internal ScaleBase ValueScale { get; set; } = new LinearScale();
-        internal IList<IChartSeries> Series { get; set; } = new List<IChartSeries>();
+        public IList<IChartSeries> Series { get; set; } = new List<IChartSeries>();
```
akorchev, line 98: Public members require XML documentation. Please add some.

Radzen.Blazor/RadzenChart.razor.cs
```diff
@@ -427,6 +427,37 @@ internal async Task DisplayTooltip()
...
+            {
+                throw new ArgumentException($"Series:{series.GetTitle()} does not exist in {nameof(this.Series)}");
+            }
+
+            if (IsJSRuntimeAvailable)
+            {
+                Point ttpos = null;
+                try
+                {
+                    ttpos = series.GetTooltipPosition(data);
```
akorchev, line 448: GetTooltipPosition never returns null and should never thrown an exception. Please remove this try catch block and the null check.

Radzen.Blazor/RadzenChart.razor.cs
```diff
@@ -427,6 +427,37 @@ internal async Task DisplayTooltip()
...
+        public async Task DisplayTooltipFor(IChartSeries series, object data)
+        {
+            if (!Series.Contains(series))
+            {
+                throw new ArgumentException($"Series:{series.GetTitle()} does not exist in {nameof(this.Series)}");
+            }
+
+            if (IsJSRuntimeAvailable)
+            {
+                Point ttpos = null;
```
akorchev, line 445: I think `ttpos` is kind of a cryptic name. Use either `point` or just `position`.

akorchev (changes requested):

Radzen.Blazor/RadzenChart.razor.cs
Quasmo: After thinking about it. I appreciate not having full access to the getter/setter. Too much is linked to the Series object, for a more stable solution, I have created a GetSeries method that returns a IReadOnlyList<IChartSeries> Additionally, I have added the requested XML documentation.

Issue #2088, opened 2025-04-11
RadzenFormField Text not displaying properly


Techoneshot:
**Describe the bug**
I used to use `<span>` with class in FormField labels for required fields, now it stoped working. I have tried reverting version of Radzen Components but it looks like Radzen.js issue.
```
<style>
.required-field::after {
    content: "*";
    color: red;
    margin-left: 2px
}
</style>
    <RadzenFormField AllowFloatingLabel="false" Text="Test<span class='required-field'>Fixed label</span>" >
        <RadzenTextBox @bind-Value="@value" />
    </RadzenFormField>

@code {
    string value = "";
}
```
**Expected behavior**
Should render HTML used in Text as before

**Screenshots**
Before:
![Image](https://github.com/user-attachments/assets/0430f600-122e-4884-a5d4-5fa79af49728)
After:
![Image](https://github.com/user-attachments/assets/c00cabf6-a1c5-441b-8288-ab64d9d2b55c)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: MS Edge

akorchev:
Indeed the Text property no longer accepts HTML strings. This is a new change introduced in Radzen.Blazor 6.4.0. You can use CSS to achieve the same effect without the extra markup

```
<style>
.required-field .rz-form-field-label::after {
    content: "*";
    color: red;
    margin-left: 2px
}
</style>
    <RadzenFormField AllowFloatingLabel="false" class="required-field" Text="Test" >
        <RadzenTextBox @bind-Value="@value" />
    </RadzenFormField>

@code {
    string value = "";
}
```

Techoneshot:
Thank you

PR #2116, opened 2025-04-22
Use HTML as Text in RadzenFormField

bikotoru:
With this you can use HTML and thus, for example, leave an icon of another color or something else within the title of an input.

image
image
image


akorchev:
We have removed MarkupString support globally on purpose (it introduces a vector for XSS attacks). Thus we won't accept pull requests adding it back.

akorchev:
This is [very easy to implement with CSS only](https://github.com/radzenhq/radzen-blazor/issues/2088#issuecomment-2796239530).

PR #2117, opened 2025-04-22
Add TextTemplate parameter to RadzenFormField Blazor component

bikotoru:
A change was made to not use MarkupString but to be able to send a RenderFragment and in this way you can have the same functionality as using MarkupString

image
image

In case it is not an allowed solution, please let me know if I can receive feedback so I can go down the right path.

### DIFF
--- a/Radzen.Blazor/Radzen.Blazor.csproj
+++ b/Radzen.Blazor/Radzen.Blazor.csproj
@@ -10,7 +10,7 @@
     <IsPackable>true</IsPackable>
     <PackageId>Radzen.Blazor</PackageId>
     <Product>Radzen.Blazor</Product>
-    <Version>6.6.1</Version>
+    <Version>6.6.2</Version>
     <Copyright>Radzen Ltd.</Copyright>
     <Authors>Radzen Ltd.</Authors>
     <Description>Radzen Blazor is a set of 90+ free native Blazor UI controls packed with DataGrid, Scheduler, Charts and robust theming including Material design and Fluent UI.</Description>


Radzen.Blazor/RadzenFormField.razor
```diff
@@ -13,7 +13,17 @@
         </div>
         }
         @ChildContent
-        <label class="rz-form-field-label rz-text-truncate" for=@Component>@Text</label>
+               
+                <label class="rz-form-field-label rz-text-truncate" for=@Component>
```
akorchev, line 17: The existing indentation must be kept. Implement it like this: https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenLink.razor#L14

akorchev (changes requested):

Radzen.Blazor/RadzenFormField.razor
bikotoru: @akorchev  se aplica el cambio

Radzen.Blazor/RadzenFormField.razor
```diff
@@ -1,32 +1,32 @@
...
+        </CascadingValue>
+    </div>
+    @if (Helper != null)
+    {
+    <div class="rz-form-field-helper">
+        @Helper
+    </div>
+    }
+</div>
+}
```
akorchev, line 32: You have changed the formatting of the whole file. Please modify just the line that renders the text.

akorchev (changes requested):

Radzen.Blazor/RadzenFormField.razor
```diff
@@ -1,32 +1,32 @@
...
+    <div class="rz-form-field-content">
+        <CascadingValue Value=@context>
+        @if (Start != null)
+        {
+        <div class="rz-form-field-start">
+            @Start
+        </div>
+        }
+        @ChildContent
+        <label class="rz-form-field-label rz-text-truncate" for=@Component>@if (TextTemplate != null) {@Text} else {@TextTemplate}</label>
```
evandixon, line 16: Should this if statement be reversed? Looks like TextTemplate will only be used if it's null

PR #2135, opened 2025-04-28
Add support to render header template after column grouping/picker in a DataGrid

askids:
I have added a new parameter to control, if the header template should be rendered before or after Column grouping/Column Picker. It will retain current default behavior, where header template is rendered First before the column grouping/picker.

I needed this option to show a search text box after the column header picker. Since existing parameters, did not provide such option, added this support using new parameter.


enchev:
Thanks for the pull request however might be easier to use CSS instead:
https://forum.radzen.com/t/grid-column-picker-position/16052/2

enchev:
image

askids:
Thanks @enchev That worked for my requirement.

PR #2144, opened 2025-05-07
V7

enchev:


akorchev:
>  I want to make sure you are aware that it is no longer "Removing" and "Adding" child components placed inside an AccordionItem and you are now just using CSS instead.

Yes, we are aware of this change. It was required to make animations work. It is indeed a breaking change and we will mention it in the changelog.

Issue #2050, opened 2025-03-21
Feature request: Supply Appointment data to SchedulerSlotRenderEventArgs


friis-kraas-trinidat:
The RadzenScheduler offers the ability to render Appointments in custom colors by providing the TItem Data attribute. The RadzenYearView does not support this currently.

 It would be nice to have the same functionality for the RadzenYearView, by including a List of TItem Appointments to the SchedulerSlotRenderEventArgs that contains all Appointments for the slot being rendered. 
That way, the SlotRender Callback can be used to set Attributes depending on the appointment data within that slot.

akorchev:
Hi @friis-kraas-trinidat,

We won't mind a pull request with the suggested implementation. Would you give it a try?

PR #2154, opened 2025-05-14
Supply Appointment data to SchedulerSlotRenderEventArgs - #2050

paulo-rico:
As per request [#2050](https://github.com/radzenhq/radzen-blazor/issues/2050), pass `AppointmentData` as a property in `SchedulerSlotRenderEventArgs` class.

Update demo site `Year` view to highlight this change.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/Rendering/YearView.razor
+++ b/Radzen.Blazor/Rendering/YearView.razor
@@ -101,7 +101,7 @@
 
     IDictionary<string, object> Attributes(DateTime start, string className, bool slotInMonth)
     {
-        var attributes = Scheduler.GetSlotAttributes(start, start.AddDays(1), AppointmentsInRange(start, start.AddDays(1)));
+        var attributes = Scheduler.GetSlotAttributes(start, start.AddDays(1), () => AppointmentsInRange(start, start.AddDays(1)));
         attributes["class"] = ClassList.Create(className).Add(attributes).ToString();
         return attributes;
     }
--- a/Radzen.Blazor/Rendering/YearView.razor
+++ b/Radzen.Blazor/Rendering/YearView.razor
@@ -59,8 +59,8 @@
                                 var excessCount = appointments.Count();
                                 string classname = $"rz-slot-title {(day.Month != offSetMonth ? "rz-other-month" : "")} {(excessCount > 0 ? "rz-has-appointments" : "")} {(day == CurrentDate && month == CurrentMonth ? " rz-state-focused" : "")}".Trim();
 
-                                    <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-slot")>
-                                        <div class="@classname">
+                                    <div class="rz-slot" @onclick=@(args => OnListClick(day, appointments))>
+                                        <div @attributes=@Attributes(day, classname, day.Month == offSetMonth)>
                                             @day.Day
                                         </div>
                                     </div>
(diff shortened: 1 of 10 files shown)


akorchev:
Hi @paulo-rico,

Thank you for the PR. My only concern is that AppointmentsInRange/Slot are being called multiple times for the same slot (once during layout and once in GetAttributesForSlot). In most cases the Appointments property in the SlotRenderEventArgs is not going to be used including all existing ones. It would still lead to overhead though. Two possible solutions come to mind:

1. Store events per slot to not request them multiple times. Again seems redundant and memory consuming. Also prone to cache-miss/stale cache problems.
2. Get Appointments on demand. Maybe the best solution. Pass a `Func<IEnumerable<AppointmentData>>` to GetSlotAttributes. Then either assign it directly to SlotRenderEventArgs or make the Appointments property call this func the first time it is used:

```cs
private Func<IEnumerable<AppointmentData>> getAppointments;
private IEnumerable<AppointmentData> appointments;

public IEnumerable<AppointmentData> Appointments => appointments ?? = getAppointments();
```

paulo-rico:
Hi @akorchev

Must say, I wasn't happy myself with calling the same function more than once per slot. Also, wasn't too sure about calling and storing the appointments. Just needed the nudge in the right direction regarding passing the function itself, so went with suggestion 2.

Radzen.Blazor/SchedulerSlotRenderEventArgs.cs
```diff
@@ -17,6 +17,13 @@ public class SchedulerSlotRenderEventArgs
         /// The end of the slot.
         /// </summary>
         public DateTime End { get; set; }
+        // used to pass a function to get appointments on demand.
+        internal Func<IEnumerable<AppointmentData>> getAppointments;
+        private IEnumerable<AppointmentData> appointments;
+        /// <summary>
+        /// List of appointments.
+        /// </summary>
+        public IEnumerable<AppointmentData> Appointments => appointments ?? (appointments = getAppointments());
```
akorchev, line 26: You can use ??= here to make it shorter.

Radzen.Blazor/Rendering/YearView.razor
```diff
@@ -59,8 +59,8 @@
                                 var excessCount = appointments.Count();
                                 string classname = $"rz-slot-title {(day.Month != offSetMonth ? "rz-other-month" : "")} {(excessCount > 0 ? "rz-has-appointments" : "")} {(day == CurrentDate && month == CurrentMonth ? " rz-state-focused" : "")}".Trim();
 
-                                    <div @onclick=@(args => OnListClick(day, appointments)) @attributes=@Attributes(day, "rz-slot")>
-                                        <div class="@classname">
+                                    <div class="rz-slot" @onclick=@(args => OnListClick(day, appointments))>
```
akorchev, line 62: This looks as a breaking change as it changes the location where Attributes are applied.

akorchev (changes requested):

paulo-rico:
Hi @akorchev 

You're right. It is a breaking change. I'll move the `SlotRender` code back where it was.

I do think we need to be able to style the _day_ as in `AppointmentRender`. But it's not the same as `AppointmentRender` as the _day_  represents the existence of appointments as opposed to a single appointment in all the other views.

It's calling out for it's own method - `DayHighlightRender` or something like that, that is defined only in `RadzenYearView` and will include all `Appointments` for the day, similar to what has just been added to `SlotRenderEventArgs`.

What do you think?

Regards
Paul

akorchev:
How about using the SlotRender event as in other views? Right now all events are set at RadzenScheduler level and introducing event for a view would be a discrepancy.

paulo-rico:
I had used `SlotRender`, but in order to change the attributes on the "Day" as opposed to the "Slot", I had to move where the attributes were applied, and this was the breaking change.

Below is an image where the `SlotRender` is applied and rendered at the "Slot" level. As you can see, it's not ideal.

year-calendar-slot-render

It is still valid for a `SlotRender` to be applied for things like "Closed", "Today", "Holiday", e.t.c.

What do you think of having a `DayNumberRender` event at the Scheduler level? This would apply to Year, Planner, Timeline and Month views, and would style the day number in the slot.

akorchev:
DayNumberRender is misleading as it won't apply for every single day number but just for day view. Do we need this event at all? Shouldn't we just use AppointmentRender for this case too (days that have appointments)?

akorchev:
The original idea of this PR was "Supply Appointment data to SchedulerSlotRenderEventArgs". Let's stick to that.

paulo-rico:
Sorted (I think).

Issue #1925, opened 2025-01-25
RadzenDatePicker - Ability to hide virtual keyboard when date selected (mobile devices)


paulo-rico:
When using the `RadzenDatePicker` component on a keyboard-less device, the virtual keyboard will automatically display when a user has clicked a date, the popup has closed, and the manual entry field gets focus.

I would like to add in the ability to turn this "auto show" off if desired.

My proposed solution is this -

Add a `bool` property in `RadzenDatePicker.razor.cs` to control whether the virtual keyboard automatically displays or not -

```
        /// <summary>
        /// Specifies whether the virtual keyboard on keyboard-less devices is shown when input is focused.
        /// </summary>
        /// <value>True or False (default)</value>
        [Parameter]
        public bool HideVirtualKeyboard { get; set; } = false;

```

Then, add the following to the `<input @ref="@input"...` element of `RadzenDatePicker.razor` -

`        virtualkeyboardpolicy="@(HideVirtualKeyboard ? "manual" : "auto")"`

If this looks acceptable, I can put this into a PR.

Regards

Paul

PR #2157, opened 2025-05-16
Date picker - Disable automatic display of virtual keyboard - #1925

paulo-rico:
This PR addresses the following issue - [#1925](https://github.com/radzenhq/radzen-blazor/issues/1925)

Just to clarify. The current behaviour of the `DatePicker` is to show the on-screen keyboard on keyboardless devices as soon as a user has clicked a date in the popup. This can be quite frustrating as, when you have selected a date, you have to click elsewhere on the screen to close the keyboard.

This PR will give the option to disable that behaviour and not display the keyboard.

Regards

Paul


paulo-rico:
Additional - You can view this example at this [DevTunnel address](https://89b9wppj-44398.uks1.devtunnels.ms/datepicker#dateonly-timeonly). It's the last example on the page. Must obviously be run on a mobile or tablet.

enchev:
Thanks for the pull request @paulo-rico however I believe that the same can be achieved easily even now:
```
<RadzenDatePicker InputAttributes="@(new Dictionary<string,object>(){ { "virtualkeyboardpolicy", "manual" }})" />
```

paulo-rico:
Hi @enchev

This will serve my own requirements, as I don't need to manually enter date information in the `TextBox` portion. I will always select a date from the popup.

But, as I've tested this, it wouldn't suit all circumstances as it totally disables the textbox from displaying the virtual keyboard. As it stands, I think there are a couple of solutions;

1. After selecting a date from the popup, the textbox doesn't receive focus. However, I believe this would be a breaking change.
2. The developer would have to control this setting manually using their own button/switch e.t.c.

I'll close this PR with this comment. It may be raised in the future (it may not). Can be addressed if it does.

I will add an additional point.

Because of how it interfered with the flow of data entry on a mobile device i.e.,

```
**Expected**
1. Open popup of `DatePicker`
2. Select date
3. Popup closes
4. Move onto next field

**Actual**
1. Open popup of `DatePicker`
2. Select date
3. Popup closes
4. Touch some blank portion of screen to close virtual keyboard (auto opened because `TextBox` received the focus)
5. Scroll into view the next field (the display of the virtual keyboard causes an auto scroll to bring current field into view)
6. Move onto next field
```

I did actually think that it was inconvenient enough to warrant including in the Api. That, as a developer, is where I first went to look for some way of disabling it (and TBH, because I was Api mode, it never occurred to me to just add it into the `InputAttributes` `Dictionary` of my `RadzenDatePicker` declaration (cheers for that, @enchev).

Stay well

Paul

PR #2169, opened 2025-05-23
Error in the case where a non-generic class

GioeEng:
If a class is not generic but still implements the IDictionary<,> or IDictionary interface, the previous line of code would not work correctly.

For example, with a class like:

```
public class NonGeneric : IDictionary<string,object>
{
    // ... implementation ...
}
```


enchev:
IDictionary<,> does not implement IDictionary - your change will break immediately our demo: https://blazor.radzen.com/datagrid-dynamic?theme=material3

PR #2186, opened 2025-06-17
fix: avoid accessing parameter view in SetParametersAsync after an async call to prevent a race condition

vincentscode:
Fixes: https://forum.radzen.com/t/the-parameterview-instance-can-no-longer-be-read-because-it-has-expired/18912/7
See also: https://stackoverflow.com/questions/66717054/the-parameterview-instance-can-no-longer-be-read-because-it-has-expired-paramet

### DIFF
--- a/Radzen.Blazor/DataBoundFormComponent.cs
+++ b/Radzen.Blazor/DataBoundFormComponent.cs
@@ -301,19 +301,15 @@ namespace Radzen
         /// <returns>A Task representing the asynchronous operation.</returns>
         public override async Task SetParametersAsync(ParameterView parameters)
         {
-            // process all parameters first to prevent stale parameters
-            var searchTextChanged = parameters.DidParameterChange(nameof(SearchText), SearchText);
+            // check for changes before setting the properties through the base call
             var dataChanged = parameters.DidParameterChange(nameof(Data), Data);
             var disabledChanged = parameters.DidParameterChange(nameof(Disabled), Disabled);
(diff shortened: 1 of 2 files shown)


enchev:
This demo is not working as expected:
image

How should work:
image

You can check also the master branch for reference.

Radzen.Blazor/DataBoundFormComponent.cs
```diff
@@ -301,23 +301,24 @@ protected virtual IEnumerable View
...
         {
+            // process all parameters first to prevent stale parameters
             var searchTextChanged = parameters.DidParameterChange(nameof(SearchText), SearchText);
+            var dataChanged = parameters.DidParameterChange(nameof(Data), Data);
+            var disabledChanged = parameters.DidParameterChange(nameof(Disabled), Disabled);
+            var result = base.SetParametersAsync(parameters);
+
             if (searchTextChanged)
             {
                 searchText = parameters.GetValueOrDefault<string>(SearchText);
```
vincentscode, line 312: @enchev I am honestly quite confused by this line. I have not changed it, only moved it but I think it should say `nameof(SearchText)` instead of `SearchText` and was just ignored before as it was overriden by the `base.SetParametersAsync(..)` call? But I am also unsure why it even exists in the first place - `base.SetParametersAsync(..)` should already set the SearchText if it was supplied so there should be no reason to manually update the backing field.
Am I missing something here?

vincentscode:
It should work fine now, I have checked the DropDowns etc.

PR #2203, opened 2025-07-02
feat(datetimepicker) "Today" button

microappstech:
feat(datetimepicker): add "Today" button to DatePicker popup for quick selection of current date
  
image
image
image

### DIFF
--- a/Radzen.Blazor/DataBoundFormComponent.cs
+++ b/Radzen.Blazor/DataBoundFormComponent.cs
@@ -301,22 +301,19 @@ namespace Radzen
         /// <returns>A Task representing the asynchronous operation.</returns>
         public override async Task SetParametersAsync(ParameterView parameters)
         {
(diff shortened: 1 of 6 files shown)


enchev:
Not sure why we need this since you can do the same using FooterTemplate.

microappstech:
At first, my team and I needed a DatePicker with a "Today" button available everywhere in our app.
 We had two options:
- Either go through the whole codebase and refactor every place where the DatePicker is used to include a FooterTemplate with the "Today" button — which is a lot of work and not ideal.
- create a custom component that extends the DatePicker and adds this feature — but that also comes with performance concerns and makes the code harder to maintain.

That's why it would be much better if this feature was already part of the component itself.

enchev:
To enable this feature in your app you need to set the property to true to all instances of the component across the application - it’s far easier to simply replace RadzenDatePicker with MyDatePicker. Not sure what performance problems you are referring to when using your own custom component - in fact all pages in Blazor are custom components. I’m afraid that we will not merge this pull request.

PR #2210, opened 2025-07-08
Suppott filtering List<string> like sub-property.

gincsai:
Suppott filtering List<string> like sub-property. In this case no nested property so use the collectionItemTypeParameter.

### DIFF
--- a/Radzen.Blazor/QueryableExtension.cs
+++ b/Radzen.Blazor/QueryableExtension.cs
@@ -347,9 +347,9 @@ namespace Radzen
 
             ParameterExpression collectionItemTypeParameter = collectionItemType != null ? Expression.Parameter(collectionItemType, "x") : null;
 
-            if (collectionItemType != null && !string.IsNullOrEmpty(filter.FilterProperty) && filter.Property != filter.FilterProperty)
+            if (collectionItemType != null && filter.Property != filter.FilterProperty)
             {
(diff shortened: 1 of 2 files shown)


enchev:
Please add a demo demonstrating how this can be used.

gincsai:
The main problem is with Intersect. Intersect works in memory but not with Oracle (with the managed Provider) with this change, the GetExpression generatates the code with Contains for List<string> not only for List<MyClass>, so it can be used on backend filtering with oracle too.

The usage is the same as for sub-property filtering except that the filter property must be empty, as the string does not have properties. 

See the following, where Awards is List<string> and the filtering is Multiple.
```
<RadzenDataGridColumn Title="Awards" Property="Awards" Type="typeof(IEnumerable<string>)" FilterValue="@selectedAwards" FilterOperator="FilterOperator.In" FilterOperators="@(new FilterOperator[]{ FilterOperator.In, FilterOperator.NotIn})">
            <FilterValueTemplate>
                <RadzenDropDown @ref=sadd @bind-Value=@selectedAwards Style="width:100%;" 
                                Change=@OnSelectedAwardsChange Data="@(awards)" AllowClear="true" Multiple="true" />
            </FilterValueTemplate>
            <Template>
                @(string.Join(',', context.Awards))
            </Template>
        </RadzenDataGridColumn>
```

enchev:
I'm reffering the component demos where you can add additional demo for this functionallity following the format of other demos

Issue #2112, opened 2025-04-22
[Data Visualization] Spider & Sankey Charts


ehabhussein:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**

I would like to have a Spider & Sankey chart Component

**Describe alternatives you've considered**

[Spider Chart](https://demos.devexpress.com/blazor/ChartPolarSpiderWeb)
[SanKey Chart](https://demos.devexpress.com/blazor/ChartSankey)

**Additional context**
None

enchev:
already implemented

PR #2220, opened 2025-07-15
Add Sankey Chart Component to Radzen Blazor

ehabhussein:
## Summary
  This PR adds a new Sankey chart component to the Radzen Blazor library. Sankey diagrams are flow diagrams that visualize the transfer of quantities between nodes, making them perfect for showing resource flows, budget allocations, and process
  relationships.

  ## What's Included

  ### Core Component
  - `RadzenSankeyChart` - Main component with full Radzen design system integration
  - `SankeyChartModels.cs` - Data models for nodes and links
  - `SankeyLayout.cs` - Layout algorithm for node positioning and link path generation

  ### Features
  - **Theme Support**: Fully integrated with Radzen themes (dark/light mode compatible)
  - **Color Schemes**: Support for Pastel, Palette, Monochrome, and Divergent color schemes
  - **Node Alignment**: Configurable alignment (Justify, Left, Right, Center)
  - **Responsive Design**: Proper margins and scaling
  - **Smooth Animations**: Bezier curve links with hover effects
  - **Automatic Layout**: Smart node positioning with collision detection

  ### Configuration Options
  - `Nodes` - Collection of nodes with Id, Label, and optional Fill color
  - `Links` - Collection of links defining Source, Target, and Value
  - `NodeWidth` - Width of node rectangles (default: 24)
  - `NodePadding` - Vertical spacing between nodes (default: 8)
  - `NodeAlignment` - Horizontal alignment strategy
  - `ColorScheme` - Automatic coloring scheme
  - `Width/Height` - Chart dimensions
  - `Margins` - Customizable margins (Left: 60, Right: 60, Top: 10, Bottom: 10)

  ### Demo Pages
  - Interactive demo with live configuration
  - Multiple examples (basic flow, energy flow, budget flow)
  - Comprehensive documentation with code examples

  ## Screenshots
image

image

  ## Testing
  - Tested with all Radzen themes
  - Verified responsive behavior
  - Tested with various data sizes and configurations
  - Cross-browser compatibility verified

  ## Breaking Changes
  None - This is a new component addition.


akorchev:
Hi @ehabhussein,

Thank you for the PR and the amount of work you have committed to it. 

Full disclaimer - this is the first time I hear of this kind of chart and I know nothing about it :) Still I see a few things that need addressing:

1. All other Radzen components support binding to a collection of items e.g. a generic `IEnumerable<T>` property. For example RadzenChart and RadzenScheduler. The properties that are of importance are specified via other parameters. Case in point:

```
<RadzenAreaSeries  Data="@revenue2023" CategoryProperty="Date" ValueProperty="Revenue">
<RadzenScheduler  TItem="Appointment" Data=@appointments StartProperty="Start" EndProperty="End" 
                 TextProperty="Text" >
```
Can you refactor your implementation to support similar configuration instead of having custom model classes? We want to maintain consistency and also keep some compatibility with Radzen Blazor Studio which allows for GUI configuration of data bound components. We don't want the API of this component to be completely different to anything else in the suite.

2. I see there are Width and Height properties and also style="width: 100%; height: 400px". This feels odd. All other Radzen Blazor components only have a Style property. The actual size is determined via JavaScript if it is required for layout calculations. I believe this component should do the same. You can check the RadzenChart implementation for tips.

Thanks!

ehabhussein:
@akorchev Thanks for the response.

let me get these done and push these updates. then I'll work on adding the spider chart to Radzen, so we would have more charts for different use cases.

please see [Feature Request: [Data Visualization] Spider & Sankey Charts #2112](https://github.com/radzenhq/radzen-blazor/issues/2112) for the motivation behind this.

I really want to thank the entire team behind Radzen, as this is the most beautiful thing ever to come to Blazor.

akorchev:
A few other thoughts:

1. Should this be a series type for RadzenChart? For example:
    ```
    <RadzenChart>
       <RadzenSankeySeries ... />
   </RadzenChart>
   ```
   Other vendors seem to do it that way e.g. [Highcharts](https://www.highcharts.com/demo/highcharts/sankey-diagram). This would eliminate some of the sizing code I've mentioned.
2. If this can't be a series type for RadzenChart we should consider changing the name of the component to probably RadzenSankeyDiagram. Having two components that have "Chart" in their name (and potentially third) is not ideal.

ehabhussein:
Renaming to RadzenSankeyDiagram is a better choice. Moving it from charts to the Data visualization menu under SSRS Viewer. 

```razor
<RadzenSankeyDiagram TNode="SankeyNode" TLink="SankeyLink"
                   Nodes="@nodes" Links="@links" 
                   NodeWidth="24" NodePadding="8"
                   MarginLeft="60" MarginRight="60" MarginTop="10" MarginBottom="10"
                   Style="width: 100%; height: 400px;">
</RadzenSankeyDiagram>
```

akorchev:
I have checked the latest changes and I have further requests.

1. You have added two generic arguments to RadzenSankeyDiagram. I've checked the new Generic demo and got absolutely lost in the configuration - I suspect a lot of users would get lost too. Having two generic arguments and two data sources for a single component seems too complicated to me. This should be avoided at all costs.
1. The SankeyLink and SankeyNode classes remain and the new ones from the generic demo simply mirror them. This wasn't my idea for the refactoring. I am OK with keeping those classes but they shouldn't be a a part of the public API. I also think RadzenSankeyDiagram should have a single data source and a single type argument. RadzenChart doesn't have an intermediate internal model but RadzenScheduler does - it is the AppointmentData class which is not part of the public API - RadzenScheduler [converts](https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/RadzenScheduler.razor.cs#L645) to it. You can check the implementation for more details. This is the direction I want RadzenSankeyDiagram to follow.
1. I now see that the colors are set in the model as well which I don't think is the right solution. People would rarely have colors stored in their database. RadzenChart provides default colors for series that are index based and can be overridden via the Fills property of the series.
1. The Width and Height properties remain. This should be changed similar to other components otherwise the SankeyDiagram won't be responsive and won't resize properly when its size is given in other units than pixels - for example in percentage.
1. There are a few other visual and example issues but we can fix those after merge (e.g. every demo should be in its own .razor file and the routable page should use all of them within RadzenExample).

We have to follow the same principles as the other components in order to be able to maintain them in the future. Not to mention we don't have the domain knowledge 

By the was this code generated by AI? We don't mind but it is good to know beforehand. Sometimes they fail to grasp key concepts from the existing codebase.

ehabhussein:
@akorchev will probably push the changes this weekend to align with your comments.

akorchev:
Hi @ehabhussein,

Thank you for the updates. This looks better for sure. A few comments (probably last)

1. Can we get the labels similar to other properties (Source, Target, Value)? I don't like having another dictionary property just for the labels.
2. There are some protected members in RadzenSankeyDiagram (margin, width, height etc). Should they be protected or can we make them private?

ehabhussein:
on it! ill address those now and push them tonight.

Thanks for the patience

ehabhussein:
@akorchev would this be as you'd like ? also made the properties private

```csharp

    <RadzenSankeyDiagram TItem="FlowData"
                         Data="@flows" 
                         SourceProperty="Source"
                         TargetProperty="Target"
                         ValueProperty="Value"
                         SourceLabelProperty="SourceLabel"
                         TargetLabelProperty="TargetLabel"
                         NodeFills="@(useCustomColors ? customNodeColors : null)"
                         LinkFills="@(useCustomColors ? customLinkColors : null)"
                         ColorScheme="@colorScheme"
                         NodeWidth="24" NodePadding="8">
    </RadzenSankeyDiagram>
</div>

@code {
    public class FlowData
    {
        public string Source { get; set; }
        public string Target { get; set; }
        public double Value { get; set; }
        public string SourceLabel { get; set; }
        public string TargetLabel { get; set; }
    }
    
    private ColorScheme colorScheme = ColorScheme.Pastel;
    private bool useCustomColors = false;
    private IEnumerable<ColorScheme> colorSchemes = Enum.GetValues(typeof(ColorScheme)).Cast<ColorScheme>();
    
    private List<FlowData> flows = new List<FlowData>
    {
        new FlowData { Source = "A", SourceLabel = "Source A", Target = "C", TargetLabel = "Process 1", Value = 10 },
        new FlowData { Source = "A", SourceLabel = "Source A", Target = "D", TargetLabel = "Process 2", Value = 15 },
        new FlowData { Source = "B", SourceLabel = "Source B", Target = "C", TargetLabel = "Process 1", Value = 20 },
        new FlowData { Source = "B", SourceLabel = "Source B", Target = "D", TargetLabel = "Process 2", Value = 25 },
        new FlowData { Source = "C", SourceLabel = "Process 1", Target = "E", TargetLabel = "Target X", Value = 15 },
        new FlowData { Source = "C", SourceLabel = "Process 1", Target = "F", TargetLabel = "Target Y", Value = 10 },
        new FlowData { Source = "C", SourceLabel = "Process 1", Target = "G", TargetLabel = "Target Z", Value = 5 },
        new FlowData { Source = "D", SourceLabel = "Process 2", Target = "E", TargetLabel = "Target X", Value = 20 },
        new FlowData { Source = "D", SourceLabel = "Process 2", Target = "F", TargetLabel = "Target Y", Value = 15 },
        new FlowData { Source = "D", SourceLabel = "Process 2", Target = "G", TargetLabel = "Target Z", Value = 5 }
    };
    
    // Custom color scheme
    private List<string> customNodeColors = new List<string>
    {
        "#FF6B6B", // Source A - Red
        "#4ECDC4", // Source B - Teal
        "#45B7D1", // Process 1 - Blue
        "#96CEB4", // Process 2 - Green
        "#FFEAA7", // Target X - Yellow
        "#DDA0DD", // Target Y - Plum
        "#F4A460"  // Target Z - Sandy Brown
    };
    
    private List<string> customLinkColors = new List<string>
    {
        "#FF6B6B80", // A to C
        "#FF6B6B80", // A to D
        "#4ECDC480", // B to C
        "#4ECDC480", // B to D
        "#45B7D180", // C to E
        "#45B7D180", // C to F
        "#45B7D180", // C to G
        "#96CEB480", // D to E
        "#96CEB480", // D to F
        "#96CEB480"  // D to G
    };
}
```

akorchev:
Thank you for the updates! I've cleaned the demo a bit - removed the tutorial and list of properties as those are not needed. 

By the way the demos look very similar - the same properties are always set. Could you please update them so they are more unique? For example the first demo shows all features at the same time - custom colors etc. We should probably have the ability to change the node alignment in the first demo as well as part of the configuration.

I suggest to refactor the demos like this:
1. One "basic" demo with minimal parameters set.
2. One  "configuration" demo based on the current "Simple" (which is not simple at all) - allows the user to change the color scheme, node alignment and use custom colors.
3. Delete the other demos.

Once the demos are done we will merge the diagram and @yordanov would apply a few styling changes.

ehabhussein:
Perfect @akorchev. I'll send this over tomorrow max.

ehabhussein:
@akorchev , @yordanov   

Added

  - Tooltips - Show node and link details on hover
  - 4-layer demo - Added intermediate "Distribution" layer for more realistic energy flow visualization
image


  Fixed

  - Layout algorithm - Improved handling of unbalanced flows, ensuring nodes fill completely without gaps
  - Link rendering - Links now properly transition between different widths at source and target nodes

  Improved

  - Demo content - More comprehensive example with realistic data flow patterns
  - Visual consistency - Smoother path transitions and better spacing
  
 Removed

  - Redundant demos - Consolidated to just Basic and Configuration demos as requested

akorchev:
The demos look fine thanks! 

The tooltip however seems unstyled when I tested:

image

Can you check?

Also the tooltip uses hardcoded text which cannot be localized. You should introduce properties for that ValueText, IncomingText, OutgoingText etc. Also users often want to format the value somehow when displaying it (e.g. convert it to a currency) - you can add a ValueFormatter parameter similar to RadzenChart.

ehabhussein:
@akorchev, should be a straightforward fix. I'll have them in a few hours. Are there any other comments before I move onto another component I need to implement?

akorchev:
I think not for the time being.

ehabhussein:
@akorchev this should conclude all comments.

yordanov:
Introduced some style updates:

1. Removed the strokes around nodes and links as they take up unnecessary spacing that leads to a wrinkled look and feel, especially when there are lots of small nodes.
2. Replaced all SCSS variables with CSS ones
3. Added stroke to the labels to improve legibility

Currently working on porting the inline styles from the razor file to _sankey.scss.

SankeyStyles

@ehabhussein would you be able to enhance the way text position is calculated so that the lables are properly aligned? Currently the start and end nodes are misaligned. 

https://github.com/radzenhq/radzen-blazor/blob/492316038157d87e22920699ec4374d7a230a44f/Radzen.Blazor/RadzenSankeyDiagram.razor#L167

ehabhussein:
@yordanov 

image

image


image


I've looked at other charts and this is how mostly do it. Is this better ?

yordanov:
Yes, way better 👏🏻

ehabhussein:
Brilliant ! Great experience working with you @akorchev @yordanov on this. see you in the Next one.

dimitrivdw:
Hello, i'm trying to use this diagram and add flows or change flow values, but the component doesn't seem to show the changed values, it just show the first list of items. I don't see a "Reload" method or something similar to update the chart. Is this something that still has to be implemented?

PR #2233, opened 2025-07-23
feature: DataGrid Columns allow Sort and Filter when bound to Array<o…

EdJohnstonG9:
The objective is to use native display, filter and sort in a DataGrid<DataRow> 
This does not work atm as the DataRow.ItemArray[] returns object which cannot be filtered, but if DataColumn Type property is set then the propertyValueGetter is not set. The changes:

public static Type GetPropertyType(Type type, string property)// returns type of underlying array type when type==array<type>[]
GetNestedPropertyExpression(Expression expression, string property, Type type = null)// adds a Converter when property is array<T>[] and T != type
protected override void OnInitialized() // changes the logic here so that both Type can be set and propertyValueGetter is set under  these circumstances

The last is the change most needing of review. I have altered existing logic without fully understanding what it is handling. Maybe the logic should be added to handle the specific case where the underlying data object type != Type


EdJohnstonG9:
Optional further change to Common: line 3488, function rewrite for clarity and protection from AmbiguousMatchException
Not included because I am not sure it matches house style

```
        private static Type GetPropertyTypeIncludeInterface(Type type, string property)
        {
            if(type == null || type.GetProperties().Count(x=>x.Name == property) != 1)
            {
                return null;
            }
            var propType = type.GetProperties().FirstOrDefault(x => x.Name == property)?.PropertyType;

            return type.IsInterface ?
                new Type[] { type }
                    .Concat(type.GetInterfaces())
                    .FirstOrDefault(t => t.GetProperty(property) != null)?
                    .GetProperty(property)?.PropertyType :
                propType.IsArray ?
                    propType.GetElementType() :
                    propType;
        }

```

enchev:
Please add the example to the ExmpleService - at the moment it's unreachable. When manually navigating to the example I've indeed received the following error:
image
We cannot merge a feature without everything fully working.

EdJohnstonG9:
I have added the demo pages into ExampleService, name of "DataTable Source"
Apologies for the AmbiguousMatch; very embarrassing I left that in there on transferring changes to the fork.

New commit available

enchev:
I"ve changed the filtering type to test CheckBoxList filters and there are lot of exceptions:
image. 

Please update the demo description also since this sentence doesn't make sence:
```
Assign your System.Data.DataTable to the <code>GridDataTable</code> property
```
For sure there is no such property.

EdJohnstonG9:
Still working on the Filter by List issue
I think I am on the right track but are there any other features I should test before resubmitting?

enchev:
All existing DataGrid features should work.

EdJohnstonG9:
A quick update
I think I have fixed all the issues found so far and also simplified the solution (fewer changes)
However I do want to do more testing of scenarios like FilterProperty != Property
I am building some bUnit tests to cover these too, which is taking longer than expected...

EdJohnstonG9:
I have worked my way through all the issues I can find:
- Display of all tested data, including Enums
- Handling nullable values including enum?
- Filtering and sorting in all modes
I have added some test which for separation, I have put in a separate folder, driven from various data sources
I found a few UI issues when mixed mode filtering, for example:
Having a CheckBoxList filter in mixed grid shows a Input box which 
Throws an error if entered
Shows the name of data object after selecting items from the list

I will see if I can find the issue and suggest a fix.

enchev:
I received the follwoing error while tried the demo without changing anything:
image

image

EdJohnstonG9:
Yes that is the case I mentioned above
When a column is configured for checkbox list and others or the grid for Simple
Then the input box appears
And using it gives this error

same issue on the current live version
Thinking the best solution is to prevent render of input box in this case. Or instead render dropdown select box

EdJohnstonG9:
To reproduce this issue in the current Demo app:

Go to the page: https://blazor.radzen.com/datagrid-checkboxlist-filter

Edit the DG code so the Grid has Simple filters, one or more columns have CheckBoxList like this:

```
<RadzenDataGrid @ref="grid" AllowFiltering="true" AllowColumnResize="true" 
    FilterMode="FilterMode.Simple" PageSize="5" AllowPaging="true" AllowSorting="true" Data="@orders">
    <Columns>
        <RadzenDataGridColumn Property="OrderID" Title="Order ID" />
        <RadzenDataGridColumn Property="Customer.CompanyName" Title="Customer" AllowCheckBoxListVirtualization="false" />
        <RadzenDataGridColumn Property="ProductDiscontinued" Title="Discontinued" FormatString="{0}" FormatProvider="@myBooleanProvider" />
        <RadzenDataGridColumn Property="Employee.LastName" Title="Employee" FilterMode="FilterMode.CheckBoxList" >

```

The filter button CheckBoxList works OK, but then displays the Getter in the InputBox
The filter Input Box when populated manually creates the error

Reason being the CheckBoxList expects the FilterValue to be an enumerable, but this cannot be interpreted and shown in the InputBox 
Conversely text entered in the InputBox cannot be interpreted as IEnumerable<string>

enchev:
The demo you've made for binding to IEnumerable<DataRow> is not using CheckBosList filtering.

EdJohnstonG9:
Looks like I have not correctly configured whatever serves the right source code under the Edit Code button, but in the actual file. {ps found the issue here, however is I fix this, the Edit tab throws an error as the component is not in the Master repository}

In the actual DataGridDataTable.razor, I have 3 columns non my demo page that have checkboxlist, LastName, Title and HireDate covering String and Date fields

I get the same behaviour if the Grid is set to Simple, or just another column has a Simple filter mode


```
        <RadzenDataGridColumn Property="@colProperty(nameof(Employee.LastName))"
                              Type="@colType(nameof(Employee.LastName))"
                              Title="Last Name" Width="160px"
                              FilterMode="FilterMode.CheckBoxList" />
        <RadzenDataGridColumn Property="@colProperty(nameof(Employee.Title))"
                              Type="@colType(nameof(Employee.Title))"
                              Title="Job Title" Width="200px"
                              FilterMode="FilterMode.CheckBoxList" />
        <RadzenDataGridColumn Property="@colProperty(nameof(Employee.TitleOfCourtesy))"
                              Type="@colType(nameof(Employee.TitleOfCourtesy))"
                              Title="Title" Width="120px"
                              FilterMode="FilterMode.SimpleWithMenu" />
        <RadzenDataGridColumn Property="@colProperty(nameof(Employee.BirthDate))"
                              Type="@colType(nameof(Employee.BirthDate))"
                              Title="Birth Date" FormatString="{0:d}" Width="160px" />
        <RadzenDataGridColumn Property="@colProperty(nameof(Employee.HireDate))"
                              Type="@colType(nameof(Employee.HireDate))"
                              Title="Hire Date" FormatString="{0:d}" Width="160px"
                              FilterMode="FilterMode.CheckBoxList" />
```

Radzen.Blazor/RadzenDataGrid.razor
enchev, line 1: There is already fix for this in the master branch:
https://github.com/radzenhq/radzen-blazor/commit/206daba741121721a5ea1276f7ee7157fa9945fb

EdJohnstonG9:
I have applied a fix to this issue which simply prevents the display of an input box if the Filter expects an Enumerable i.e. CheckBoxList. 
I have not tried to put something else in place, as in this case, you will always have a Filter icon displayed in the top TH, but let me know if you think something else should be displayed in its place.

However I have also found another unrelated issue, with Filtering sub properties that I am working on

EdJohnstonG9:
@enchev 
I understand the issue with Filtering sub properties. I think it is my misunderstanding of the DataColumn property FilterProperty
Is it the intention that FilterProperty can be used to define the value/property when that Column has a Template and no Property value
Or is it the intention to allow one Property to be displayed (i.e. FirstName) and a different FilterProperty to be used for filtering (i.e. LastName)?

Hoping the first is the intention (rather than 2nd which was my assumption)

enchev:
The combination of Property/FilterProperty can be used in two scenarios:
- Use different property for filtering when Property or Type is not IEnumerable
- FilterProperty refers to collection item property name when Property or Type is IEnumerable

EdJohnstonG9:
Thanks for the previous info

I think I have all that working correctly now 
Also I have removed my 'fix' to FilterMode  Simple/CheckBoxList issue and merged with the latest Master code

However I have not pushed the latest version as I found another issue on both the Pull and Live code and have not yet resolved it

- Demo Page: https://blazor.radzen.com/datagrid-composite-columns?theme=material3
- Set Filter Mode: Check Box List
- Set any filter on any column
- Result: stack overflow

I suspect in generating the IQueryable<TItem> View in RadzenDataGrid it is getting into a recursive loop, but ran out of time today to track it down.

EdJohnstonG9:
I have found one issue which is not the cause and a theory that I cannot find a way to prove easily

The issue is changing from one Filter Mode to the other, does not update the FilterPropertyType to/from IEnumerable:

- When chagning to FilterMode.CheckBoxList from any other, the column needs to know the FilterPropertyType is an IEnumberable
- However this is only set during OnInitialize which is not called under those circumstances
- The reverse is true also, changing from CheckBoxList to any other
- This could result in some issues with the current live code but I think is accidentally improved with the Fork I am working on

Testing with the initial grid FilterMode = CheckBoxList, the issue persists but with the Expression created by public static IQueryable<T> Where<T>(.. being equivalent to simple grids

That leaves only this code in QueryableExtension.cs: 

```
public static IQueryable<T> Where<T>(this IQueryable<T> source,...
            var lambda = Expression.Lambda<Func<T, bool>>(combinedExpression, parameter);

            return source.Where(lambda);
```

It looks like the Where(lambda) is operating recursively, finding a parent or child Employee through Employee1/Employees1 which in turn is finding parent/child Employee etc, rather than operating only on the primary Data<Employee> enumerable/queryable.

However I cannot yet work out for sure if this is happening or how to stop it if it is.

enchev:
In my opinion these issues are unrelated to the DataTable binding and we can fix them in other iterations. I've not found anything specific that might cause a problem with this pull request and we ca merge it if you think you are ready.

EdJohnstonG9:
Agreed re the above issue

Ive pushed the merged code which also has the fix for nullable value filter columns hoping I have covered all cases now.

**One question re Tests:**
In my tests I deliberately use local json data, not Northwind data or direct tests of Demo pages as this could mess with CICD pipelines.
Would it be possible to build some tests of the Demo pages as built, to shorten the test cycle, or would that break your CICD?

**One question re DataTables:**
The logical next step to this change would be to allow DataGrid to accept a new DataTable parameter, and DataColumn to accept names of DataTable.Columns.
Would that be a step too far?? Or is your preference to allow users to create their own custom components?

enchev:
It’s OK in my opinion to use some sample data and not the demos. Accepting DataTable as Data directly is not something we want to support honestly.

EdJohnstonG9:
Hi @enchev anything holding this back from merging?
Let me know if you need anything more from me.

enchev:
Unfortunately I reverted this since the DropDownDataGrid component was completely broken. We will do our best to add DataTable support in the next update before the end of the week however we will use different approach - instead ItemArray the column name indexer similar to what we have already for IDictionary.

EdJohnstonG9:
Deepest apologies for missing that. Let me know if I can help in any way. 
I started by trying to use the Items indexer but ran into complications as it accepts int, DataColumn or string index values.
I thought about creating a dictionary but instead chose ItemArray as it appeared easier, except for the boxing/unboxing of nullables and enums; possibly something you handle better elsewhere.

EdJohnstonG9:
I have had another look and can fix that particular problem with DropDownDataGrid
There are important subtilties in the use of PropertyType and FilterPropertyType which I had broken for this object.

The fundamental issue is here though, in Common.cs which returns null when type == DataRow and can only return the correct underlying type when it gets access to an instance of DataRow. Hence the need to add Type parameter values etc

This I have worked around where I have found the issues, but as this function is called from a lot of different places, there is a significant risk of other issues cropping up because I have not found a combination of features that does not correctly handle the null return here.

```
        /// <summary>
        /// Gets the type of the property.
        /// </summary>
        /// <param name="type">The type.</param>
        /// <param name="property">The property.</param>
        /// <returns>Type.</returns>
        public static Type GetPropertyType(Type type, string property)
        {
            if (property.Contains("."))
            {
                var part = property.Split('.').FirstOrDefault();
                return GetPropertyType(GetPropertyTypeIncludeInterface(type, part), property.ReplaceFirst($"{part}.", ""));
            }

            return GetPropertyTypeIncludeInterface(type, property);
        }
```

enchev:
The DataTable support is already in the master branch and will be released tomorrow:
https://github.com/radzenhq/radzen-blazor/commit/46118964ead84aeaf5d1bff2ddf490b184782452

EdJohnstonG9:
Well that is an abject lesson in how to make a neat and tidy change if you know what you are doing!

Many thanks. Eating plenty of humble pie!

EdJohnstonG9:
I was trying to work out how you did this so easily, and this just solves so many issues right off the bat
`
public static string GetDynamicPropertyExpression(string name, Type type)
`

However there is still an issue rendering Enums which currently are displaying as their Int value not the representation.

GetDynamicPropertyExpression deliberately returns (Enum)it[{property}] for all enums, rather than the underlying type- not sure why. 
Even if changed, the function GetNestedPropertyExpression cannot add the relevant converter of int->{myEnumType}

I resolved this in a very clumsy way but sure you must know a better way.

PR #2251, opened 2025-08-11
enhance tailwind compatibility

duongphuhiep:
## Problem:

Using a Tailwind CSS utility class to set the text color on a native `<input>` HTML element doesn't work in a Radzen project.

```html
<input class="text-fuchsia-700 text-7xl" /> <!---it doesn't work-->
```

## Root cause:

Radzen's stylesheet sets the style of the standard HTML input in an intrusive way:

```css
input {
  color: ...
}
```

Meanwhile, the Tailwind CSS utility sets the color in a more gentle way:

```css
@layer utilities {
   .text-fuchsia-700 {
     color:...
   }
}
```

The Tailwind CSS rule is more specific but cannot take precedence because it's defined inside a CSS layer. Tailwind follows the best practices of a modern CSS framework, using a CSS layer instead of `!important` everywhere.

## Correction

Modifying a native style of an HTML tag directly without any specificity (such as `@layer` or `.rz-` class) is generally not recommended, Instead, these rules are now placed within a `radzen` CSS `@layer`. This ensures Radzen's styles are applied but allows other CSS frameworks, such as Tailwind, to properly override or extend them when needed, making the rule less intrusive and more compatible.

## Note

This PR added only 2 lines of code, other changes are white-space indentation
image

### DIFF
--- a/Radzen.Blazor/themes/components/blazor/_input.scss
+++ b/Radzen.Blazor/themes/components/blazor/_input.scss
@@ -29,116 +29,118 @@ $input-disabled-placeholder-color: var(--rz-text-disabled-color) !default;
 $input-disabled-opacity: 1 !default;
 $input-transition: var(--rz-transition-all), width 0, height 0 !default;
 
-%input-hover {
-  box-shadow: var(--rz-input-hover-shadow);
-  background-color: var(--rz-input-hover-background-color);
-  border: var(--rz-input-hover-border);
-  border-block-end: var(--rz-input-hover-border-block-end);
-}
+@layer radzen {
+  %input-hover {
+    box-shadow: var(--rz-input-hover-shadow);
+    background-color: var(--rz-input-hover-background-color);
+    border: var(--rz-input-hover-border);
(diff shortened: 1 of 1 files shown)


duongphuhiep:
Once this PR is merged, then not only Tailwind, it would be possible to "marry" Radzen to other tailwind-based components such as <DaisyUI.com>. It means that, a typical Radzen application would 

- gain all DaisyUI themes / colors (Cupcake, Valentine, Halloween...): it is possible to bridge certain Radzen's themes (such as Default and Material) to use DaisyUI colors.
- gain all the  DaisyUI components in addition. Some of them is not consistent with Radzen components. Some looks nicer (arguably)

yordanov:
Thanks for the PR @duongphuhiep.

We’ll be closing this PR for now, as we consider the introduction of cascade layers a breaking change. Even if no visual differences are immediately noticeable, cascade layers alter how styles are applied. As a result, all components must be carefully reviewed - particularly those that rely on utility classes making use of the `!important ` rule (which we would also have to remove at some point).

For this reason, we would only consider adopting such an approach in a future **major** version.

Issue #2257, opened 2025-08-14
Datepicker time changes do not fire onChange event unless user clicks "ok" button.


mrfixij:
**Describe the bug**
When using the datepicker time component - the valueChanged callback is only called when the OK button is pressed and not when clicking outside of the popup to close it.  This also leaves the state in a confusing place because `value` is not updated in the below example.

**To Reproduce**
1. Using the basic Define Hour datepicker as a template: update the code to the following.
```<RadzenStack Orientation="Orientation.Horizontal" JustifyContent="JustifyContent.Center" AlignItems="AlignItems.Center" Gap="0.5rem" class="rz-p-12">
    <RadzenLabel Text="Select Date" Component="DatePickerHourFormat" />
    <RadzenDatePicker TValue=DateTime? Value=@value ValueChanged=@OnValueChanged ShowTime="true" HourFormat="12" DateFormat="MM/dd/yyyy h:mm tt" Name="DatePickerHourFormat" />
    
</RadzenStack>
<EventConsole @ref=@console />

@code {
    DateTime? value = DateTime.Now;
    EventConsole console;
    private void OnValueChanged(DateTime? newValue)
    {
        console.Log(newValue); 
        value = newValue.Value;
    }
}
```
2. Open the datepicker, change the date. Observe that the console has logged the dateTime updating.
3. Open the datepicker. Change the time using the up/down arrows on the UI.  Close the datepicker by clicking outside of the modal. Observe that the console has not logged the new time.
4. Repeat step 3, but close the datepicker by clicking the OK button instead.  Observe that the console has logged the new time.

**Expected behavior**
When closing the modal without the OK button, the ValueChanged event should fire

enchev:
This is intentional. When the Ok button is present value will be updated only when On button is pressed - if you set ShowTimeOkButton=false, value will be updated immediately. Changing this behavior will be a breaking change.

Anspitzen:
Current behavior seems correct, butt display of data on cancle is wrong.
On close without changing the bound Value displayed Value should reset to real data bound value and not stay on (wrong because not set) internal component value.
Else User WILL be confused as why the order/Birthday/whatever is saved incorrectly, even if the input is showing their correct data. Just because they did not click OK. It will show they did input the correct date without ever doing confirm. 

Thats at least how I would expect it to work

mrfixij:
[...]

PR #2259, opened 2025-08-14
Add support for datepicker callbacks, fire callback on close.

mrfixij:
This addresses https://github.com/radzenhq/radzen-blazor/issues/2257

Should allow the Close function in a datepicker to fire appropriately when clicking away from the datepicker instead of using the OK button.  This will provide a means to keep the time synced with the model without needing the OK button.

### DIFF
--- a/Radzen.Blazor/RadzenDatePicker.razor
+++ b/Radzen.Blazor/RadzenDatePicker.razor
@@ -27,7 +27,7 @@
             }
         }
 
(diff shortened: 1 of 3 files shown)


mrfixij:
Sorry about some of the minor formatting changes, my work VS autoformats on save.  If the whitespace changes are a dealbreaker I can rollback the formatting changes.

mrfixij:
This should probably not be merged pursuant to discussions in the linked thread.

Issue #2188, opened 2025-06-19
rz-layout height on mobile


albertolg:
Hi, on mobile devices the height of the layout in the material theme is not calculated right, as a workaround i did

`.rz-layout {
    height: 100dvh;
}
`

is possible to include it in a future release?

thanks

akorchev:
Hi @albertolg,

We need additional information about this issue:

1. What is not calculated right?
2. Does the problem appear in other themes or just Material?

albertolg:
hi @akorchev

1. when the sidebar have some menu and submenu and are expanded, some of them are not visible, I found that the 100vh on the mobile have issues, so after some research i found this article [https://medium.com/@alekswebnet/fix-mobile-100vh-bug-in-one-line-of-css-dynamic-viewport-units-in-action-102231e2ed56](https://github.com/radzenhq/radzen-blazor/issues/url)
2. all the themes are affected.

I searched in the css files and found 
`/* RadzenLayout styles */
.rz-layout {
  box-sizing: border-box;
  height: 100vh;
  overflow: hidden;
  display: grid;
  grid-template-columns: auto 1fr;
  grid-template-rows: auto 1fr auto;
  grid-template-areas: "rz-header rz-header" "rz-sidebar rz-body" "rz-footer rz-footer";
  background-color: var(--rz-layout-background-color);
}`

and the 
`@media (max-width: 768px) {
  body:has(> .rz-layout) {
    overflow-x: hidden;
  }
  .rz-layout {
    height: 100dvh;
  }
  .rz-header,
  .rz-footer,
  .rz-body {
    width: 100vw;
  }
  .rz-layout:has(> .rz-sidebar-collapsed + .rz-sidebar-expanded) {
    justify-content: end;
  }
  .rz-layout:has(> .rz-sidebar-collapsed + .rz-sidebar-expanded) .rz-header, .rz-layout:has(> .rz-sidebar-collapsed + .rz-sidebar-expanded) .rz-footer {
    margin-inline-start: auto;
  }
  .rz-layout:has(> .rz-sidebar-expanded + .rz-sidebar-expanded) .rz-body {
    width: auto;
  }
}`

where the dvh is used, but my tablet have an higher pixel width so it doesn't use it, am I wrong?

PR #2260, opened 2025-08-14
Expose Responsive MaxWidth in RadzenSidebar

mtboeke:
I would like the RadzenSidebar to automatically retract when the page is viewed on an iPad, which has a viewport of 820px.

### DIFF
--- a/Radzen.Blazor/RadzenSidebar.razor
+++ b/Radzen.Blazor/RadzenSidebar.razor
@@ -4,7 +4,7 @@
 {
     @if (IsResponsive)
     {
-        <RadzenMediaQuery Query="(max-width: 768px)" Change=@OnChange />
+        <RadzenMediaQuery Query="@Query" Change=@OnChange />
     }
     <div @ref="@Element" style="@GetStyle()" @attributes="Attributes" class="@GetCssClass()"  id="@GetId()">
         @ChildContent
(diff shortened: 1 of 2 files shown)


enchev:
Maybe ResponsiveMaxWidth should be string in order to accept not only pixels.

enchev:
@yordanov your input will be appreciated before merging this.

yordanov:
I think we should also remove the following CSS. It is not needed since the responsive logic is handled by RadzenMediaQuery, also width and opacity are controlled by the expanned/collapsed CSS classes respectively:

https://github.com/radzenhq/radzen-blazor/blob/95c42f3266c9970d0d83b116b48a2e34cec7d490/Radzen.Blazor/themes/components/blazor/_sidebar.scss#L22-L27

----

On a side note there is this issue https://github.com/radzenhq/radzen-blazor/issues/2188. Although it is not directly related to the Sidebar, it relates to a similar `ResponsiveMaxWidth` logic that needs to be implemented on a `RadzenLayout` level. And that makes me wonder how a `RadzenLayout` property would be interacting with a similar one in `RadzenSidebar`.

In short, we have `@media (max-width: 768px)` accross multiple components and at some point we need to come up with a solution that addresses all of them at once. Most probably a breaking change that would impact the solution in this PR as well.

Issue #1401, opened 2024-02-29
RadzenDropDown: when using multi-select, collections bound to value are set to null on clear


pianomanjh:
**Describe the bug**
When using `RadzenDropDown` to make multiple selections bound to a collection, with `AllowClear` set to true, a user can click to clear all choices from the collection.  This results in the bound collection property being set to null, instead of clearing the collection entries.  I understand this may be by design, but, especially when using nullable annotations in a codebase, results in easy-to-miss null reference exceptions.

**To Reproduce**
```
public class Foo
{
    public ICollection<string> Settings { get; set; } = new List<string>();
}

var foo = new Foo() { Settings = ["Two"] };
var values = new string[] { "One", "Two", "Three" };

<RadzenDropDown Multiple="true" AllowClear="true" @bind-Value="@foo" Data="@values" />
```
After clicking the X in the control to clear the selections, `foo.Settings` is null instead of an empty collection.

**Expected behavior**
If the property bound to value started as a non-null collection instance, I would expect that collection to remain, but be empty.

I understand this may be breaking behavior, so, as a potential mitigation, if the property bound to value started as null, we could have that property be set to null on clear...

**Additional context**
If this is something the team is ok changing, let me know and I'm happy to post a PR.

Riedi2070:
#vote up

enchev:
Blazor bindings works only null/not null expression - they will not understand if a property is collection, if you clear the collection, remove or add items. We prefer not to do that since it will be exception compared to all other two-way bindings.

pianomanjh:
> #vote up

@Riedi2070 this was resolved with #2261

Riedi2070:
> > #vote up
> 
> [@Riedi2070](https://github.com/Riedi2070) this was resolved with [#2261](https://github.com/radzenhq/radzen-blazor/pull/2261)

Thx u very much

PR #2261, opened 2025-08-15
RadzenDropDown/Grid: Modify bound collection in-place

pianomanjh:
I have hit another scenario where binding a collection to `RadzenDropDown` or `RadenDropDownGrid` is problematic.

I am using a custom collection that does not have a public default constructor.  `RadzenDropDown` uses `Activator.CreateInstance` to generate new collections on changes to the dropdown selection.  This makes the control unusable for this binding scenario.

I've posted another example in issue #1401 , where Clear nulls instead of Clears the bound collection.  There are other subtle issues that can occur when the bound collection is wholesale replaced (i.e. change tracking).  
In these cases I had to build component wrappers around the `RadzenDropDown` which are less than ideal.

In any case, I decided this latest limitation was a good time to try and resolve / support this scenario in a non-breaking way.

I've located the point in the code where these collections are replaced, and abstracted the existing behavior to a class, that can be swapped out with an implementation that preserves the bound collection, modifying it in place.

So as not to pollute the Parameter space, the opt-in flag is a protected field on `DropDownBase`, such that this behavior can be adjusted via subclassing.  I wasn't sure the appetite for this behavior, so did not want to take a broader approach until the maintainers have evaluated its utility and safety.

## Benefits ##
- A bound collection is modified in place, easing situations where
  - reference equality is important
  - it is not as feasible to rely on `Activator.CreateInstance` allocation, i.e. custom collection ctors
  - it is important to receive accurate notifications from collections implementing `INotifyCollectionChanged`
- Is entirely opt-in, the default behavior is maintained.  ~~I did make a small refactor to simplify some type gymnastics that seemed unnecessary, that can be easily reverted if warranted~~ (I reverted this to reduce risk)
- Performs well for large collections via use of HashSets internally

The diff looks large, but only due to moving existing code into a class.  Reviewers are encouraged to review commit by commit.

@SimonLissack

### DIFF
--- a/Radzen.Blazor.Tests/DropDownTests.cs
+++ b/Radzen.Blazor.Tests/DropDownTests.cs
@@ -2,7 +2,9 @@ using AngleSharp.Dom;
 using Bunit;
 using Microsoft.AspNetCore.Components;
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -43,6 +45,7 @@ namespace Radzen.Blazor.Tests
 
             return component;
         }
+
 
         [Fact]
         public async Task Dropdown_SelectItem_Method_Should_Not_Throw()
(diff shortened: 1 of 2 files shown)


pianomanjh:
A couple questions for maintainers:
1) Should `ValueChanged` be raised at all in the case where the collection instance does not actually change?
2) Along those lines, if a user does not assign a `ValueChanged` (i.e. does an assignment instead of a binding `Value="@MyCollection"`), should the control still update the contents of the collection?  It seems yes, but I notice this does not happen today

I reverted away from some refactoring I had done on some unnecessary type coercion.  If there is an appetite for this, I would be open to help simplify.

enchev:
In my opinion ValueChanged should be raised only if needed - i.e. `@bind-Value` is used.

pianomanjh:
@enchev any issues with the approach here, or changes requested?

enchev:
No issues found so far - we can merge it if you think you are ready.

pianomanjh:
Yep, I've tested it through my scenario and it fits the bill.  I did not find any regressions...if any are reported feel free to flag me

PR #2269, opened 2025-08-25
When filtering a column of type List<DateTime?> the filter value is D…

gincsai:
When filtering a column of type List<DateTime?> the filter value is DateTime (on server side the filter value is DateTime or null, but never DateTime?).

The type of the constatnt expression must match with the collection item type. So if the collectionItemType is not empty use that parameter instead of valueType.


enchev:
Can you provide details on how to reproduce the problem? Filtering by OrderDate on this demo gives `IEnumerable<DateTime?>` correctly:
https://blazor.radzen.com/datagrid-checkboxlist-filter?theme=material3

image
image

gincsai:
When filtering a column of type List<DateTime?> (multiple dates in one cell), and the filter value is DateTime. 
So you have a column with a custom template that shows the date list concatenated and a custom filter template that has a DateTimePicker. In this case, the filter value will be DateTime on the server (after deserializing the FilterDescriptors).

enchev:
Please provide runnable example demonstrating your scenario. You can use our demos for reference.

gincsai:
After a detailed investigation, this can be handled without a change in Radzen.

PR #2313, opened 2025-10-13
Fix does not contains (and NotIn) filtering on collection.

gincsai:
On the demo page, the Awards and WorkStatus DoesNotContain filtering shows invalid data. (Eg: does not countains Award3, Award4)
https://blazor.radzen.com/datagrid-filtervalue-template?theme=material3

### DIFF
--- a/Radzen.Blazor/QueryableExtension.cs
+++ b/Radzen.Blazor/QueryableExtension.cs
@@ -478,10 +478,12 @@ namespace Radzen
                 _ => null
             };
 
-            if (collectionItemType != null && primaryExpression != null &&
-                !(filter.FilterOperator == FilterOperator.In || filter.FilterOperator == FilterOperator.NotIn))
+            if (collectionItemType != null && primaryExpression != null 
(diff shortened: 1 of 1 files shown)


enchev:
This change will brake In/NotIn

gincsai:
How? The modified code will not run for In / NotIn as it was in the original code.
Is there any test that fails after this modification?

enchev:
You are right, there isn't test that fails - feel free to add such.

gincsai:
The IN operator is correct, but I think the NOTIN should return those items that were not in the IN result.
All items = IN + NOTIN
Current IN -> `list.Any( x => filter.Contains(x))`
Current NOTIN -> `list.Any( x => !filter.Contains(x))`
My code -> `list.All(x => !filter.Contains(x))` - These items were not in the IN result.

But this is a breaking change.
Do you see any option to pass a parameter to the filtering, specifying how to handle 'IN' and 'NOT IN'? I mean, use `Any()` or `All()`?

enchev:
I've added yesterday various tests realted to filtering. Try to rebase your branch to the master branch and check what's failing after your change.

Issue #2316, opened 2025-10-16
RadzenDataGrid: DateTimeOffset filter sends UTC datetime to LoadData instead of local datetime


bllkrkmz:
**Describe the bug**
When filtering a DateTimeOffset column in RadzenDataGrid, the date/time value delivered to the LoadData event is converted to UTC. The user selects a local date/time, but LoadData receives the value as UTC, which causes incorrect filtering behavior for local timezones.

**To Reproduce**
1. Use a model with a DateTimeOffset property and display it in a RadzenDataGrid column with filtering enabled.
2. Apply a filter to that DateTimeOffset column by selecting a date/time in the grid UI.
3. Inspect the value received by the LoadData event handler (or the filter sent to the server).
4. Observe that the received value is in UTC instead of preserving the local DateTimeOffset selected by the user.

**Expected behavior**
The date/time provided to LoadData for a DateTimeOffset column filter should preserve the local DateTimeOffset (or at least the selected local date/time) rather than being automatically converted to UTC.

**Desktop (please complete the following information)**
- OS: Windows 11
- Browser: Chrome
- Version: Latest

Issue #2315, opened 2025-10-16
RadzenDataGrid DateTime filtering fails with Npgsql and 'timestamp with time zone' fields


ogix:
**Describe the bug**
Filtering on a DateTime property in a RadzenDataGrid component does not work when using Npgsql Entity Framework Core Provider with Radzen. The issue occurs when the DateTime field is configured as 'timestamp with time zone' in the PostgreSQL database. The provider expects the DateTime to have a UTC kind, but Radzen likely provides it with an Unspecified kind.

**To Reproduce**
1. Set up a PostgreSQL database with a table containing a 'timestamp with time zone' column
2. Configure the Npgsql Entity Framework Core Provider in your project
3. Create a RadzenDataGrid component that includes the DateTime field
4. Enable filtering on the RadzenDataGrid for the DateTime column
5. Attempt to filter the data using the DateTime field in the RadzenDataGrid
6. Observe that the filtering does not work as expected

**Expected behavior**
The filtering on the DateTime property in the RadzenDataGrid should work correctly, regardless of the time zone configuration in the database.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 7.4.3

**Additional context**
This issue appears to be caused by a mismatch between how the RadzenDataGrid component handles DateTime values for filtering and how the Npgsql provider expects them when working with 'timestamp with time zone' fields. It may require coordination between Radzen's RadzenDataGrid filtering mechanism and the DateTime handling to ensure compatibility with various database configurations.

enchev:
You can specify Kind like this:
```
<RadzenDataGridColumn TItem="PGServerTestDB.Models.TestDB.MyTable" Property="Time" Title="Time">
    <FilterValueTemplate>
        <RadzenDatePicker Kind="DateTimeKind.Utc" DateFormat="MM/dd/yyyy" style="display: block; width: 100%" @bind-Value="@(context.FilterValue)" Name="Time" />
    </FilterValueTemplate>
</RadzenDataGridColumn>
```

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0c3d683f-e4b3-411b-a2b4-ff5b9817650c" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b828c2a1-cf19-4e64-bfc0-df1c9523579e" />

[...]

PR #2317, opened 2025-10-16
set filter datetimes according to specified DateTimeKind

bllkrkmz:
This pull request ensures that DateTime/DateTimeOffset values used for RadzenDataGrid column filters are set according to the specified DateTimeKind before being sent to LoadData

https://github.com/radzenhq/radzen-blazor/issues/2316

### DIFF
--- a/Radzen.Blazor/RadzenDataFilterProperty.cs
+++ b/Radzen.Blazor/RadzenDataFilterProperty.cs
@@ -150,12 +150,6 @@ namespace Radzen.Blazor
         [Parameter]
         public Type Type { get; set; }
 
-        /// <summary>
-        /// Gets or sets the kind of DateTime bind to control
-        /// </summary>
-        [Parameter]
-        public DateTimeKind Kind { get; set; } = DateTimeKind.Unspecified;
-
         Func<TItem, object> propertyValueGetter;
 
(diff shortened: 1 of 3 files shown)


enchev:
Myabe you can get the DateTime Kind from the value instead declaring new parameter:
```
DateTimeOffset? offset = DateTime.SpecifyKind((DateTime)value, ((DateTime?)value).Value.Kind);
```

bllkrkmz:
> Myabe you can get the DateTime Kind from the value instead declaring new parameter:
> 
> ```
> DateTimeOffset? offset = DateTime.SpecifyKind((DateTime)value, ((DateTime?)value).Value.Kind);
> ```

There shouldn't be any issues with the approach you suggested, and it would work as well. However, since the DatePicker component also has a Kind parameter, I thought adding the parameter here would be more appropriate for project consistency. Additionally, developers might want to explicitly specify a particular DateTimeKind.

I'm happy to make the change if you prefer just let me know.

enchev:
In my opinion it will be better to get Kind from the value. Here is how to use DatePicker with specific Kind in DataGrid filtering: https://github.com/radzenhq/radzen-blazor/issues/2315#issuecomment-3414390847

PR #2321, opened 2025-10-21
Added ability to show group footers on every group level. Added a flag in the GroupDescriptor to control footer visibility.

PetarSlavov:
Currently RadzenDataGrid renders group footers on for the inner most group. It is also possible to have a grand total for all data in the grid. In the example below, data is grouped by Country and then by City. Average age is aggregated in the footer via GroupFooterTemplate and in the grand total via FooterTemplate. This display the average age by city and in total, but there is no way to see it aggregated by country.
image

The goal of this pull request is to address this issue without changing current DataGrid behavior. The result of the new grouping setup looks like the example below, with the aggregates per country also displayed.

image

This behavior is configured by the ShowFooter property of the GroupDescriptor, if it's not specified. only group footers for the inner most group will be displayed, as it was before. 

For the example above the following groups were applied:
`args.Grid.Groups.Add(new GroupDescriptor() { Property = "Country", SortOrder = SortOrder.Descending, ShowFooter = true });
args.Grid.Groups.Add(new GroupDescriptor() { Property = "City", SortOrder = SortOrder.Descending });` 

With this new capability, it will be also possible to render group footers only for the country group, but not for the city.
`            args.Grid.Groups.Add(new GroupDescriptor() { Property = "Country", SortOrder = SortOrder.Descending, ShowFooter = true });
            args.Grid.Groups.Add(new GroupDescriptor() { Property = "City", SortOrder = SortOrder.Descending, ShowFooter = false });`

image

ShowFooter property nullable property and is null by default. It is interpreted as true, for the inner-most group and false for all others.

### DIFF
--- a/Radzen.Blazor/Common.cs
+++ b/Radzen.Blazor/Common.cs
@@ -2623,6 +2623,12 @@ namespace Radzen
         public string FormatString { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to show the footer for the group.
+        /// </summary>
+        /// <value><c>true</c> if the footer should be shown; otherwise, <c>false</c>.</value>
+        public bool? ShowFooter { get; set; }
+
+        /// <summary>
         /// Gets the title of the group.
         /// </summary>
         /// <returns>System.String.</returns>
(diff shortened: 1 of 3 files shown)


enchev:
Thanks for the pull request @PetarSlavov! Please keep the original formatting for the unchanged code in order to check easily what's changed.

PR #2332, opened 2025-10-30
Fix for Dialog autofocus for HTMLEditors wrapped in RadzenFormField

MVSAlex:
If a Dialog had an HTMLEditor which was wrapped inside of a RadzenFormField it would always get focused no matter where it is breaking the Autofocus of the first Element.

This was the razor code causing this issue and it would only focus on the HTMLEditor:
```
<RadzenFormField Text="TextBox">
  <RadzenTextBox />
</RadzenFormField>
<RadzenFormField Text="Html">
  <RadzenHtmlEditor />
</RadzenFormField>
```

image
image

This change should fix this and keep the functionality as intended


MVSAlex:
I missed a bug that would completely ignore the TabIndex being -1
Here is the new changed pull request [#2334](https://github.com/radzenhq/radzen-blazor/pull/2334)

Issue #2328, opened 2025-10-29
link to section replaces url with wrong id when section above is not visible anymore


Anspitzen:
**Describe the bug**
When going to for example [Layout - Breakpoints](https://blazor.radzen.com/breakpoints?theme=material3#breakpoints) it jumps correctly to the section breakpoints. But the url points to the next section, Usage, after finishing automatic scrolling. So on reload you end up on wrong section. Correct url is only shown when scrolling up so text from above is visible (previous section content)

**To Reproduce**
1. Go to [page](https://blazor.radzen.com/breakpoints?theme=material3#breakpoints)
2. F5 / reload
5. Scroll up and see correct url getting inserted when full block + somehting on top is visible
6. Click on the Link icon next to the section headline and see scrolling to align top of viewprot with headline and then again wrong url

**Expected behavior**
URL does not get replaced until the block is really going out of view

**Desktop (please complete the following information):**
 - OS: windows
 - Browser chrome
 - Version latest

**Additional context**
Does also happen on other points with sections to jump to, not only the breakpoints

akorchev:
Indeed this is the current behaviour when more than one section is determined to be visible at a time. We are open to suggestions for improvement. The relevant code is here: https://github.com/radzenhq/radzen-blazor/blob/b98dffda8fd559a4e87b70ecd0d0acc6da2649cb/Radzen.Blazor/wwwroot/Radzen.Blazor.js#L2577-L2607

Anspitzen:
So current behaviour: 
Section in Url is the lowest element.top above (vertical) center of viewport
One easy way to do would be moving that "center" line up, like lowest headline visible above the top quarter of viewport

One other way would be chcking if the previous element is still fully visible on page and then not taking any the new url/match closer to center.
Take the top most still visible rect.top (which is above center) as section:
```
const contTop = container.tagName === 'HTML' ? 0 : container.getBoundingClientRect().top;
const center = contTop + container.clientHeight / 2; 
  
 let min = Number.MAX_SAFE_INTEGER; 
 let match; 
  
 for (let i = 0; i < elements.length; i++) { 
   const element = elements[i]; 
   if (!element) continue; 
  
   const rect = element.getBoundingClientRect(); 
  
   if (!match && rect.top < center) { 
     match = selectors[i]; 
     min = rect.top; 
     continue; 
   } 
  
   if (match && rect.top >= center) continue; 
   
[...]

PR #2340, opened 2025-11-04
Topmost visible section instead of closest above center

Anspitzen:
Referencing https://github.com/radzenhq/radzen-blazor/issues/2328, take the top most visible section as url reference instead of the clostest element.top to and above center line of app container

### DIFF
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -2574,37 +2574,37 @@ window.Radzen = {
 
       this.unregisterScrollListener(element);
       element.scrollHandler = () => {
-        const center = (container.tagName === 'HTML' ? 0 : container.getBoundingClientRect().top) + container.clientHeight / 2;
-
-        let min = Number.MAX_SAFE_INTEGER;
-        let match;
-
-        for (let i = 0; i < elements.length; i++) {
-          const element = elements[i];
-          if (!element) continue;
-
-          const rect = element.getBoundingClientRect();
-          const diff = Math.abs(rect.top - center);
-
-          if (!match && rect.top < center) {
-            match = selectors[i];
-            min = diff;
-            continue;
-          }
-
-          if (match && rect.top >= center) continue;
-
-          if (diff < min) {
-            match = selectors[i];
-            min = diff;
-          }
-        }
-
-        if (match !== currentSelector) {
-          currentSelector = match;
-          this.navigateTo(currentSelector, false);
-          ref.invokeMethodAsync('ScrollIntoView', currentSelector);
-        }
+        const contTop = container.tagName === 'HTML' ? 0 : container.getBoundingClientRect().top;
+	    const center = contTop + container.clientHeight / 2; 
+	  
+	    let min = Number.MAX_SAFE_INTEGER; 
+	    let match; 
+	  
+	    for (let i = 0; i < elements.length; i++) { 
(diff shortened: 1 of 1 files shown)


akorchev:
Hi @Anspitzen,

Thank you for the pull request. For some reason I still see similar behavior as before - reloading the page scrolls to the wrong section. Here is what I see breakpoints

akorchev:
I did further investigation and it seems that most of the problem is the initial scroll - when the browser itself scrolls to the current anchor. Then our code runs and may detect a different section. Thus I propose this change which simply ignores the first scroll provided the conditions are met:

```diff
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -2573,7 +2573,19 @@ window.Radzen = {
       const elements = selectors.map(document.querySelector, document);
 
       this.unregisterScrollListener(element);
+      let suppressInitialScroll = false;
+
+      if (selectors.indexOf(location.hash) !== -1) {
+        currentSelector = location.hash;
+        suppressInitialScroll = true;
+      }
+
       element.scrollHandler = () => {
+        if (suppressInitialScroll) {
+          suppressInitialScroll = false;
+          ref.invokeMethodAsync('ScrollIntoView', currentSelector);
+          return;
+        }
         const center = (container.tagName === 'HTML' ? 0 : container.getBoundingClientRect().top) + container.clientHeight / 2;
```

You can test the changes in the [suppress-initial-scroll](https://github.com/radzenhq/radzen-blazor/tree/suppress-initial-scroll) branch.

akorchev:
Hi @Anspitzen,

Could you please test my change in the linked branch? It seems to work fine but I want to be sure your case works too.

Anspitzen:
I did check it, and it still has some unwanted behaviour. Still thinking on how to possible solve that here.
Scrolling does not happen only on initial scroll, another programmical triggered scroll is also when copying a link for a section. That is also scrolling the section to the top, which will trigger scrolling detection and may replace section again. Which is equally confusing as you have copyed a link to url, it scrolls to that url and then shows another section

Anspitzen:
Another trigger would be when you have some scrolling in a component (like datagrid)
Example for demonstration purposes: https://blazor.radzen.com/example-upload?theme=standard&rtl=false#url
Scroll to the point that the "Upload multiple files" Headline comes into the URL (#multiple as url parameter)
Then just select two pictures to upload in the "Upload files to a server" part. The EventConsole will grow as it shows 5 lines. This growing in turn shoves the Multiple part below the halfline, so it should ge off the url and the #url-part should be the active on scrolling. The growing itself will not trigger it, only a scroll.
Now the point: The EventConsole now has 6 entries, so it allows scrolling inside the component, no scroll on the body itself!  This scrolling will trigger the script also and url will update.

I do confess, this handling is pretty unlikely to happen in the wild, but I did it to check when the scrolling scrippt will be triggered.
Possible handling for it to happen would be jumping to the url-section, having a scroll-component and scrolling inside the grid you jumped to. When another section is showing above the middle of viewport the scrolling inside a component will update url even if you never moved the headlines

akorchev:
Those observations are valid but I don't have a solution for them. Sometimes just more than one section is "visible" so there is no clear way to determine what to highlight. Not to mention the case when all items are visible (sections are too short).

Perhaps we should just add an option to disable active element tracking on scroll.

akorchev:
I am closing this PR since the proposed solution doesn't seem to address the original problem completely.

PR #2362, opened 2025-11-20
Textbox/TextArea : add debounce parameter to make Immediate easier on the UI Thread

joriverm:
We noticed that when using immediate if you would type faster than it takes to process the input & re-render it will start acting weird. some text is removed, input shows up delayed, drops some letters being typed etc etc.

Sadly this only happens if the UI is very busy/has a lot of controls etc etc so its very hard to reproduce -_-
I used the existing debounce classes to delay the processing of the input, but only if the delay is given which is 0 by default so it behaves as before.

### DIFF
--- a/Radzen.Blazor/RadzenTextArea.razor
+++ b/Radzen.Blazor/RadzenTextArea.razor
@@ -5,12 +5,12 @@
 {
     if (Immediate)
     {
-    <textarea @ref="@Element" id="@GetId()" disabled="@Disabled" readonly="@ReadOnly" name="@Name" rows="@Rows" cols="@Cols" style="@Style" @attributes="Attributes" class="@GetCssClass()"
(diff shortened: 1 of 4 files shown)


enchev:
Thanks @joriverm however in our opinion this change is too radical - maybe you should avoid Immediate in Blazor server apps and use it only in WASM.

joriverm:
can you elaborate? we figured that with mudblazor and fluentui having the feature it wouldn't be consider too big of a change or radical

EDIT: we also kinda need immediate since we need to process codes and replace them with actual text as the user is typing (+/- a delay because of debounce if we can to relief the UI thread).
for example typing '!tp502 ' would replace it with 'the tertiary vein is blocked and needs surgery to relief pain in the body'. I have yet to try wasm though

enchev:
Why not use @oninput event directly? We merged your Immediate changes however it seems they lead to even more updates needed.

joriverm:
thankfully we have a wrapper component around the radzen textbox that allows us to implement on input on all radzen textbox's ourselves.
but yes, it does seem to work once we disable immediate and add the oninput/debounce on the wrapper div, thanks! :)

feel free to close the PR if you think this will not be merged

PR #2364, opened 2025-11-24
Set default MinWidth to 80px to prevent columns from disappearin…

s7788:
## issue
When resizing columns to minimum width, they could shrink to zero and 
become completely hidden with no way to restore them. 

## fix 
Setting a default MinWidth of 80px ensures columns remain visible and interactive while 
still allowing user customization.


enchev:
Thanks for the pull request @s7788  however this is a breaking change and we cannot accept it.

s7788:
> Thanks for the pull request @s7788 however this is a breaking change and we cannot accept it.

Hello @enchev  , could you please tell me the reason? Thank you.

enchev:
I already posted what's the reason - it's a breaking change.

Issue #2369, opened 2025-11-26
Feature Request: improve radzenTocItem selection when scrolling in radzenToc


tharreck:
**Is your feature request related to a problem? Please describe.**
The toc is always looking at the center of the container to decide which toc item should be selected. In my opinion it's more intuitive when you look at the top or bottom of the container depending on scroll direction.
The same behaviour can be seen in [bootstrap scrollspy component](https://getbootstrap.com/docs/5.0/components/scrollspy/) and wikipedia.
The current implementation does not allow to select the first section when it is small as it can't be set to the center.

**Describe the solution you'd like**
When you scroll down/up the TOC could respectively look at the bottom/top of the container to decide which TOC Item to select instead of always looking at the center of the container. 
This behaviour looks more intuitive for the users and allows small sections in the start or end to be selected.

**Additional context**
I will add a pull request. see #2370

akorchev:
Hi @tharreck ,

I am not sure this would solve the problem as there could be more than one item visible at a time. Always selecting the first one could still be problematic. Especially so if the user clicks to a link within the same page which leads to a scroll.

See my comments here: https://github.com/radzenhq/radzen-blazor/issues/2328 and here: https://github.com/radzenhq/radzen-blazor/pull/2340

Still you can open a pull request with your suggested implementation and we will review if it improves the behavior without breaking the existing one.

PR #2370, opened 2025-11-26
Correctly select RadzenTocItem based on scroll direction in RadzenToc

tharreck:
This allows the RadzenToc component to look at the bottom/top depending on the scroll direction to decide which radzenTocItem to select.

Implements #2369 
I will have a look at the issues in #2328 and #2340

### DIFF
--- a/Radzen.Blazor/wwwroot/Radzen.Blazor.js
+++ b/Radzen.Blazor/wwwroot/Radzen.Blazor.js
@@ -21,6 +21,7 @@ var rejectCallbacks = [];
 var radzenRecognition;
 
 window.Radzen = {
+    selectedNavigationSelector: undefined,
     isRTL: function (el) {
         return el && getComputedStyle(el).direction == 'rtl';
     },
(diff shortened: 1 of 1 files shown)


akorchev:
Just tested your suggestion and it too seems to fail. Try it with the `/breakpoints` page from our online demos. Here is what I see:

scroll-problem

When I scroll to the first section `/breakpoints#breakpoints` and reload the browser it scrolls to the second section `breakpoints#usage`.

tharreck:
hi @akorchev ,

The code has been updated with changes to address the reload page problem you mentioned, Also the global variable has been removed.
In my tests it looks like all issues are solved. Could you please check if I am missing something?

akorchev:
This case indeed works (in the breakpoints page). But I discovered another page that doesn't work: https://localhost:5001/appearance-toggle. Here is a screen recording of the problem I found - clicking an item selects a different item.
scroll-problem

akorchev:
I have tested the reported issues and they both seem to be addressed. Thanks!

Issue #2384, opened 2025-12-09
Feature: Add Immediatie to radzenNumeric


tharreck:
**Describe the solution you'd like**
Add the immediate flag to the radzenNumeric control as seen in the RadzenTextBox

**Additional context**
Pull request incoming

Issue #2352, opened 2025-11-11
Add Immediate="true" in RadzenMask


Andrey2025Ya:
Please,

Add  **Immediate** property   in **RadzenMask**
as in **RadzenTextArea**

Example: **Immediate="true"**
Gets or sets a value indicating whether the component should update the value immediately when the user types. Set to false by default.

Thanks.

joriverm:
i think, in general, this is a wanted feature for many of the input fields ( numeric, datetime, ... )

PR #2386, opened 2025-12-10
Feature: add immediate in numeric and mask

tharreck:
Add Immediate flag into RadzenMask and RadzenNumeric
Created example in the Numeric page

Implements #2384 and #2352

### DIFF
--- a/Radzen.Blazor/RadzenMask.razor
+++ b/Radzen.Blazor/RadzenMask.razor
@@ -4,16 +4,7 @@
 @inherits FormComponentWithAutoComplete<string>
 @if (Visible)
 {
-    @if(Immediate)
-    {
     <input @ref="@Element" disabled="@Disabled" readonly="@ReadOnly" name="@Name" style="@Style" @attributes="Attributes" class="@GetCssClass()" tabindex="@(Disabled ? "-1" : $"{TabIndex}")"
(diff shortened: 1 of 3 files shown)


enchev:
Thanks for the pull request however in my opinion this change will brake the Mask component completely when Immediate="true":
image

image

tharreck:
The problem in the initial commit was that the value did not get updated on the blazor side when immediate was set to true. 
In the new commit the mask is always applied.  This is done by firing the change event when immediate = true.

enchev:
This is what I see using your latest changes:
image

image

tharreck:
hi @enchev,

When testing the immediate on the numeric control on your commit, i noticed that when having a min value of 10 and starting with typing a 1, the minmax validation changes the 1 to 10.  the example below show this behavior.  The code below is used in the gif.

RadzenNumeric

`
<div class="rz-p-12 rz-text-align-center">
    <RadzenNumeric @bind-Value=@value TValue="int?" Min="10" Max="100" Change=@(value => OnChange(value)) Immediate="true" InputAttributes="@(new Dictionary<string, object>() { { "aria-label", "enter value" } })" />
    <EventConsole @ref="console" />
</div>

@code {
    int? value;
    EventConsole console;

    void OnChange(int? value)
    {
        console.Log($"value changed to {value?.ToString() ?? "null"}");
    }
}
`

enchev:
Yes, this is the expected behavior - the same will happen with Immediate="false" however when onchange is raised on lost focus.

Issue #2381, opened 2025-12-08
Handle Drop Event in HTML Editor


Anspitzen:
**Is your feature request related to a problem? Please describe.**
There is a Paste Eventhandler for the [HtmlEditor](https://blazor.radzen.com/html-editor), but not for the Drop Event

**Describe the solution you'd like**
Another EventListener in the javascript, same as the paste event listener. 
(Paste has the same implemented DataTransfer object, it is called clipboardData, on drop event it would be called dataTransfer. Same object, so function could be even reused)

**Describe alternatives you've considered**
- Implementing it myself on the drop event, but would need to get instance reference of Editor, insert own javascript (copy the current paste function) and invoking it from js. 
- Making a full blazor binding onto the event and also copy paste function, doesn't need reference but have to copy and adapt JS function into C# code

akorchev:
Hi @Anspitzen,

What is the expected API for the Drop event handler? What do you expect to happen? What is your actual requirement - what kind of stuff do you drop in RadzenHtmlEditor?

Anspitzen:
The Drop API is basically the same like the Paste API or drag. Same structured data transfer object (literally called DataTransfer), bound to ondrop.
So the js function at the [Paste Listener](https://github.com/radzenhq/radzen-blazor/blob/edd85f8ec0beb669635f0fc8ac1e8aa7b050c8ef/Radzen.Blazor/wwwroot/Radzen.Blazor.js#L2051) can be reused,
only every e.clipboardData has to be exchanged with e.dataTransfer.

Usecase: When you mark stuff somewhere (like a document, or files in explorer) and drag&drop them into the Editor, instead of hitting CTRL+C/CTRL+V or context menu copy&paste. 
So same procedure as dropping files into the Upload Component for example. Just also for text.

API documentation on mozilla has some mentions about maybe having to cancle event ondragover to keep drop event consistently. Haven't looked into it too deeply.

akorchev:
> The Drop API is basically the same like the Paste API or drag

The existing Paste event provides the pasted value as a string. Doesn't seem the same as a drop API. 

I am still not sure how this API is supposed to be used. Please provide some sample code that shows what you plan to do.

Anspitzen:
```
ref.dropListener = function (e) {
      var item = e.dataTransfer.items[0];

      if (item.kind == 'file') {
        e.preventDefault();
        var file = item.getAsFile();

        if (uploadUrl) {
[...]

PR #2394, opened 2025-12-16
Rework paste handler and add same function to new drop handler

Anspitzen:
Referencing https://github.com/radzenhq/radzen-blazor/issues/2381 

Reworked paste to accept multiple files and not just pick first of copied files, without user knowing which file it is.
Added Drop handler with the same functionality as the paste handler

Real changes in the Radzen.Blazor.js file are between lines 2051 and 2151. 
The many other changes on the js file are result of pretty print from VS 2026, so only formated whitespaces.

### DIFF
--- a/Radzen.Blazor/RadzenHtmlEditor.razor.cs
+++ b/Radzen.Blazor/RadzenHtmlEditor.razor.cs
@@ -110,24 +110,6 @@ namespace Radzen.Blazor
         public EventCallback<HtmlEditorPasteEventArgs> Paste { get; set; }
 
         /// <summary>
-        /// A callback that will be invoked when the user drops content in the editor. Commonly used to filter unwanted HTML.
-        /// </summary>
-        /// <example>
-        /// <code>
-        /// &lt;RadzenHtmlEditor @bind-Value=@html Drop=@OnDrop /&gt;
-        /// @code {
-        ///   string html = "@lt;strong&gt;Hello&lt;/strong&gt; world!";
-        ///   void OnDrop(HtmlEditorPasteEventArgs args)
-        ///   {
-        ///     // Set args.Html to filter unwanted tags.
-        ///     args.Html = args.Html.Replace("&lt;br&gt;", "");
-        ///   }
-        /// </code>
-        /// </example>
-        [Parameter]
-        public EventCallback<HtmlEditorPasteEventArgs> Drop { get; set; }
-
-        /// <summary>
         /// A callback that will be invoked when there is an error during upload.
         /// </summary>
         [Parameter]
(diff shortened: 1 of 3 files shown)


akorchev:
Hi @Anspitzen,

Please revert the original formatting of Radzen.Blazor.js. I cannot review the change when the whole file is changed. 

Also use the existing Paste event - there is no need to add a new Drop event with the same arguments and behavior as Paste.

Anspitzen:
Seems like it worked with copying the foramting from before over to this. Even if Visual Studio insists it should be more indented.

As for using a new function, I don't do that. I literally use the same functions for paste and drop. Only differences are literally the property name for the dataTransfer Object from the events, the hasDelegate flag and the function name which I have to invoke in Blazor.

One could argue that the HtmlEditorPasteEventArgs could be renamed, but I left them for now and just use them the same way on drop-paste

akorchev:
I mean you should delete the Drop event. It is the same as the Paste event - even uses the same event args.

Anspitzen:
No, it is not the same event. Drop is a different event. 
When dropping into the textarea it will be shown like a paste. But if I drop files from my explorer it is indeed a different event and not a paste. When droping a picture (or several) into the area it will open them as files in the browser currently. When I mark them, CTRL+C and then paste them into the area it will paste them and not open them

akorchev:
It seems to be the same event - even uses the paste event args and provides the same HTML string as argument.

Anspitzen:
Yes, I break it down to the same dataTransfer objects on js side before hadeling them the same either way.
But Paste is a ClipboardEvent, containting a list of dataTransfer objects
Drop is a Drag/Mouse Event, not a clipboardEvent. But also containing a list of dataTransfer objects.

I simply extract these dataTransfer objects from both events and from there on I can handle them the same way. But they are coming from different event objects

akorchev:
I understand the implementation but I don't like adding a new Drop event which is exactly the same as the existing Paste event. There is no reason for a separate event to exist when just raising the existing Paste event would suffice.

The actual implementation does the same - it reads data and inserts it in the editor.

Anspitzen:
HtmlEditorPasteDifferenceFromDrop

Anspitzen:
Or do you mean on Blazor side? 
I mean of course I could change it to fixed OnPaste callback instead of splitting it by drop and paste. Just cleaner to have it seperate as mostly files will get dropped and not pasted, but text will get pasted and seldom droped. So thats just preference

akorchev:
I don't understand what this gif is supposed to mean.

Let me try one more time.

1. The implementation is ok - it adds the ability to insert content by dragging and dropping it. I am OK with that.
2. Creating a new Drop EventCallback parameter is redundant as it does the same as the existing Paste EventCallback parameter.

What I am requesting is that:

1. You delete the Drop event callback property from RadzenHtmlEditor.razor.cs - we should use the existing Paste for both cases.
2. Trigger Paste via JS interop in both cases - paste and drop.

Hope this clears things up. If not I will do it myself tomorrow.

Anspitzen:
Ah, ok, that I did misunderstand. I was event = javascript, and did not consider the c# EventCallback as an event itself
Will change it to same OnPaste c# function and Callback

PR #2397, opened 2025-12-17
Radzen Examples Split View

paulo-rico:
Hi guys

Something that's niggled me a little when using the example site. I do like to scroll through the code of an example while being able to refer to the rendered output. Not too easy when switching tabs. Also, you may scroll to some code, and then want to view how that affects the output. When switching back, you lose your position in the code window.

This PR adds a split view to the `RadzenExample` component. Not only can you review the code while viewing the output, but if you make a change and run your code, the output is updated and your code position remains.

I find it a much more pleasant experience.

What do you think?

BTW, if you do like the idea, there are some CSS issues that I did not know how to deal with. You will notice that the top of the `CodeViewer` component is trimmed. Not sure why placing it in a `RadzenSplitterPane` should do that.

Also, I assume that the `CodeViewer` or the `Monaco` component must have a fixed height, and cannot be dynamic?

Regards

Paul

### DIFF
--- a/RadzenBlazorDemos/Shared/RadzenExample.razor
+++ b/RadzenBlazorDemos/Shared/RadzenExample.razor
@@ -45,6 +45,18 @@
                     <CodeViewer ComponentName="@ComponentName" @bind-Value="@source" ComponentSource="@Source" PageName="@ExampleUrl" Compiled="OnCompiled" />
                 </div>
             </RadzenTabsItem>
+            <RadzenTabsItem Text="Split View" Title="Split view between example and source code">
+                <RadzenSplitter>
+                    <RadzenSplitterPane Size="50%">
+                        @(dynamicContent ?? ChildContent)
+                    </RadzenSplitterPane>
+                    <RadzenSplitterPane>
+                        <div class="source-tab-item rz-pb-2">
+                            <CodeViewer ComponentName="@ComponentName" @bind-Value="@source" ComponentSource="@Source" PageName="@ExampleUrl" Compiled="OnCompiled" />
+                        </div>
+                    </RadzenSplitterPane>
+                </RadzenSplitter>
+            </RadzenTabsItem>
             @foreach (var p in AdditionalSourceCodePages)
             {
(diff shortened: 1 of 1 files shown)


akorchev:
Hi @paulo-rico,

We prefer not to introduce another tab - there are demos with a lot of tabs already. We will however accept a solution for the editor scrollbar position resetting. Maybe changing the RenderMode of the Tabs to client would deal that. Or storing and restoring the scroll position after tab changes.

paulo-rico:
Hi @akorchev

Understand about the tabs issue. From a personal point of view, I would still like to see the code and output side-by-side.

May I put forward another suggestion.

What do you think about a separate component, something like `RadzenPlayground`, that would basically be the 'Split' in this PR, in it's own page.

It could be used, thus -

- Have it's own menu / toolbar item (some way of getting to it)
- There could be a small link icon in `RadzenExample` that would open this page, passing the code from the example.
- It could be the copy / paste 'go to' link for forum code suggestions.

Just throwing ideas out. Let me know what you think.

Regards and stay well.

Paul

akorchev:
Hi @paulo-rico,

This should ideally be a new page instead of a component (it would need a URL if we want to link to it). Later we could also add some sort of persistence (in maybe S3 compatible blob storage) so people could share their snippets.

akorchev:
We decided to do that ourselves - there will be a new `/playground` page with ability to load an existing demo and also save the modifications. Will do that soon-ish.

akorchev:
The playground is now live! You can open any demo in it:

1. Go to the **Edit Source** tab.
2. There is a new button "Open in playground" 
image

3. You can save snippets and share them.
image

paulo-rico:
Excellent!!! (only just seen this message)

To be honest, I've had a play around with it, and I wasn't really sure which path to take with regards passing the code, e.t.c.
Much better you guys to do it.

Cheers Atanas

EDIT - Only saw the 'do it ourselves post' before I posted. It then refreshed showing that you'd actually done it and put it live. Looks fantastic!

Issue #2406, opened 2026-01-02
Allow to define Http method for Upload component


bykka:
Hi,

The upload component uses the POST method to upload files to a URL. This does not allow the component to be used for uploading files to an S3 PreSigned URL, as the latter requires the PUT method.

It would be useful to introduce an optional new parameter called 'Method' with a default value of 'POST'.

The method is hardcoded in the file https://github.com/radzenhq/radzen-blazor/blob/master/Radzen.Blazor/wwwroot/Radzen.Blazor.js line 923.

PR #2407, opened 2026-01-02
#2406 - Allow to define Http method for Upload component

bykka:
A new property Method with a default value of 'POST' has been introduced for the component Upload.

### DIFF
--- a/RadzenBlazorDemos/Pages/UploadUrl.razor
+++ b/RadzenBlazorDemos/Pages/UploadUrl.razor
@@ -1,8 +1,17 @@
 ﻿<RadzenCard Variant="Variant.Outlined">
     <RadzenUpload 
+        ChooseText="Upload a single file as form data"
         Url="upload/single" 
+        Style="width: 100%"
+        Progress=@OnProgress 
+        Complete=@OnComplete
+        InputAttributes="@(new Dictionary<string,object>{ { "aria-label", "select file" }})" />
+    <RadzenUpload
+        ChooseText="Upload a single file as stream"
+        Url="upload/stream" 
         Method="PUT"
         Stream="true"
+        Multiple="false"
         Style="width: 100%"
         Progress=@OnProgress 
         Complete=@OnComplete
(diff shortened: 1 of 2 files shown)


RadzenBlazorDemos/Pages/UploadUrl.razor
```diff
@@ -1,5 +1,9 @@
 ﻿<RadzenCard Variant="Variant.Outlined">
-    <RadzenUpload Url="upload/single" Style="width: 100%" Progress=@OnProgress 
+    <RadzenUpload 
+        Url="https://localhost:54380/upload/temp?X-Amz-Expires=300&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minioadmin%2F20260102%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260102T090111Z&X-Amz-SignedHeaders=host&X-Amz-Signature=b9682b0d3f7948dce9639fca039b346c1722f17ae2b978ddb56cb47793ee328b"
```
enchev: Thanks for the pull request! I don't think that this change is needed.

bykka: That is true, but the POST version is more difficult to use because additional fields need to be injected. In the case of URL uploading, the Upload component always creates a new blank FormData instance and includes only file fields. 

https://docs.aws.amazon.com/sdkfornet/v4/apidocs/items/S3/TCreatePresignedPostResponse.html

And BTW,  the POST version will handle one file only as well.

So, what do you think about adding an extra parameter called 'Stream' with a default value of 'false'?
If the "Stream" parameter is set to "true", then XHR will send the selected file itself instead of the FormData object.

enchev:
Thanks for the updates! Make sure all the demos are working and there is no additional obsolete code from your tests.

enchev:
Please rebase your branch against the master branch since there are important upload fixed. Tried also the Stream=true demo after local rebase and didn't worked.

PR #2410, opened 2026-01-08
fix strange oninput behavior on textbox and textarea (server rendering)

tharreck:
this pr fixes an issue when using onInput  and server rendering  on a RadzenTextArea / RadzenTextBox  that input is lost in some cases. 

I noticed something similar was done for the RadzenNumeric. see commit [a0cfc5d2](https://github.com/radzenhq/radzen-blazor/commit/a0cfc5d26736c5d7f7397356818ceef6059ba152)

### DIFF
--- a/Radzen.Blazor/RadzenTextArea.razor
+++ b/Radzen.Blazor/RadzenTextArea.razor
@@ -3,27 +3,30 @@
 @inherits FormComponent<string>
 @if (Visible)
 {
+    if (Immediate)
+    {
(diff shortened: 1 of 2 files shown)


enchev:
Thanks for the pull request! Maybe the fix can be applied only for the Immediate mode - i.e. the preprocessor should be inside the `if (Immediate)` only. In this case the rest of the code will remain unchanged.

tharreck:
I have updated the PR so that only the Immediate mode is updated.

PR #2458, opened 2026-03-01
Legend Mouse Hover for Pie and Donut Charts

paulo-rico:
As per [forum request](https://forum.radzen.com/t/radzenchart-tooltips/21411) this PR implements Pie and Donut tooltips for legend items activated by `mouseover` and `mouseout` events.

Regards

Paul

### DIFF
--- a/Radzen.Blazor/RadzenPieSeries.razor.cs
+++ b/Radzen.Blazor/RadzenPieSeries.razor.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using static System.Runtime.InteropServices.JavaScript.JSType;
 
 namespace Radzen.Blazor
 {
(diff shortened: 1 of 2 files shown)


akorchev:
Hi @paulo-rico,

We prefer the implementation suggested in the linked thread.

paulo-rico:
Hi @akorchev

Not sure if it's slipped your mind as it's not in any examples (my bad and I should address it), but I had created a `LegendClick` event as per [this PR](https://github.com/radzenhq/radzen-blazor/pull/499) for the same reason that the forum user needed the Legend Tooltips (very small slices).

I don't know if that results in a confusing API to have a 'click' assignable in multiple ways.

What are your thoughts?

EDIT - I was using `LegendClick` to display detailed records in a `DataGrid`. I know I would find it frustrating to require a tooltip, but have the DataGrid refresh to another data set each time I clicked.

akorchev:
We have an implementation ready and just wait for the customer to confirm before pushing it. It is using the LegendClick event and shows the tooltip on clicking a legend item. I hit problems with having mouseenter/mouseleave (you were using mouseover/mouseout which are problematic). Also attaching the mouse events and doing nothing in every other series type is a performance overhead. That's the reason I've closed the PR - should have explained myself better for which I apologise.

akorchev:
Your implementation with DataAt allowed me to support hover tooltips as well! Commit was recently pushed.

PR #2484, opened 2026-03-17
SSRSViewerPage Documentation Update

mvoxland:
I ran into an issue with my usage of the controller code suggested in the demo, and thought I would add another note to the documentation about it.

I thought it would be better to put at the bottom rather than alter the existing documentation, since the current advice seems to be working well enough for people.


mvoxland:
Does not completely fix my issues, so would be an incomplete or wrong addition

PR #2497, opened 2026-03-23
Fix CSP inline event handler violations in RadzenProfileMenu

panasico:
### Summary                                                                                                                                                                                                                         
                                                                                                                                                                                                                                      
  `RadzenProfileMenu` renders an inline `onclick="Radzen.toggleMenuItem(this)"` attribute on the toggle element. This violates Content Security Policy (CSP) when `unsafe-inline` is not allowed for scripts, causing the profile menu
   dropdown to silently fail to open.                                                                                                                                                                                                 
                                                                                                                                                                                                                                    
  ### Changes

  - Replaced the inline `onclick` handler with a Blazor `@onclick` event binding (`OnToggleClick`)
  - Added an `ElementReference` (`toggleRef`) to pass the DOM element to `Radzen.toggleMenuItem` via JS interop (`IJSRuntime.InvokeVoidAsync`)
  - No changes to behavior or public API — the fix is internal to the component

  ### Before / After

  | Before | After |
  |---|---|
  | `onclick="Radzen.toggleMenuItem(this)"` | `@onclick="OnToggleClick"` → JS interop call |
  | Blocked by CSP `script-src` without `unsafe-inline` | Works with strict CSP policies |

### DIFF
--- a/Radzen.Blazor/RadzenProfileMenu.razor
+++ b/Radzen.Blazor/RadzenProfileMenu.razor
@@ -6,10 +6,10 @@
         role="menu" aria-label="@ToggleAriaLabel" tabindex="0"
         @onkeydown="@OnKeyPress" @onkeydown:preventDefault=preventKeyPress @onkeydown:stopPropagation="stopKeydownPropagation">
         <li class="rz-navigation-item">
-            <div class="rz-navigation-item-wrapper" role="button" tabindex="0" aria-haspopup="menu"
+            <div @ref="@toggleRef" class="rz-navigation-item-wrapper" role="button" tabindex="0" aria-haspopup="menu"
                  aria-label="@ToggleAriaLabel"
                  aria-expanded="@((!Collapsed).ToString().ToLowerInvariant())" aria-controls="@($"{GetId()}-menu")"
-                 onclick="Radzen.toggleMenuItem(this)" @onkeydown="OnToggleKeyDown">
+                 @onclick="OnToggleClick" @onkeydown="OnToggleKeyDown">
                 <div class="rz-navigation-item-link">
                     <div class="item-text" @onkeydown:stopPropagation="stopGuardKeydownPropagation" @onkeydown="OnGuardKeyDown">
                         @if (Template != null)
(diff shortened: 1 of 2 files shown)


enchev:
Thanks for the pull request @panasico however our menus/dropdowns intentionally are using plain JS handlers to avoid lag problems with server-side Blazor: RadzenAutoComplete, RadzenDataGrid (filter menu), RadzenDropDown, RadzenMask, RadzenMenu, RadzenNumeric, RadzenProfileMenu, RadzenSplitButton, etc.

panasico:
@enchev Thanks for the explanation! Totally understand the reasoning behind plain JS handlers
to avoid SignalR roundtrips - makes sense for keeping interactions snappy.

I'd like to suggest an alternative that keeps the same client-side performance
(no SignalR, no Blazor interop) while also working with nonce-based CSP.

The idea is to replace inline handlers with event delegation via data-attributes
in `Radzen.Blazor.js` — this is the [recommended approach](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/CSP#:~:text=Refactor%20patterns%20disallowed%20by%20the%20strict%20CSP%2C%20such%20as%20inline%20event%20handlers%20and%20eval().%20For%20example%2C%20replace%20inline%20event%20handlers%20with%20addEventListener()%20calls%20inside%20scripts.) when adopting strict CSP.

Instead of:
```html
<div class="rz-navigation-item-wrapper" onclick="Radzen.toggleMenuItem(this)">
```

Markup becomes:
```html
<div class="rz-navigation-item-wrapper" data-radzen-toggle="menu-item">
```

And a single delegated handler in `Radzen.Blazor.js`:
```js
document.addEventListener('click', function(e) {
    var el = e.target.closest('[data-radzen-toggle="menu-item"]');
    if (el) Radzen.toggleMenuItem(el);
});
```

Same pure JS on the client, same speed, but no inline attributes that CSP can block.
A similar approach is used in Bootstrap, React.
I have generated an example to evaluate this change - [csp-test.html](https://github.com/user-attachments/files/26182773/csp-test.html)

This issue has [come up before](https://forum.radzen.com/t/csp-is-throwing-error-while-clicking-on-radzen-profile-menu-component/19121)
and will likely keep appearing as more projects adopt nonce-based CSP.

If this approach works for you, I can update this PR with the fix for `RadzenProfileMenu`,
and the rest of the affected components should be easy to patch the same way.

enchev:
We preffer differnt approach similar to how createDraggable works at the moment. We will do our best to provide this soon!

PR #2533, opened 2026-04-25
HTML Editor with tabels

Frank-NTD:


enchev:
Thanks for the pull request - here are our comments:

## Overview
Adds table support to RadzenHtmlEditor: insert/edit dialog, toolbar buttons, context menu, keyboard nav (Tab), Shift+Click range selection, copy/paste cells, column resizing, and a property panel. ~2,300 LOC across new C# components (RadzenHtmlEditorTable.razor, RadzenHtmlEditorTableCommandButton.razor), DTOs, ~785 lines of JS, themes, demo, and tests.

## Critical issues

- **Missing JS function definitions.** Radzen.Blazor.js calls `getSelectedTableCells`, `copySelectedTableCells`, `pasteSelectedTableCells`, `selectTableCellRange`, and `clearSelectedTableCells` (used by `getSelectedCellBounds`, `execTableCommand` for `copyCells`/`pasteCells`, the keydown Ctrl+C/V handler, the click handler, and the contextmenu handler) but **none of them are defined** anywhere in the diff or in the file. Result: runtime `TypeError` on cell range selection, copy, paste, merge, and on every Shift+click. This is the headline feature of the PR ΓÇö it cannot work as shipped.

- **Table-edit dialog loses border/padding/width-px/color fields.** Radzen.Blazor.js (lines 1900-1933 of the diff) computes `borderTop/Right/Bottom/Left`, `borderStyle`, `borderWidthPx`, `borderColor`, `columnWidthPx`, `cellPaddingPx`, `cellBorder` inside `getTableSelectionFromContext`, but the returned object literal omits all of them. When the user reopens the dialog on an existing table, those fields default to null/false, so submitting the form silently strips the table's styling.

## High-priority concerns

- **Document-level listener leakage.** `mouseup` and `mousemove` are added to `document` per editor instance and stored as `ref.mouseupResizeListener` / `ref.documentMouseMoveResizeListener`. With two editors on the same page, every drag in editor A also fires in editor B. Removal happens in destroy, but if a drag starts and the component unmounts mid-drag, the listeners still hit a removed `ref`.

- **Resize spams interop.** `documentMouseMoveResizeListener` invokes `instance.invokeMethodAsync('OnChange', ref.innerHTML)` on every mousemove during drag. This serializes the entire editor HTML and re-marshals it across the JS<->.NET boundary at ~60 Hz. Fire `OnChange` on `mouseup` only (or debounce).

- **Cell index stored on DOM.** `cell._rzColumnIndex` / `cell._rzRowIndex` mutate DOM nodes and become stale after structural edits unless `getCellMatrix` is re-run. Several call sites mix "use stored index" vs "recompute from `sectionRowIndex`/`indexOf`" ΓÇö accident-prone. Prefer always computing from a fresh matrix.

## Code quality / conventions

- **API surface bloat.** ~40 new `TableXxxText` `[Parameter]`s on `RadzenHtmlEditor` plus matching pass-through properties on `RadzenHtmlEditorTable`. Other tools (e.g. `RadzenHtmlEditorImage`) keep localization local to the tool component. Consider keeping these on `RadzenHtmlEditorTable` only, or grouping them into a single `Strings`/options object ΓÇö the current pattern doubles every new label.

- **Duplicated context lookup.** `getTableContext` and `getTableContextFromTarget` are 95% identical. Extract `buildContext(element)` and call from both.

- **`switch` block is large.** `execTableCommand` is a ~250-line monolith. Splitting into per-command helpers would aid testability and lets you unit-test against jsdom.

- **`<RadzenHtmlEditorTableCommandButton>` repetition in RadzenHtmlEditor.razor.** Nine near-identical lines in the default toolbar. Acceptable, but a `RadzenHtmlEditorTableTools` aggregator component would be cleaner and easier for users to opt in/out of.

- **`Radzen.execTableCommand` mutates `setAttribute('border', border)` with a numeric value** ΓÇö should be string. JS coerces, but it's sloppy in a TS-allergic codebase.

- **`createTableHTML`** uses `document.execCommand('insertHTML', ...)` which is deprecated. Project already relies on it elsewhere, so this is acceptable but worth tracking.

## Test coverage

- Only labels, defaults, command-button-disabled, and one notification path are covered.
- **No tests for the actual table operations** (`insertTable`, `addRowBefore/After`, `mergeCellRight/Down`, `splitCell`, `pasteCells`, `getTableSelection` round-trip). The two critical bugs above would have been caught by a single bUnit + JSInterop integration test that round-tripped the selection object.
- Existing tests now register `ContextMenuService` + `NotificationService` in every method ΓÇö extract a `CreateContext()` helper.

## Security

- `cell.innerHTML += ' ' + nextCell.innerHTML;` (and similar in `mergeCellDown`) preserves raw HTML ΓÇö fine for content already inside the editor, but be aware that `pasteCells` (once implemented) needs to sanitize copied content if it ever crosses editor instances or origins.
- `RadzenHtmlEditor` already trusts its own DOM, so this is consistent with existing behavior.

## Recommendation

**Request changes.** The two critical issues block merge: ship the missing JS helpers and complete the `getTableSelectionFromContext` return object. Then trim the localization API surface, debounce the resize `OnChange`, and add at least one end-to-end interop test per table command before re-review.

enchev:
We’ve decided to reimplement this in our 11 branch—it will be released in the middle of May. I'm closing this.

PR #2550, opened 2026-05-14
New Scheduler Agenda View

paulo-rico:
Here is a PR for a new `RadzenScheduler` view, `RadzenAgendaView`

This simply displays appointments as a list by day(s), similar to the [Syncfusion Scheduler Agenda View](https://blazor.syncfusion.com/demos/scheduler/agenda-view?theme=fluent2)

Regards

Paul

### DIFF
--- a/Radzen.Blazor/IScheduler.cs
+++ b/Radzen.Blazor/IScheduler.cs
@@ -107,8 +107,9 @@ namespace Radzen.Blazor
         /// Renders the appointment.
         /// </summary>
         /// <param name="item">The item.</param>
+        /// <param name="isAgenda">Indicates if the appointment is in agenda view.</param>
         /// <returns>RenderFragment.</returns>
-        RenderFragment RenderAppointment(AppointmentData item);
+        RenderFragment RenderAppointment(AppointmentData item, bool isAgenda = false);
 
         /// <summary>
         /// Notifies the scheduler that the user has moved the mouse over the specified appointment.
--- a/Radzen.Blazor/RadzenAgendaView.razor
+++ b/Radzen.Blazor/RadzenAgendaView.razor
@@ -1,0 +1,17 @@
+@using Radzen.Blazor
+@using Radzen.Blazor.Rendering
+
+@inherits SchedulerViewBase
+
+@code {
+    public override RenderFragment Render()
+    {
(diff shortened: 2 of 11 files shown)


enchev:
Thanks for the contribution! A few things to address before merge:

Keyboard navigation looks broken. In AgendaView.razor the StartTime and EndTime parameters are declared but RadzenAgendaView never passes values, so they default to TimeSpan.Zero. That means AppointmentsInSlot(dayStart, dayEnd) is called with start == end and currentAppointments is always empty. In OnDayClick the same issue makes the SelectDay callback receive an empty list. The up/down branch around line 145 also has a bounds check that compares newDate (CurrentDate +/- 1 day) against CurrentDate''s own day range, so it''s always false and the assignment never runs. End result is arrow keys and Space do nothing in agenda view.

A few smaller things:

- RadzenAgendaView.NumberOfDays XML doc says default is 2 but the code sets it to 1.
- AgendaSlotEvents.razor declares StartDate, EndDate and MinutesPerSlot parameters that aren''t used, and has an unused internal RenderedAppointment class - can be removed.
- The agenda branch in RadzenScheduler.RenderAppointment hard-codes font-weight: bold and font-style: italic as inline styles. Better to move these to the SCSS so they follow the theme.
- The CssClass expression in AgendaSlotEvents.razor re-runs Where/OrderBy/ThenByDescending/ElementAtOrDefault inside the foreach for every appointment. Computing the focused one once outside the loop would be much cheaper.
- About the isAgenda parameter added to IScheduler.RenderAppointment and the Appointment component - IScheduler is public so this is a contract change, and it leaks an agenda-specific concern into shared code. Could the agenda formatting stay inside the agenda view (its own template) instead?
- themes/_css-variables.scss has a lot of unrelated whitespace changes - blank line separators stripped across the whole file. Only the three new --rz-scheduler-header-agenda-* lines belong here. Reverting the rest will keep the diff focused and avoid merge conflicts.

Also a few of the new files are missing a trailing newline (AgendaSlotEvents.razor, AgendaView.razor, SchedulerAgenda.razor).

enchev:
Thanks for the contribution! Merged into the `11` branch as 6fa5a3295 with review feedback addressed (keyboard navigation, removal of `isAgenda` from `IScheduler.RenderAppointment`, agenda styling moved to SCSS, etc.). Closing this PR since the work has been integrated.

PR #2552, opened 2026-05-20
Introduce SortComparer to RadzenDataGridColumn, for custom IComparer sort support

pianomanjh:
In my project I have a number of DataGrid columns whose values are ids (stages, divisions, categories, etc.) that I display via a lookup to a friendly name. I'd like those columns to sort by the displayed name rather than the underlying id.

Today the DataGrid's client-side sort orders by a member path only — the order expression is built from the column's sort property — so a column bound to an id sorts in id order, with no way to supply a custom comparison. `FilterTemplate`/`FilterProperty` already let you customize filtering, but there isn't an equivalent for sorting.

This adds a `SortComparer` parameter to `RadzenDataGridColumn`. When set, the grid orders that column's in-memory data with the supplied `IComparer` instead of the default member-path ordering — e.g. comparing id values by their mapped display text. It composes with multi-column sorting (OrderBy/ThenBy) and honors ascending/descending.

- Entirely opt-in and non-breaking: when `SortComparer` is null the existing sort path is untouched.
- Applies to in-memory data. Server-paged data (LoadData/OData) sorts on the server and is unaffected; self-referencing (tree) data is also excluded. Both are noted in the XML docs.
- The comparer receives the column's sort-property values, so it's easy to back with a small lookup/dictionary (or any custom ordering — natural sort, a fixed enum order, etc.).

I considered instead wrapping the value in an `IComparable` type, but that pushes the wrapper through display/filtering/equality; a per-column comparer keeps the row value (the id) clean and isolates the sort concern to the one column.

Added tests (single and multi-column, ascending/descending, the default no-comparer path, and unmapped values) and a "Custom Sort Comparer" demo under DataGrid → Sorting.

### DIFF
--- a/Radzen.Blazor.Tests/DataGridTests.cs
+++ b/Radzen.Blazor.Tests/DataGridTests.cs
@@ -3593,6 +3593,108 @@ namespace Radzen.Blazor.Tests
 
             Assert.Equal(new[] { "A", "B", "C" }, order);
         }
+
+        [Fact]
+        public void DataGrid_SortComparer_ComposesWithSecondaryColumn()
+        {
+            using var ctx = new TestContext();
+            ctx.JSInterop.Mode = JSRuntimeMode.Loose;
+            ctx.JSInterop.SetupModule("_content/Radzen.Blazor/Radzen.Blazor.js");
+
+            // Two rows share Code=2 (Normal); secondary sort by Reference breaks the tie.
+            var data = new[]
+            {
+                new SortComparerItem { Reference = "B2", Code = 2 },
+                new SortComparerItem { Reference = "A", Code = 1 },
+                new SortComparerItem { Reference = "B1", Code = 2 },
+            };
+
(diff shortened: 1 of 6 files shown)


enchev:
Thanks for the PR! One small nuance worth calling out in the docs: if a user mixes a SortComparer column with a non-comparer column under multi-column sorting, the whole sort goes through OrderByComparers, so every sort key is evaluated in memory not just the comparer column. The current docs only mention LoadData/OData and tree data as exclusions; the "any comparer forces the whole sort in-memory" detail is implicit.

pianomanjh:
Good call — I've added a note to the `SortComparer` docs making the in-memory behavior explicit: when a column with a `SortComparer` takes part in a multi-column sort, the whole sort is evaluated in memory (every column's key, not just the comparer column), in addition to the existing LoadData/OData/tree-data exclusions. Pushed.

PR #2553, opened 2026-05-20
Add FilterData to RadzenDataGridColumn for lookup-driven CheckBoxList filtering

pianomanjh:
In my project I have a number of DataGrid columns whose values are ids (stages, divisions, categories, etc.) that I display via a lookup to a friendly name. I'd like the CheckBoxList filter to show, search, and order those options by name while still filtering on the underlying id.

Today the built-in CheckBoxList filter derives its options from the distinct values in the grid data and renders each option from its raw value (enum description / FormatString / ToString). There's no way to feed it a lookup so it can show a name for an id — so I'd been replacing the whole `FilterTemplate` with a hand-rolled `RadzenListBox`, which loses the native Clear/Apply/virtualization wiring.

This adds `FilterData` + `FilterTextProperty` + `FilterValueProperty` to `RadzenDataGridColumn`, mirroring `RadzenDropDown`'s `Data`/`TextProperty`/`ValueProperty`. When `FilterData` is set, the built-in CheckBoxList binds its `RadzenListBox` to it — displaying and searching by `FilterTextProperty`, selecting by `FilterValueProperty`. The column's `Property` stays the underlying id, so the existing CheckBoxList setup (`FilterPropertyType = IEnumerable<idType>`, operator `Contains`) still applies and selections filter as `id in (…)` — translating to both in-memory and OData/IQueryable sources.

- Entirely opt-in and non-breaking: when `FilterData` is null the existing distinct-values path is unchanged.
- Applies to `FilterMode.CheckBoxList`. Precedence is `FilterValueTemplate` > `FilterData` > native distinct values.
- Independent of the existing `LoadColumnFilterData` event (that drives the native path only); `FilterData` simply takes precedence per column.

Added tests (filter-by-value, option labels rendered from `FilterTextProperty`, and a regression that the native path is unchanged when `FilterData` is null) and a "CheckBoxList with Lookup" demo under DataGrid → Filtering.

### DIFF
--- a/Radzen.Blazor.Tests/DataGridTests.cs
+++ b/Radzen.Blazor.Tests/DataGridTests.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Radzen.Blazor.Tests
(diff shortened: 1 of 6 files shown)


enchev:
Thanks for the PR. A few things I would like to address before merging:

1. Naming. `FilterValueProperty` reads almost identically to `FilterProperty` despite meaning the opposite side of the relation (lookup-side vs grid-side). Could we rename to something like `FilterLookupData` / `FilterLookupTextProperty` / `FilterLookupValueProperty`, or at least make the XML docs explicit that these refer to the lookup source?
2. Duplication in `RadzenDataGridHeaderCell.razor`. The new `else if (Column.FilterData != null)` branch copy-pastes most of the existing `RadzenListBox` block. It would be cleaner to keep a single `RadzenListBox` and switch `Data` / `TextProperty` / `ValueProperty` / `LoadData` / `AllowFiltering` based on whether `FilterData` is set.
3. `AllowFiltering="true"` is hardcoded in the new branch. Fine for typical small lookup data, but a parameter to opt out would be nice for symmetry.
4. The interaction with `LoadColumnFilterData` (mutually exclusive per column) is worth a sentence in the XML doc on `FilterData`.

Make sure also there are no merge conflicts.

pianomanjh:
Addressed all four points:

1. **Naming** — renamed to `FilterLookupData` / `FilterLookupTextProperty` / `FilterLookupValueProperty` so they read as the lookup side and no longer collide with `FilterProperty`. XML docs updated to match.
2. **Duplication** — collapsed to a single `RadzenListBox` in the CheckBoxList branch; `Data` / `TextProperty` / `ValueProperty` / `Count` / `LoadData` / `AllowFiltering` are now switched based on whether `FilterLookupData` is set, with the item template handling the lookup label vs. the enum/format fallback.
3. **`AllowFiltering`** — added `FilterLookupAllowFiltering` (default `true`) to opt out of the search box.
4. **`LoadColumnFilterData`** — documented on `FilterLookupData` that the two are mutually exclusive per column (when `FilterLookupData` is set it supplies the options and `LoadColumnFilterData` isn't used for that column).

Also rebased onto current `master` (now includes the merged #2554) and squashed to a single commit — no conflicts.

PR #2566, opened 2026-06-08
RadzenSeriesDataLabels - Add Parameter Custom Formatter for Data Labels

msdevcode:
Hello,

a description can be found here: [Radzen Forum](https://forum.radzen.com/t/feature-request-radzenseriesdatalabels-custom-formatter-for-data-labels/21575)

Best regards

### DIFF
--- a/Radzen.Blazor/RadzenSeriesDataLabels.razor
+++ b/Radzen.Blazor/RadzenSeriesDataLabels.razor
@@ -42,7 +42,7 @@
                         builder.AddContent(1,
                                 @<g>
                                     <Text @key="@($"{label.Position}-{(Chart != null ? Chart.Series.IndexOf(series) : -1)}")"
-                                            Value="@label.Text" Position="@label.Position" Fill="@Fill" TextAnchor="@label.TextAnchor" class="@GetSeriesDataLabelClass()" />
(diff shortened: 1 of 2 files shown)


enchev:
Thanks for the PR! Before merging, a couple of concerns:

**This is already achievable via `RadzenValueAxis`.** Both `FormatString` and `Formatter` (`Func<object, string>`) exist on the value axis today and are applied to `label.Text` in every series'' `GetDataLabels` (via `AxisBase.Format`). So `<RadzenValueAxis FormatString="{0:C}" />` or `<RadzenValueAxis Formatter="@(v => ...)" />` already formats data labels. Could you confirm what your scenario can''t accomplish with the existing API?

**If the goal is to format data labels differently from axis ticks**, that''s a legitimate request — but the current design doesn''t quite achieve it. `ChartDataLabel.Text` is already a formatted `string`, so your `Func<object, string>` only ever receives a string, not the underlying numeric value. To do this properly, `RadzenSeriesDataLabels` should expose its own `FormatString` / `Formatter` (mirroring `AxisBase`) and the raw value should be threaded through `ChartDataLabel` from each series'' `GetDataLabels`.

**Minor:** please revert the `<Text>` `Value=` indentation and the unrelated whitespace tweaks to the `Visible`/`Fill` doc comments.

msdevcode:
Thank you for your feedback. Unfortunately, I missed the RadzenValueAxis configuration option. There was no example in the donut and pie charts, which i want to use, so I assumed that this feature might not be available yet.

The added spaces were intended to align with the existing parameters, as they were missing there.

grafik

I am closing this pull request because RadzenValueAxis is available.

PR #2575, opened 2026-06-16
Sync DOM input with bound value on text-input components (Blazor 10)

jspuij:
## Problem

`RadzenMask`, `RadzenTextBox` (Immediate=false), `RadzenTextArea` (Immediate=false), `RadzenPassword` (both Immediate branches), and `RadzenAutoComplete` (both Multiline branches) bind their `<input>`/`<textarea>` with the pre-.NET-7 legacy pattern:

```razor
<input value="@Value" @onchange="@OnChange" ... />
```

Microsoft's binding docs explicitly call this pattern out as buggy:

> The preceding approach can cause synchronization errors, where the `<input>` element displays a different value than the value held in the bound variable. **Migrate to `:get`/`:set` modifiers to avoid this issue.**

— [Use `@bind:get`/`@bind:set` modifiers and avoid event handlers for two-way data binding](https://learn.microsoft.com/aspnet/core/blazor/components/data-binding?view=aspnetcore-10.0#use-%60@bindget%60-%60@bindset%60-modifiers-and-avoid-event-handlers-for-two-way-data-binding)

On the .NET 10 renderer this is consistently reproducible: when a parent rejects the user-typed value (e.g. the downstream parser produces `null` and the bound variable stays unchanged), Blazor's render diff does not write the DOM `value` attribute because the rendered parameter is unchanged across renders. The user-typed text lingers in the field while the bound .NET value reverts to its previous state — visible UI/state divergence.

This was reproduced live downstream when upgrading a Blazor app from .NET 9 to .NET 10. Verified that the same scenarios pass on .NET 9 (the renderer happened to mask the bug there).

## Fix

Migrate each affected razor template to:

```razor
<input @bind:get="@Value" @bind:set="@SetValue" @bind:event="onchange" ... />
```

**The setter must keep `_value` in sync with the typed value when no two-way binding is wired.** An earlier iteration of this PR removed the local `Value = newValue` assignment unconditionally, on the theory that `@bind:get`/`:set` would force-sync the DOM from the bound .NET value by itself. That broke every component in the demos: most use `<RadzenTextBox Change=... />` (or just `Value=`) with no `@bind-Value` or `ValueChanged`, and with nothing updating `_value`, `@bind:get` re-evaluates to the stale initial value on the post-handler render and the framework wipes the DOM on every blur.

The framework's "post-handler sync" is actually plain parameter re-flow. After the `@bind:set` callback fires, Blazor re-renders the component; during that render, `@bind:get` is evaluated and the renderer force-writes the resulting value to the DOM. The mechanism that fixes the rejection case is *parameter re-flow from the parent*, not "_value vs DOM divergence":

- **Bound + accepted:** parent updates its variable, re-renders, passes the new `Value` parameter; the child's `Value` setter overwrites `_value`; `@bind:get` returns the new value → DOM stays at the typed value.
- **Bound + rejected:** parent re-renders with the *same* `Value` parameter as before; Blazor skips `SetParametersAsync` (parameter unchanged), so `_value` stays at the pre-event bound value; `@bind:get` returns that value → DOM is force-synced back.
- **Unbound (`Value=`, no `ValueChanged`):** no parent re-flow happens at all. The handler is the only thing that can update `_value` — if it doesn't, `@bind:get` re-evaluates to the stale initial value and the DOM is wiped on every blur.

The handlers therefore assign locally only when there is no two-way binding:

```csharp
if (!IsBound)
{
    Value = newValue;
}
await ValueChanged.InvokeAsync(newValue);
```

`IsBound` is `ValueChanged.HasDelegate` — already exists on `FormComponent<T>`. This preserves Microsoft's recommended `@bind:get`/`:set` pattern for two-way-bound consumers (the original fix that prompted the PR) while keeping the unbound and one-way `Value=` consumers working the way they did before the migration.

`@onchange` event semantics are preserved (`@bind:event="onchange"`); `@oninput` handlers (e.g. `Radzen.mask`, `OpenScript` in AutoComplete) are kept as-is.

## Components migrated

| Component | Branch | Before | After |
|---|---|---|---|
| `RadzenMask` | both | `value="@Value" @onchange="@OnChange"` | `@bind:get="@Value" @bind:set="@SetValue" @bind:event="onchange"` |
| `RadzenTextBox` | Immediate=false | same | same |
| `RadzenTextArea` | Immediate=false | same | same |
| `RadzenPassword` | both | `@oninput="@OnChange"` / `@onchange="@OnChange"` | `@bind:event="oninput"` / `@bind:event="onchange"` |
| `RadzenAutoComplete` | both | `value="@Value" @onchange="@OnChange"` | `@bind:get/:set` (OpenScript JS oninput and OnFilterKeyPress preserved) |

`RadzenTextBox` Immediate=true already used `@bind:get/:set`; only its `UpdateValue` helper was adjusted to apply the `!IsBound` conditional. The previously-public `OnChange` method on TextBox is retained (still used by Immediate=true `@onchange`).

## Not migrated (intentional)

- `RadzenDatePicker` and `RadzenTimeSpanPicker` also use the legacy attribute pattern but already handle unparseable input by calling `Radzen.setInputValue` via JS interop in their parse handlers — they don't exhibit this bug. Migrating them to `@bind:get`/`@bind:set` would be cleaner architecturally (their `FormattedValue`/`Value` layout differs from `FormComponent<string>`) but is out of scope for this fix.
- `RadzenNumeric` already uses `@bind:get`/`@bind:set` — good pattern, no change.
- `RadzenCheckBox`, `RadzenSwitch`, `RadzenRadioButtonList`, `RadzenCheckBoxList`, `RadzenSlider`, `RadzenRating`, `RadzenChipList`, `RadzenColorPicker`, `RadzenFileInput`, `RadzenSelectBar`, `RadzenSecurityCode`, `RadzenToggleButton`, `RadzenDropDown` family — no editable text input, not affected.

## Tests

**2267/2267 passing locally on net10.0.**

Two regression tests per migrated component, covering both code paths:

- `*_SyncsDomValue_WhenParentRejectsInput` — renders the component inside a wrapper that holds `Value` at `"original"` and ignores `ValueChanged`; after the user types `"user-typed"`, both `Instance.Value` and the DOM value must be back to `"original"`. Exercises the bound-but-rejected path.
- `*_KeepsTypedValue_WhenUsedWithoutTwoWayBinding` — renders the component with no `Value`/`ValueChanged` wiring at all (the demo's `<RadzenTextBox Change=... />` pattern); after the user types `"user-typed"`, both `Instance.Value` and the DOM value must equal `"user-typed"`. Exercises the unbound path.

Three existing `RadzenTextBox` trim tests that previously asserted `Instance.Value` after a raw `ValueChanged` callback now use a small wrapper component so the parameter round-trips properly — same effect a user gets when they use `@bind-Value`.

## Behavioural impact for consumers

- **`@bind-Value` consumers**: no behavioural change. Bound variable updates the same way; DOM now stays in sync with the bound value across rejection scenarios that previously left the field out of sync.
- **`Value=` one-way or unbound consumers**: no behavioural change. `Value`/`_value` still updates locally from the typed value, exactly as it did with the legacy pattern.
- **`@bind-Value` consumers that explicitly reject input**: this is the case the migration fixes. Previously the DOM kept the typed text while the bound variable stayed at its old value (visible drift). Now the DOM is force-synced to the bound value on the post-handler render.

### DIFF
--- a/Radzen.Blazor.Tests/AutoCompleteTests.cs
+++ b/Radzen.Blazor.Tests/AutoCompleteTests.cs
@@ -282,6 +282,22 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void AutoComplete_KeepsTypedValue_WhenUsedWithoutTwoWayBinding()
+        {
+            // Demo-style usage: no @bind-Value or ValueChanged. The user-typed value
+            // must survive the post-handler @bind:get/:set sync, otherwise the input
+            // clears itself on every blur.
+            using var ctx = new TestContext();
+
+            var component = ctx.RenderComponent<RadzenAutoComplete>();
+
+            component.Find("input").Change("user-typed");
+
+            Assert.Equal("user-typed", component.Instance.Value);
+            Assert.Equal("user-typed", component.Find("input").GetAttribute("value"));
+        }
+
+        [Fact]
         public void AutoComplete_SyncsDomValue_WhenParentRejectsInput()
         {
             using var ctx = new TestContext();
--- a/Radzen.Blazor.Tests/AutoCompleteTests.cs
+++ b/Radzen.Blazor.Tests/AutoCompleteTests.cs
@@ -280,5 +280,32 @@ namespace Radzen.Blazor.Tests
         {
             public IEnumerable CurrentView => View;
         }
+
+        [Fact]
+        public void AutoComplete_SyncsDomValue_WhenParentRejectsInput()
+        {
+            using var ctx = new TestContext();
+
+            var wrapper = ctx.RenderComponent<RadzenAutoCompleteRejectWrapper>();
+
+            wrapper.Find("input").Change("user-typed");
+
+            Assert.Equal("original", wrapper.FindComponent<RadzenAutoComplete>().Instance.Value);
+            Assert.Equal("original", wrapper.Find("input").GetAttribute("value"));
+        }
+
+        private sealed class RadzenAutoCompleteRejectWrapper : Microsoft.AspNetCore.Components.ComponentBase
+        {
+            public string HeldValue { get; private set; } = "original";
+
+            protected override void BuildRenderTree(Microsoft.AspNetCore.Components.Rendering.RenderTreeBuilder builder)
+            {
+                builder.OpenComponent<RadzenAutoComplete>(0);
+                builder.AddAttribute(1, nameof(RadzenAutoComplete.Value), HeldValue);
+                builder.AddAttribute(2, nameof(RadzenAutoComplete.ValueChanged),
+                    Microsoft.AspNetCore.Components.EventCallback.Factory.Create<string>(this, _ => StateHasChanged()));
+                builder.CloseComponent();
+            }
+        }
     }
 }
--- a/Radzen.Blazor.Tests/AutoCompleteTests.cs
+++ b/Radzen.Blazor.Tests/AutoCompleteTests.cs
@@ -288,6 +288,7 @@ namespace Radzen.Blazor.Tests
             // must survive the post-handler @bind:get/:set sync, otherwise the input
             // clears itself on every blur.
             using var ctx = new TestContext();
+            ctx.JSInterop.Mode = JSRuntimeMode.Loose;
 
             var component = ctx.RenderComponent<RadzenAutoComplete>();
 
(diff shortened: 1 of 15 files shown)


enchev:
Such attempt was tried before (https://github.com/radzenhq/radzen-blazor/pull/2453) and later reverted since a regression was introduced in v9.0.2 where typing fast in a filtered RadzenDropDown/RadzenDropDownDataGrid on Blazor Server with network latency caused the filter input to flicker/jump — characters appearing, disappearing, reappearing, and the field "fighting back" against backspace.

jspuij:
OK, I'll have a look at that and see if I can reproduce these effects and think of something.

jspuij:
I've changed the PR a bit to distinguish between a bound and unbound component and clicked through all the demo pages, with a big latency (200ms) and I haven't been able to find any differences between the master branch and this branch.

PR #2586, opened 2026-06-26
Refine chat mention paging API

MPapst:
Refactoring of the Mention User Search API to align it more with the `LoadData` API

- Replace SetMentionSearchResults with MentionUsers and MentionUsersCount
- Add MentionDisplaySize paging and scroll-driven next-page loading
- Update demo and tests to use the property-based mention flow


MPapst:
sorry for the additional PR - I did not expect the first to be merged that fast without any discussion :)

enchev:
It looked good to me and that’s why we merged it - I’ll merge this as well! 🙂

PR #2595, opened 2026-06-29
Added a new property to sort columns of column picker in RadzenDataGrid

maxueljunior:
# RadzenDataGrid: allow sorting columns inside the column picker (`AllowSortingColumnPicker`)

## Why

The column picker of `RadzenDataGrid` is rendered with a `RadzenDropDown`, which presents the pickable columns as a flat, **unsortable** list. When a grid has many columns, end users have no way to reorder/sort that list to quickly find and toggle the column they want — they have to scan the whole list every time.

Our users rely heavily on this picker and repeatedly asked for the ability to **sort the columns** in the picker (e.g. alphabetically) the same way they can sort data in a normal grid.

We tried several times to implement this on our side, in our own project, by wrapping/overriding the picker — but every approach ended up **conflicting with Radzen's internals** (the picker's selection state, `selectedColumns` binding, and `ToggleColumns` visibility logic are internal to the component, so any external reimplementation either broke selection sync or duplicated/fought the built-in behavior). The clean, conflict-free place to add this is inside the component itself, which is why we are contributing it back upstream.

## What was done

Added a new opt-in `[Parameter] bool AllowSortingColumnPicker` to `RadzenDataGrid`. It is **`false` by default**, so existing behavior is 100% unchanged.

- **`false` (default)** → the picker renders the existing `RadzenDropDown`, exactly as before.
- **`true`** → the picker renders a `RadzenDropDownDataGrid` with `AllowSorting="true"`, so the column list becomes sortable.

Because `RadzenDropDownDataGrid` (unlike `RadzenDropDown`) does **not** render a selection checkbox per row automatically, the new branch declares an explicit `<Columns>` definition that reproduces the multi-select UX (mirroring the official `DropDownDataGridMultiple` demo) while staying wired to the grid's existing picker plumbing:

- A non-sortable checkbox column with:
  - a **"select all" checkbox** in the `HeaderTemplate` (selects/clears every pickable column and calls the existing `ToggleColumns`);
  - a **per-row checkbox** in the `Template` that calls `columnPickerGrid.SelectItem(...)`, which updates the bound `selectedColumns` and raises `Change=@ToggleColumns` — i.e. reuses the exact same selection/visibility path as the original dropdown.
- A sortable `ColumnPickerTitle` column so the user can sort the columns in the picker.

Both branches keep the same bindings (`@bind-Value="selectedColumns"`, `Change="ToggleColumns"`, `Data="allPickableColumns"`, `TextProperty="ColumnPickerTitle"`, `AllowSelectAll`, `MaxSelectedLabels`, `AllowFiltering`, etc.), so picked-column state, settings persistence (`SaveSettings`), and `PickedColumnsChanged` continue to work identically.

### Files changed

- `Radzen.Blazor/RadzenDataGrid.razor`
  - New `@if (!AllowSortingColumnPicker) { … } else { … }` around the column picker.
  - New `RadzenDropDownDataGrid` branch with `@ref="columnPickerGrid"`, the checkbox column (header "select all" + per-row), and the sortable `ColumnPickerTitle` column.
- `Radzen.Blazor/RadzenDataGrid.razor.cs`
  - New parameter `public bool AllowSortingColumnPicker { get; set; }`.
  - New private field `RadzenDropDownDataGrid<object>? columnPickerGrid;` used by the new branch's `@ref`.

### Usage

```razor
<RadzenDataGrid Data="@orders" AllowColumnPicking="true" AllowSortingColumnPicker="true">
    ...
</RadzenDataGrid>
```

## The reason (rationale for upstreaming)

We attempted to build this purely inside our own project several times, but each attempt **conflicted with Radzen** — the picker's selection state and visibility toggling are internal, so an external workaround either desynced `selectedColumns`/`ToggleColumns` or fought the built-in dropdown. Since our users specifically and consistently want the **sort** capability in the column picker, the correct, maintainable solution is a small, backward-compatible, opt-in feature in `RadzenDataGrid` itself.

## Backward compatibility

- New parameter defaults to `false`; with it unset, the rendered markup and behavior are identical to before.
- No changes to public selection/visibility APIs; the new branch reuses `selectedColumns`, `ToggleColumns`, `SaveSettings`, and `PickedColumnsChanged`.
- Builds clean against `net8.0`, `net9.0`, and `net10.0` (0 warnings, 0 errors).


enchev:
Thanks for the contribution - few things in my opinion should be addressed before we merge:

1. **`AllowPickAllColumns` is ignored in the new branch.** `AllowSelectAll` isn't referenced anywhere in `RadzenDropDownDataGrid`, so the custom "select all" header checkbox always renders regardless of `AllowPickAllColumns`. In the default branch `AllowPickAllColumns="false"` hides select-all, so the two branches now behave differently. Please gate the header checkbox on `AllowPickAllColumns`.

2. **Select-all "checked" semantics.** The header checkbox uses `.Any()`, so it shows as checked whenever *any* column is picked (even one). A select-all should be checked only when all are selected (ideally indeterminate otherwise).

3. **Hardcoded aria-label "select column".** The rest of the picker uses localizable parameters (e.g. `SelectVisibleColumnsAriaLabel`). Please introduce a localizable parameter instead of an inline English string so localization stays consistent.

4. Nice-to-have: a small bUnit test covering the sorting branch (toggle a column, sort the picker) would help, and please add a trailing newline to the new demo file.

maxueljunior:
Thanks for the review @enchev ! I've updated the implementation according to your suggestion.

We already added the example in DataGridColumnPickingPage.razor in the first commit.

PR #2600, opened 2026-07-03
Set the width of theY-axis in pixels on a graph

wimsoetens-cmd:
Sets the width of the axis in pixels. If not specified, the width is calculated automatically. Setting a fixed width can help reduce unnecessary spacing when using multiple Y-axes.

### DIFF
--- a/Radzen.Blazor.Tests/AxisBaseWidthTests.cs
+++ b/Radzen.Blazor.Tests/AxisBaseWidthTests.cs
@@ -1,0 +1,91 @@
+using Microsoft.AspNetCore.Components;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit;
+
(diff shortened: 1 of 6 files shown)


enchev:
Few things should be addressed before merging:

1. Width is not included in AxisBase.ShouldRefreshChart, so changing it at runtime will not re-layout the chart. All other layout-affecting parameters (Min, Max, Visible, etc.) are listed there.

2. The parameter only has effect for vertically rendered value axes. It is ignored on the category axis in normal charts and on value axes in inverted (bar) charts, since horizontal measurement uses AxisMeasurer.XAxis which is not changed. This should at least be documented.

3. There is no validation - Width="0" or a negative value is returned as-is and bypasses the existing minimum of 24px, which can break the layout.

4. No tests added, the comment in AxisMeasurer restates the code, and there is some unrelated formatting churn.

PR #2602, opened 2026-07-04
Implement Excel-style tiled clipboard paste

panoskentros:
pasting functionality for multi-cell selection or even entire column

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/PasteCommandTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/PasteCommandTests.cs
@@ -1,6 +1,6 @@
 using Xunit;
-
 using Radzen.Documents.Spreadsheet;
+
 namespace Radzen.Blazor.Spreadsheet.Tests;
 
 /// <summary>
@@ -17,7 +17,7 @@ public class PasteCommandTests
         var clipboard = new SpreadsheetClipboard();
         var stack = new UndoRedoStack(sheet);
 
-        stack.Execute(new PasteCommand(clipboard, sheet, new CellRef(0, 0), "X\tY"));
+        stack.Execute(new PasteCommand(clipboard, sheet, new RangeRef(new CellRef(0, 0), new CellRef(0, 0)), "X\tY"));
 
         Assert.Equal("X", sheet.Cells[0, 0].Value);
         Assert.Equal("Y", sheet.Cells[0, 1].Value);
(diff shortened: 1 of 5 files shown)


akorchev:
Hi @panoskentros, thanks for the contribution - and welcome! Paste-into-selection is a feature we definitely want, so we're glad you picked it up.

Before reviewing we tested real Excel to pin down what "Excel-style" means here, and the result reframes the approach a bit:

**Internal clipboard (content copied inside Excel):**

| Scenario | Excel behavior |
|---|---|
| Copy 1 cell, paste into a multi-cell selection | Fills the entire selection |
| Copy a block, paste into a selection that is an exact multiple of it | Tiles the block to fill the selection |
| Copy a block, paste into a larger selection that is not an exact multiple | Pastes once, anchored top-left |
| Copy a block, paste into a smaller selection | Pastes the full block - never truncates |

**External plain text (clipboard content from another application):**

| Scenario | Excel behavior |
|---|---|
| Single value, multi-cell selection | Pasted once at the anchor - no fill |
| Multi-line text, selection an exact multiple | Pasted once - no tiling |
| Ragged rows (`A\tB\nC`) | Pastes fine, missing cells left empty |

So the fill/tiling behavior belongs to the internal clipboard path, while this PR adds it to the external-text path. Because internal copies round-trip through the OS clipboard as text, the new branch also runs before `TryPaste` can recognize an internal copy - so pasting an internal copy into a multi-cell selection now loses formulas and formats, and cut+paste stops clearing the source. There are also a few edge cases in the text splitting (ragged rows crash, tab-prefixed text is dropped).

The good news: the Excel-faithful version is less code. Leave the external-text path as it is, and add the fill/tiling logic inside the internal-clipboard path (around `SpreadsheetClipboard.Paste`, after `TryPaste` matches): single cell: fill the selection; exact multiple: tile via repeated `PasteRange`; otherwise - the current behavior. Everything then inherits formula adjustment, formats, cut semantics and undo for free - and it covers your use case (copy a value, select a column, paste everywhere) exactly the way Excel does it.

Would you be up for reworking it along these lines? We know it's a different shape than the current diff, but it's a smaller and more robust change - and we're glad to help along the way. (One small ask: please also remove the accidental duplicated lines in `PasteCommandTests.cs` and restore the comments that got dropped there.) Thanks again for kicking this off!

panoskentros:
Hi @akorchev and thanks for the detailed answer! I'll do my best to make it work just like you described it. (Also excuse me for the accidental duplicated lines inside the unit tests)

panoskentros:
I've pushed the reworked logic based on the feedback. I checked all 7  use cases mentioned above and it works as intended. For some reason,tho, it doesnt copy the format of the cell for example: the background color or the underline. May you please continue with the feature implementation from here? Thanks!!

akorchev:
Thanks!

The format not being copied is a preexisting issue which we will handle separately.

PR #2603, opened 2026-07-05
Readjust column width to fit longest cell by double clicking edge

panoskentros:
with a cap of 800px

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/AutoFitXlsxRoundTripTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/AutoFitXlsxRoundTripTests.cs
@@ -117,6 +117,60 @@ public class AutoFitXlsxRoundTripTests
     }
 
     [Fact]
+    public void Write_AutoFitColumn_WidensToFitContentWithExcelMetrics()
+    {
+        var wb = Build();
+        var sheet = wb.Sheets[0];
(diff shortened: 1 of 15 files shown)


akorchev:
Hi @panoskentros, thanks for another contribution! Auto-fit on double-click is a feature we definitely want, and the wiring here is solid: the JS event handling, the `[JSInvokable]` entry point, the feature gate, and the undoable `ResizeColumnCommand` are all exactly right.

We verified Excel's actual auto-fit behavior, since that's the spec we want to follow:

| Scenario | Excel behavior |
|---|---|
| Formula cell (long formula, short result) | Sizes to the **displayed value**, not the formula text |
| Number-formatted value (e.g. a date) | Sizes to the **formatted** text, not the raw value |
| Auto-fit on an empty column | No-op - keeps the current width |
| Long content in a hidden row | Included |
| Long text in a merged cell | Ignored |

The current implementation differs on several of these, and the `length * 8` width estimate can't ship as-is - the width has to be measured against the rendered text, which depends on the active theme's font, so it needs JS-side text measurement. That infrastructure is a bigger lift than is fair to ask from you, so we'll build it on top of your branch ourselves, keeping your commit as the foundation - your event wiring and command plumbing stay.

No changes needed from your side. Thanks for kicking the feature off - we appreciate the initiative!

PR #2609, opened 2026-07-08
Enable pasting into unlocked cells in protected sheets

panoskentros:
Issue: 
Currently, when sheet.Protection.IsProtected is set to true, the RadzenSpreadsheet component globally disables clipboard pasting. This happens even for cells where Format.Locked is explicitly set to false.

Example:
Try pasting into column E: [https://blazor.radzen.com/spreadsheet-protection](https://blazor.radzen.com/spreadsheet-protection)

Changes:
1. RadzenSpreadsheet.cs: Updated ExecuteAsync to allow the PasteCommand if the target cell is editable, even when the sheet is protected.
2. Worksheet.cs: Updated InsertDelimitedString to verify the Locked property of each cell before writing data.

Tests:
1. Paste_Respects_CellLocking_InProtectedSheet: Verifies that within a protected sheet, locked cells are ignored while unlocked cells are correctly updated during a multi-cell paste.
2. Paste_Works_When_Protection_Is_Disabled: Ensures that existing paste behavior remains unchanged and fully functional when protection is not enabled.

All existing and new unit tests have passed.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/SpreadsheetClipboardTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/SpreadsheetClipboardTests.cs
@@ -58,4 +58,36 @@ public class SpreadsheetClipboardTests
 
         Assert.Equal("=B2", target.Cells[1, 1].Formula);
     }
+    
+    [Fact]
+    public void Paste_Respects_CellLocking_InProtectedSheet()
+    {
+        var sheet = new Worksheet(5, 5);
+    
+        sheet.Cells[0, 0].Format = new Format { Locked = true };
+        sheet.Cells[0, 1].Format = new Format { Locked = false };
+        sheet.Protection.IsProtected = true;
+
+        var clipboard = new SpreadsheetClipboard();
+    
+        clipboard.Paste(sheet, RangeRef.Parse("A1"), "BlockedValue\tAllowedValue");
+
+        Assert.Null(sheet.Cells[0, 0].Value);
+    
+        Assert.Equal("AllowedValue", sheet.Cells[0, 1].Value); 
+    }
+    [Fact]
+    public void Paste_Works_When_Protection_Is_Disabled()
+    {
+        var sheet = new Worksheet(5, 5);
+
+        sheet.Cells[0, 0].Format = new Format { Locked = true };
+        sheet.Protection.IsProtected = false;
+
+        var clipboard = new SpreadsheetClipboard();
+
+        clipboard.Paste(sheet, RangeRef.Parse("A1"), "AllowedValue");
+
+        Assert.Equal("AllowedValue", sheet.Cells[0, 0].Value);
+    }
 }
--- a/Radzen.Blazor.Tests/Spreadsheet/SpreadsheetClipboardTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/SpreadsheetClipboardTests.cs
@@ -59,7 +59,6 @@ public class SpreadsheetClipboardTests
 
         Assert.Equal("=B2", target.Cells[1, 1].Formula);
     }
-    
     [Fact]
     public void PasteCommand_Succeeds_When_All_Target_Cells_Unlocked_In_Protected_Sheet()
     {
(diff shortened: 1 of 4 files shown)


akorchev:
Hi @panoskentros,

Thanks for looking into this and for including tests. The underlying issue is real: typing into unlocked cells on a protected sheet works, but pasting doesn't, which is inconsistent. However, we verified how Excel actually behaves before merging, and the suggested implementation diverges from it, so we can't accept it as-is.

**What Excel does** (verified against Excel desktop; Google Sheets behaves the same way with protected ranges):

- Paste into a destination where *every* cell is unlocked: allowed.
- Paste into a destination containing *any* locked cell: the **entire paste is rejected** with an error. Excel never skips locked cells and pastes the rest - even when the anchor cell is unlocked and only a spilled-into cell is locked.

The PR implements partial paste (skip locked, write unlocked), which neither Excel nor Google Sheets does, and it silently drops data without feedback.

There are also two implementation issues:

1. `ExecuteAsync` checks only `Worksheet.Selection.Cell`. A paste anchored at an unlocked cell that spans locked cells passes the check. The full destination range needs validation - `clipboard.GetPasteRange(...)` gives you that range.
2. The lock check was added only to `InsertDelimitedString`, which handles external text paste. Internal copy/paste (`TryPaste` -> `PasteRange`) has no lock checks, so once the gate lets a `PasteCommand` through, an in-app copy/paste would overwrite locked cells. That's a protection bypass.

**What we'd accept instead:**

- In the protection gate (or `PasteCommand`), compute the full destination range via `GetPasteRange` and allow the paste only if every cell in it passes `IsCellEditable`; otherwise block the whole command.
- Remove the skipping logic from `Worksheet.InsertDelimitedString` - protection enforcement should stay at the command layer, consistent with the rest of the codebase.
- Update the tests to go through the command/gate path and assert the all-or-nothing behavior (paste into fully unlocked range succeeds; paste overlapping any locked cell changes nothing).

Happy to review an updated version along these lines.

panoskentros:
Thanks for the detailed answer @akorchev !! I refactored the implementation to meet Excel's behavior. Please let me know if I missed anything

akorchev:
Thanks for the quick turnaround - this is much closer. I ran the full test suite against your branch and there are no regressions.

There is one remaining protection bypass, plus some cleanup, before we can merge:

**Cut/paste can still clear locked cells.** On a protected sheet: cut a locked cell (Ctrl+X - `CutSelectionAsync` has no protection check), then paste into an unlocked cell. The destination check passes, and the move operation then clears the locked source cell. I verified this on your branch with a test - the locked cell's value gets wiped. Excel blocks the *cut itself* with the protected-sheet error when the selection contains locked cells (verified against Excel desktop), so the preferred fix is to reject the cut in `CutSelectionAsync` when the sheet is protected and the selection contains any locked cell. Alternatively `PasteCommand` could also validate the move source range, but failing at cut time matches Excel.

**Cleanup:**

- `Worksheet.cs` should have no diff now: `InsertDelimitedString` was restored but re-indented with 3 spaces and the blank line before the following doc comment was dropped. Please make it a byte-identical revert.
- In `ExecuteAsync`, the `ReadOnly` and feature checks were merged into one `if` - unrelated churn, please revert.
- A short comment on the `command is not PasteCommand` special case explaining why paste bypasses the sheet-level `EditCell` block (per-cell enforcement happens inside the command) would help future readers.

A test for the cut scenario would be good too: protected sheet, locked A1 with a value, cut A1, paste into an unlocked cell - the locked cell must keep its value.

panoskentros:
- I restored the Worksheet.cs file. Now it has no diff to the master version
- I changed the signature of CutSelectionAsync from private to internal to be able to use it in unit tests
- Fixed protection during cut operation
- PasteCommand: Added the requested comment explaining the PasteCommand exception.
- Reverted the seperate if's in ExecuteAsync 


One question.... would it be better if we had a IsRangeEditable(RangeRef) function inside the Worksheet.cs so it doesnt get repetitive with the for loops?

akorchev:
Thanks for the quick turnaround - this is very close now. 

To answer your question: yes, add `Worksheet.IsRangeEditable(RangeRef range)` and use it instead of the repeated loops. 

One remaining issue: the same locked-cell check was added to `CopySelectionAsync`, and that one shouldn't be there. Copying doesn't modify anything, so protection should not block it - verified in Excel and Google Sheets. Remove the check from `CopySelectionAsync` and keep it only in `CutSelectionAsync`.

There is some trailing whitespace on the new `var range = Worksheet.Selection.Range;` lines and a few indented blank lines around the new blocks.

panoskentros:
- Removed trailing whitespaces 
- Removed check from CopySelectionAsync
- Added new Worksheet.IsRangeEditable function

Excuse me for the stupid mistakes

PR #2614, opened 2026-07-10
Add Gantt resource view and multi-year timescale

### LINKED ISSUE #2488: Feature Request: Gantt - Resource View

I´ve have looked into the newly created Gantt-Component and its samples. For my case only half of what i 
want to do is possible with this.
I want to create an overview about my resources (computing slots) and want to see the blocked times in 
the same line as the resource is presented.

A resource view should work for this task.
The view should provide a required column (the resource) which can be multi hierachical. The same 
possibility to add custom columns like its done yet.
In the time view part, there should be a bar for each data item in this time window. 
If the time window is i overlapping, the bars should wrap over.
All other features like chaining and moving should also be possible. 

Bonus: I histogram (per time unit accumulated values + synchronized line chart on the bottom of the gantt view would be nice.

I considered using DHTML Gantt but it has no Blazor bindings and looks and behaves alien like in the application.

**Additional context**

Pure resource view:
Image

With Histogram:
Image

enchev:
Fix #2488; #2588

### DIFF
--- a/Radzen.Blazor.Tests/DataGridTests.cs
+++ b/Radzen.Blazor.Tests/DataGridTests.cs
@@ -3495,8 +3495,9 @@ namespace Radzen.Blazor.Tests
             });
 
             Assert.Equal("presentation", component.Find("table.rz-grid-table").GetAttribute("role"));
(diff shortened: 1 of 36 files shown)


enchev:
Merged into the v12 branch.

PR #2615, opened 2026-07-10
Extend IsLoading and LoadingTemplate to DropDown and ListBox

### LINKED ISSUE #1343: Feature Request: Extend IsLoading functionality across more data bound components
Datagrid IsLoading is a very useful piece of functionality. Could this be extended to more components in the library?

Components I've identified that could be useful to also have this functionality are:

- AutoComplete
- DropDown
- ListBox 

Could we add the IsLoading property to DataBoundFormComponent?

Alternatively we could create a wrapper component that contains the inline spinner

artnim:
Closes #1343

## What

Extends the `IsLoading` / `LoadingTemplate` capability to `RadzenDropDown` and `RadzenListBox` by declaring both parameters once on the shared `DataBoundFormComponent<T>` base class — exactly as suggested in #1343.

```razor
<RadzenDropDown @bind-Value=@value Data=@items IsLoading=@isLoading LoadData=@OnLoadData>
    <LoadingTemplate>
        <RadzenProgressBarCircular Size="ProgressBarCircularSize.ExtraSmall" ShowValue="false" Mode="ProgressBarMode.Indeterminate" />
    </LoadingTemplate>
</RadzenDropDown>
```

## Changes

- **`DataBoundFormComponent<T>`**: new `IsLoading` and `LoadingTemplate` parameters (docs match the existing AutoComplete wording).
- **`RadzenAutoComplete`** and **`RadzenDropDownDataGrid`**: their own `IsLoading`/`LoadingTemplate` declarations are removed in favor of the inherited ones — public API unchanged (DropDownDataGrid keeps forwarding `IsLoading` to its inner `RadzenDataGrid`).
- **`RadzenDropDown`**: while `IsLoading` and a `LoadingTemplate` is set, the popup renders the template instead of the items; the `EmptyTemplate` is suppressed while loading.
- **`RadzenListBox`**: same behavior in its list area, including the `EmptyText` message.

Behavior mirrors the AutoComplete precedent: no default spinner is rendered when `LoadingTemplate` is not set.

## Tests

7 new bUnit tests: LoadingTemplate rendering (on/off) and empty-state suppression for both DropDown and ListBox, plus an `IsLoading` forwarding regression test for DropDownDataGrid. The existing AutoComplete loading tests cover the base-class consolidation. Full suite passes.

## Demos

- **DropDown → Loading indicator**: `LoadData` with a simulated delay; loading state shows while filtering.
- **ListBox → Loading indicator**: loading state on initial load and when re-filtering.

Both registered with ToC entries on their respective demo pages.


enchev:
Thanks @artnim! One note: RadzenDropDownDataGrid now inherits LoadingTemplate too, but it isn't forwarded to the inner RadzenDataGrid (which has its own LoadingTemplate), so setting it there does nothing. Could you forward it to the inner grid instances as well?

artnim:
Good catch - done. `LoadingTemplate` is now forwarded to all three inner `RadzenDataGrid` instances (it renders inside the grid's loading overlay), with a test covering the forwarding.

PR #2625, opened 2026-07-22
Marching ants effect on copy & cut

panoskentros:
I implemented excel's marching ants

Copy Behavior: When a range is copied, the marching ants persist even after pasting, allowing users to paste multiple times.
Cut Behavior: When a range is cut, the marching ants automatically clear immediately after a paste operation. The clipboard is not cleared

Also i noticed that Excel clears the clipboard on ESC key pressed. I can implement this in another pr if it is something we want

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/CopyOverlayTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/CopyOverlayTests.cs
@@ -12,38 +12,37 @@ public class CopyOverlayTests : TestContext
     public void CopyOverlay_RendersNothing_When_Clipboard_Empty()
     {
         var sheet = new Worksheet(10, 10);
-        var spreadsheet = new RadzenSpreadsheet();
+        var clipboard = new SpreadsheetClipboard();
         var context = new CopyMockContext(sheet);
 
         var cut = RenderComponent<CopyOverlay>(parameters => parameters
             .Add(p => p.Worksheet, sheet)
-            .Add(p => p.Spreadsheet, spreadsheet)
-            .Add(p => p.Context, context));
+            .Add(p => p.Context, context)
+            .AddCascadingValue(clipboard));
 
         Assert.Empty(cut.FindAll(".rz-spreadsheet-copy-overlay"));
     }
 
     [Fact]
-    public void CopyOverlay_RendersSvg_When_Clipboard_Has_Source()
+    public void CopyOverlay_RendersEdges_When_Clipboard_Has_Source()
     {
         var sheet = new Worksheet(10, 10);
         sheet.Selection.Select(RangeRef.Parse("B2:C3"));
         
-        var spreadsheet = new RadzenSpreadsheet();
-        spreadsheet.clipboard.Copy(sheet);
+        var clipboard = new SpreadsheetClipboard();
+        clipboard.Copy(sheet);
 
         var context = new CopyMockContext(sheet);
 
         var cut = RenderComponent<CopyOverlay>(parameters => parameters
             .Add(p => p.Worksheet, sheet)
-            .Add(p => p.Spreadsheet, spreadsheet)
-            .Add(p => p.Context, context));
+            .Add(p => p.Context, context)
+            .AddCascadingValue(clipboard));
 
-        var svgs = cut.FindAll(".rz-spreadsheet-copy-overlay");
-        Assert.Single(svgs);
+        var overlays = cut.FindAll(".rz-spreadsheet-copy-overlay");
+        Assert.Single(overlays);
 
-        var rect = cut.Find("rect");
-        Assert.NotNull(rect);
+        Assert.NotEmpty(cut.FindAll(".rz-copy-edge"));
     }
 
     [Fact]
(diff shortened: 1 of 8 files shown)


akorchev:
Thanks - the behavior is right and we want the feature. The wiring needs rework before it goes in.

## 1. Derive the range from the clipboard, don't track it separately

`SpreadsheetClipboard` already stores `range`, `sheet` and `operation`. `CopiedRange` duplicates that and drifts from it: `TryPaste` clears the clipboard when the pasted text doesn't match the internal CSV (external paste after a copy), but `CopiedRange` survives, so the ants keep marching around a source that is no longer on the clipboard.

Add an accessor next to `TryGetMoveSource`:

```csharp
public RangeRef GetSource(Worksheet targetSheet) =>
    range.HasValue && ReferenceEquals(sheet, targetSheet) ? range.Value : RangeRef.Invalid;
```

That removes `CopiedRange`, `IsCut`, both `isCut` blocks and the duplicated clear logic in `OnPasteAsync` / `PasteFromClipboardAsync`. It also fixes a bug: `CopiedRange` is spreadsheet-level, not sheet-level, so copying on Sheet1 and switching tabs draws the marquee at the same coordinates on Sheet2. The `ReferenceEquals` check handles that.

## 2. Make it an overlay component, not inline markup in RadzenSpreadsheet.razor

Please add a `CopyOverlay` alongside `SelectionOverlay` / `AutofillOverlay` / `ImageOverlay`, taking `Context` and using `Context.GetRanges(...)` + `Context.GetRectangle(...)`.

This isn't only consistency - the inline math is wrong for frozen panes. `GetRectangle` subtracts scroll only for non-frozen rows/columns and adds `SplitterWidth`/`SplitterHeight`, and `GetRanges` splits a range across the freeze line. `cols.Start - grid.ScrollLeft` does neither, so with frozen rows/columns the marquee is off by the scroll offset plus the splitter, and a range straddling the freeze line is drawn as one wrong box. Merged cells at the edges are ignored too. Reaching into `grid.View` and `grid.ScrollLeft` from the top-level razor is the signal it's at the wrong layer.

## 3. StateHasChanged should fall out of 1 and 2, not be deleted on its own

To be clear, as the PR stands those five calls are load-bearing - every path in is JS interop (Ctrl+C/X/V and Escape via `[JSInvokable] OnKeyDownAsync`, the native clipboard events via `OnCopyAsync`/`OnPasteAsync`, the context menu via a plain `Action` plus `_ = InvokeAsync(...)`), none of which render automatically. Remove them in isolation and the marquee never appears.

But `StateHasChanged` on `RadzenSpreadsheet` is the wrong lever: it re-renders `VirtualGrid`, and since nothing under `Spreadsheet/` overrides `ShouldRender`, that re-renders every visible `CellView`, `ColumnHeader` and `RowHeader` on every copy, cut, paste and Escape. The root component calls `StateHasChanged` twice in ~3200 lines today; this takes it to seven.

The pattern here is the opposite: state lives in the model, the model raises a change event, the view binds to it with `EventBinding` and renders itself (`SelectionOverlay` on `Selection.Changed`, `CellEditor` on `Editor.Changed`, `VirtualGrid` on `Rows.Changed`). Do that and the five calls disappear because only the overlay re-renders.

## 4. RTL

`left:` / `top:` are hardcoded, but the component is consistently logical-property based (`inset-inline-start`, `border-inline-end`, plus a `[dir="rtl"]` rule in `_spreadsheet.scss`). Using `PixelRectangle.ToStyle()` / `.ToPx()` like `RangeSelectionItem` fixes RTL and drops the four hand-rolled `ToString("0.##", InvariantCulture)` calls.

## Smaller items

- The `Math.Min` / `Math.Max` over `Start` / `End` are dead code - `RangeRef`'s constructor already normalizes.
- `+ 4`, `- 2`, `- 2` are unexplained. If they're the splitter size and the 2px border, take them from `SplitterSize` and the CSS var.
- Rename `marching-ants-animation` to `rz-spreadsheet-marching-ants` - theme keyframes are `rz-` prefixed and this one is global.
- `z-index: 100`, but the spreadsheet overlay stack is 1-4.
- The marquee uses `--rz-primary`, the same as the selection border, so copied and selected ranges look identical. Excel differentiates them.
- Add a `prefers-reduced-motion` guard - this animation runs indefinitely by design.
- Trailing whitespace on several added lines.

## Also worth covering

- Stale marquee after structural edits: insert/delete rows or columns, sort, filter or undo and it points at the wrong cells. Excel moves the marquee with the source; clearing it on those commands is an acceptable minimum.
- No tests. `Radzen.Blazor.Tests/Spreadsheet/` already covers both layers (`SpreadsheetClipboardTests`, `RangeSelectionItemTests`, `ImageOverlayTests`), so there's a home for: marquee survives a copy-paste, clears on cut-paste / Escape / external paste, isn't drawn on a sheet other than the source, and matches `Context.GetRectangle` with frozen rows/columns.
- On Escape clearing the clipboard: let's do it here rather than in a separate PR. `CancelEditAsync` is already the Escape handler you're hooking, and once the marquee is clipboard-derived it's the same change.

panoskentros:
I reset to master and started over again. Please let me know if I missed something! I'm glad to review the code as many times as needed

panoskentros:
Also, I made the escape global so it clears the copy marquee and cancels edits even when focus is in the formula bar or toolbar to follow Excels behavior

akorchev:
Thanks - the architecture is right now and the clipboard state machine is solid. I checked out the branch and ran the suite: all 15 tests pass, and scroll tracking, RTL and the cross-sheet check all work. A few things left:

## 1. The marquee paints over the headers

Copy a range and scroll it under the header row - the ants draw on top of the headers. Headers sit at z-index 2 (4 over frozen areas), so the hardcoded `z-index: 3` beats them. Use the selection's ladder instead: base 1, `rz-spreadsheet-frozen-row`/`-column` 3, both 4, driven by `RangeInfo.FrozenRow`/`FrozenColumn` like `AutofillOverlay.HandleClass`. That also stops the marquee painting over frozen panes.

## 2. Draw only the outer edges, with divs instead of SVG

A range straddling a freeze line currently shows an ants line through its middle, because every split region draws all four edges. Use the `Top`/`Left`/`Bottom`/`Right` flags to draw only the original range's edges, like `RangeSelectionItem` does. Once it's per-edge, replace the SVG (the only one in the overlays; its stroke is also clipped 1px by the viewport) with one strip div per flagged edge: 2px thick, `repeating-linear-gradient`, animated `background-position` - reversed on bottom/left so the ants circle the range. And add a frozen-pane test - `CopyMockContext` has no frozen rows/columns, so this path is untested.

## 3. Cascade the clipboard instead of injecting RadzenSpreadsheet

No other overlay takes the parent component. The workaround exists because `[Parameter]` must be public and `SpreadsheetClipboard` is internal - but `[CascadingParameter]` doesn't have to be public. Wrap the grid content in `<CascadingValue Value=@clipboard IsFixed="true">` and declare `[CascadingParameter] internal SpreadsheetClipboard Clipboard` in `CopyOverlay`. The `Spreadsheet` parameter goes away, `clipboard` returns to `private`, and the tests can cascade a bare `SpreadsheetClipboard` instead of newing up a `RadzenSpreadsheet`.

## 4. Escape while editing should not clear the marquee

In Excel, Escape during an edit cancels the edit and keeps the ants; a second Escape clears them. Early-return from `CancelEditAsync` after `Editor.Cancel()`. Note `global: true` also makes the JS `preventDefault()` Escape anywhere inside the component - please give the filter and cell-menu popups a manual pass.

## 5. A no-op Undo should not clear the clipboard

Repro: Ctrl+C a formula cell, Ctrl+Z with nothing to undo, Ctrl+V - the ants vanish and the formula pastes as literal text (cleared clipboard → `TryPaste` misses its CSV match → plain-text fallback). Clear only when `CanUndo`/`CanRedo` was true.

## Nits

- Three members named Clear in `SpreadsheetClipboard` (`Clear`, `ClearInternal`, static `Clear(Worksheet, RangeRef)`) - collapse the first two, rename the third.
- `IsStructuralCommand` duplicates `DescribeCommand`'s type list and can be `static`.
- Make `Changed` an `Action` like `Selection.Changed`; the handler becomes a direct `StateHasChanged`.
- Keep at most one of `translateZ(0)` / `will-change`, and move the new CSS next to the other overlay styles.
- Missing space in `Bind("Escape", _ => CancelEditAsync(),global:true)`; trailing whitespace; five files without trailing newline; drop the narration comments in the tests.
- Test gap: nothing verifies the overlay re-renders when `Changed` fires after the initial render.

One question to settle: structural commands clear the ants but cell edits don't - worth checking Excel's behavior there before merging.

panoskentros:
For the question that needs to be settled: I verified that Excel does in fact clear the marching ants for both structural commands and cell edits - and that includes adjusting cell styles, for example making the text bold.  The ants only survive if resizing columns. Im testing Excel online on MacOs (im not sure if that makes a difference) without an access of a windows computer. May I continue the implementation with this logic in mind? Here is a quick demo:


https://github.com/user-attachments/assets/b06fb682-f5b5-44a4-aca0-83e3a418b4a9

akorchev:
Thanks for verifying - and your observation holds up. I tested Windows desktop Excel and here is what happened:

| Operation | Marching ants |
|---|---|
| Cell edit (typed), bold, number format, border, merge, sort, autofill, autofilter | cleared |
| Insert row, delete row | cleared |
| Column width via the Format dialog | cleared |
| Column resize by drag, autofit | **kept** |
| Hide / unhide column | **kept** |
| Shape move / resize | **kept** |
| Undo - both with an empty stack and undoing a real edit | **kept** |

So desktop matches your recording, with two refinements:

- The rule isn't "resize is special" - it's direct-manipulation gestures vs commands. The same width change through the Format dialog clears the ants; the drag keeps them. Hiding a column also keeps them.
- Excel never clears the ants on undo - please remove the `clipboard.Clear()` calls from `Undo()`/`Redo()` entirely (this supersedes my earlier point 5).

Please continue with this logic. Since Excel clears on almost everything, keeping the `IsStructuralCommand` allowlist would mean adding ~25 more command types to it, and every future command would silently keep the ants until someone remembers to list it. Flip it: clear by default, enumerate the few keepers. Every mutation funnels through `ExecuteAsync`, so it's one condition:

```csharp
if (executed && command is not (PasteCommand
    or ResizeColumnCommand or ResizeRowCommand
    or MoveAnchoredCommand<SheetImage> or MoveAnchoredCommand<SheetChart>
    or ResizeAnchoredCommand<SheetImage> or ResizeAnchoredCommand<SheetChart>))
{
    clipboard.Clear();
}
```

Why each exception:

- `PasteCommand` - not from the Excel test but from the feature's own spec: copy once, paste many times. Clearing here would make the second Ctrl+V stop working. Cut-paste needs no special case - the clipboard already clears itself after a move.
- `ResizeColumnCommand` / `ResizeRowCommand` - what the resize drag handle and autofit produce; the "kept" rows in the table.
- `MoveAnchoredCommand` / `ResizeAnchoredCommand` - image/chart move and resize; kept in the test, and they don't touch cells.

Notes:

- Hide/unhide column calls `Worksheet.Columns.Hide()` directly and never reaches `ExecuteAsync`, so it matches Excel's "kept" behavior with no code - don't go looking for a command to add to the list.
- The default branch automatically covers cell edits (`AcceptEditCommand`), formatting (`FormatCommand`, `BorderCommand`) and every future command, which the test says must clear.
- `IsStructuralCommand` goes away, which also resolves the earlier nit about it duplicating `DescribeCommand`'s type list.

panoskentros:
Quick question...does it show any trailing whitespaces now or no newlines at the end of files? I did change some settings on Rider (the IDE im using) to prevent these errors. Also did a git diff check and it doesn't show anything to me.

PR #2666, opened 2026-08-18
Fix range navigator labels to respect date format while dragging

wimsoetens-cmd:
Fix range navigator labels to respect date format while dragging


enchev:
Thanks for the contribution! The idea is right, but the sequential replace calls break formats with month names: with "{0:MMM dd, yyyy}" (the example from the HandleLabelFormatString docs) the drag label shows 044 19, 2025 - MMM gets consumed as MM plus a leftover M. Literal text also gets corrupted ({0:dd 'days'} -> '19ays'), and yy, hh, tt aren't handled.

Could you switch to a single-pass tokenizer instead? Something like:

`format.replace(/yyyy|yy|MMMM|MMM|MM|M|dd|d|HH|mm|ss/g, token => { ... })`

with month names for MMM/MMMM (English is fine - the server side formats with InvariantCulture). That way earlier replacements can't be re-matched by later ones.

Issue #2665, opened 2026-08-18
Feature request: RadzenMarkdownEditor — markdown input with toolbar and preview


artnim:
**Is your feature request related to a problem? Please describe.**

Radzen has a great story for *rendering* markdown since `RadzenMarkdown` (#2039), but no way to let users *author* it. For user-generated content (comments, descriptions, notes) `RadzenHtmlEditor` is currently the only rich-text input — which means storing HTML and carrying the full XSS sanitization burden on every write and read path. Plain markdown as the storage format avoids that class of problems entirely, but a bare `RadzenTextArea` is a poor authoring experience for users who don't know markdown syntax.

**Describe the solution you'd like**

A `RadzenMarkdownEditor` component in the spirit of the GitHub comment editor:

- **Write / Preview tabs** — Write is a plain `textarea` bound to the raw markdown string (`@bind-Value`), Preview renders it via the existing `RadzenMarkdown`, so both sides share the same parser and there is no renderer mismatch.
- **Toolbar** that inserts or wraps markdown syntax around the current selection: bold, italic, strikethrough, heading, link, image, inline code / code block, quote, ordered/unordered/task list, table. Composable toolbar sub-components following the `RadzenHtmlEditor*` pattern, so consumers can pick tools and add custom ones.
- **Keyboard shortcuts** matching the HtmlEditor where they map (Ctrl+B, Ctrl+I, Ctrl+K…).
- Familiar form-component surface: `Value`/`ValueChanged`/`Change`, `Disabled`, `Placeholder`, works inside `RadzenTemplateForm`/`RadzenFormField`.

Implementation-wise this is much lighter than the HtmlEditor: no `contenteditable`, just small JS interop for `selectionStart`/`selectionEnd` manipulation on the textarea. Preview reuses `RadzenMarkdown` as-is, keeping the component dependency-free like the rest of the suite.

**Describe alternatives you've considered**

- `RadzenHtmlEditor` — works, but forces HTML storage and sanitization on both write and read paths.
- Third-party Blazor markdown editors — these wrap JS libraries (EasyMDE/SimpleMDE/Toast UI), pulling in JS dependencies and a second markdown dialect that doesn't match `RadzenMarkdown`'s parser. As far as I can tell, no major Blazor component suite ships a native, dependency-free markdown editor — this would be a differentiator for Radzen.

**Additional context**

[...]

PR #2680, opened 2026-08-28
Add RadzenMarkdownEditor component

artnim:
## RadzenMarkdownEditor

Implements #2665.

Adds `RadzenMarkdownEditor`, a Markdown editor with a WYSIWYG **Design** mode and a raw-text **Source** mode, built on the existing `RadzenMarkdown` parser.

- **Modes**: `Design` (default) edits the rendered document in a `contenteditable` surface and round-trips it to Markdown; `Source` edits the Markdown text in a textarea. `Mode` is bindable and a built-in switcher toggles between the two (`ShowModeSwitch`).
- **Toolbar** following the `RadzenHtmlEditor` pattern: default tool set when no child content, fully composable otherwise (`RadzenMarkdownEditorUndo`, `…Redo`, `…Bold`, `…Italic`, `…Strikethrough`, `…FormatBlock`, `…Quote`, `…Code`, `…CodeBlock`, `…UnorderedList`, `…OrderedList`, `…TaskList`, `…Link`, `…Image`, `…HorizontalRule`, `…Separator`, `…CustomTool`). `…FormatBlock` is a level picker (Normal, Heading 1 to 6) modelled on `RadzenHtmlEditorFormatBlock`; it follows the caret and is disabled while the selection holds no paragraph or heading. Tools reflect the selection state (active format, undo/redo availability).
- **Commands** are applied to the DOM in Design mode and computed in C# (`MarkdownFormatter`, unit-tested) in Source mode, with the same normalization in both: whitespace is trimmed off the selection, a caret expands to its word, a selection that is already formatted has the format removed from its words (splitting the token when it contains more), otherwise it is wrapped once per block/line.
- **Undo/redo** is a component-owned history with toolbar tools and Ctrl/Cmd+Z, Ctrl/Cmd+Shift+Z, Ctrl/Cmd+Y.
- **Parser additions**: GFM strikethrough (`~~text~~`, `Strikethrough` node, `INodeVisitor.VisitStrikethrough` with a default implementation) and task lists (`ListItem.Checked`), rendered by `RadzenMarkdown` as well.
- **Paste** of rich content is laundered through Markdown so only supported constructs survive; `:shortcode:` emoji are expanded while typing.
- `Execute` event + `ExecuteCommandAsync` for custom tools, shortcuts via the same key-string scheme as `RadzenHtmlEditor`, `Immediate`/`Input`/`Change` semantics matching `RadzenTextArea`.
- Localized strings (en, de, es, fr, it, ja), styles for all themes via the existing `--rz-editor-*` tokens, demos and docs entry.

**Note:** Design-mode edits reserialize the whole value to canonical Markdown (`__bold__` becomes `**bold**`, `snake_case` becomes `snake\_case`); the rendered result is unchanged. This is documented on the component and the demo page.

Out of scope for now (happy to follow up): image upload, Tab-indent / list continuation.

**Theme notes:** styles use the existing `--rz-editor-*` tokens; the premium theme CSS under `RadzenBlazorDemos/wwwroot/css` is not regenerated by this PR (maintainer process), so the demo shows the new styles under the non-premium themes until the next "Update premium themes".

### DIFF
--- a/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
+++ b/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
@@ -40,4 +40,13 @@ public class HtmlVisitorTests
             "<ul><li><p>a</p></li><li><p>b</p></li></ul>",
             HtmlVisitor.ToHtml("- a\n\n- b"));
     }
+
+    [Theory]
+    [InlineData("[x](javascript:alert(1))", "<p><a href=\"\">x</a></p>")]
+    [InlineData("[x](JaVaScRiPt:alert(1))", "<p><a href=\"\">x</a></p>")]
+    [InlineData("![a](javascript:alert(1))", "<p><img src=\"\" alt=\"a\"></p>")]
+    public void ToHtml_blanks_dangerous_urls(string markdown, string expected)
+    {
+        Assert.Equal(expected, HtmlVisitor.ToHtml(markdown));
+    }
 }
--- a/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
+++ b/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
@@ -1,4 +1,3 @@
-using Radzen.Blazor.Documents.Markdown;
 using Xunit;
 
 namespace Radzen.Documents.Markdown.Tests;
--- a/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
+++ b/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
@@ -1,3 +1,4 @@
+using Radzen.Blazor.Documents.Markdown;
 using Xunit;
 
 namespace Radzen.Documents.Markdown.Tests;
--- a/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
+++ b/Radzen.Blazor.Tests/Markdown/HtmlVisitorTests.cs
@@ -1,0 +1,43 @@
+using Xunit;
+
+namespace Radzen.Documents.Markdown.Tests;
+
+public class HtmlVisitorTests
+{
+    [Theory]
+    [InlineData("# Hi", "<h1>Hi</h1>")]
+    [InlineData("plain *em* **strong** ~~del~~ `code`",
+        "<p>plain <em>em</em> <strong>strong</strong> <del>del</del> <code>code</code></p>")]
+    [InlineData("[link](https://a.b \"t\")", "<p><a href=\"https://a.b\">link</a></p>")]
+    [InlineData("![alt](img.png)", "<p><img src=\"img.png\" alt=\"alt\"></p>")]
+    [InlineData("> quote", "<blockquote><p>quote</p></blockquote>")]
+    [InlineData("- a\n- b", "<ul><li>a</li><li>b</li></ul>")]
+    [InlineData("3. a", "<ol start=\"3\"><li>a</li></ol>")]
+    [InlineData("- [x] done", "<ul><li><input type=\"checkbox\" checked> done</li></ul>")]
+    [InlineData("```csharp\nvar x = 1;\n```", "<pre><code class=\"language-csharp\">var x = 1;\n</code></pre>")]
+    [InlineData("---", "<hr>")]
+    [InlineData("a  \nb", "<p>a<br>b</p>")]
+    [InlineData("a\nb", "<p>a b</p>")]
+    [InlineData("<script>x</script>", "<p>&lt;script&gt;x&lt;/script&gt;</p>")]
+    [InlineData("a < b & c", "<p>a &lt; b &amp; c</p>")]
+    [InlineData("| a | b |\n| --- | ---: |\n| 1 | 2 |",
+        "<table><thead><tr><th>a</th><th style=\"text-align:right\">b</th></tr></thead><tbody><tr><td>1</td><td style=\"text-align:right\">2</td></tr></tbody></table>")]
+    public void ToHtml_renders(string markdown, string expected)
+    {
+        Assert.Equal(expected, HtmlVisitor.ToHtml(markdown));
+    }
+
+    [Fact]
+    public void ToHtml_separates_blocks_without_text_between_them()
+    {
+        Assert.Equal("<h1>A</h1><p>b</p>", HtmlVisitor.ToHtml("# A\n\nb"));
+    }
+
+    [Fact]
+    public void ToHtml_wraps_loose_list_items_in_paragraphs()
+    {
+        Assert.Equal(
+            "<ul><li><p>a</p></li><li><p>b</p></li></ul>",
+            HtmlVisitor.ToHtml("- a\n\n- b"));
+    }
+}
--- a/Radzen.Blazor.Tests/Markdown/InlineSpanTests.cs
+++ b/Radzen.Blazor.Tests/Markdown/InlineSpanTests.cs
@@ -1,0 +1,52 @@
+using Xunit;
+
+namespace Radzen.Documents.Markdown.Tests;
+
+public class InlineSpanTests
+{
+    [Fact]
+    public void Finds_strong_span()
+    {
+        var spans = InlineParser.ScanSpans("a **bold** b");
+        var span = Assert.Single(spans);
+        Assert.Equal((2, 10, 2, '*'), (span.Start, span.End, span.DelimiterLength, span.Char));
+    }
+
+    [Fact]
+    public void Finds_underscore_emphasis()
+    {
+        var spans = InlineParser.ScanSpans("_it_");
+        var span = Assert.Single(spans);
+        Assert.Equal((0, 4, 1, '_'), (span.Start, span.End, span.DelimiterLength, span.Char));
+    }
+
+    [Fact]
+    public void Finds_nested_strong_and_emphasis()
+    {
+        var spans = InlineParser.ScanSpans("***a***");
+        Assert.Contains(spans, s => s is { Start: 1, End: 6, DelimiterLength: 2 }); // **a** inner
+        Assert.Contains(spans, s => s is { Start: 0, End: 7, DelimiterLength: 1 }); // *…* outer
+    }
+
+    [Fact]
+    public void Finds_code_span_and_ignores_emphasis_inside()
+    {
+        var spans = InlineParser.ScanSpans("x `a *b*` y");
+        var span = Assert.Single(spans);
+        Assert.Equal((2, 9, 1, '`'), (span.Start, span.End, span.DelimiterLength, span.Char));
+    }
+
+    [Fact]
+    public void Finds_strikethrough()
+    {
+        var spans = InlineParser.ScanSpans("~~gone~~");
+        var span = Assert.Single(spans);
+        Assert.Equal((0, 8, 2, '~'), (span.Start, span.End, span.DelimiterLength, span.Char));
+    }
+
+    [Fact]
+    public void Flanking_rejects_space_before_closer()
+    {
+        Assert.Empty(InlineParser.ScanSpans("**hello **world"));
+    }
+}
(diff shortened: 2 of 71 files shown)


enchev:
Thanks @artnim! In my opinion the textarea should use Blazor's two-way binding instead of raw `value=@Value @oninput=@OnInput`:

```razor
<textarea @ref=textarea ... @bind:get="@Value" @bind:set="@OnInputAsync" @bind:event="oninput" @onchange=@OnChange></textarea>
```

The raw pattern re-writes the DOM value on every server round-trip, which clobbers fast typing on Blazor Server (typing "hello world" quickly produces garbled text) — same issue we fixed in #2453. `@bind:get/@bind:set` uses the framework's echo suppression and avoids it. See `RadzenTextArea.razor` for reference.

artnim:
Good catch, thanks @enchev — done in 45d4ec9ef. The textarea now uses `@bind:get="@Value" @bind:set="@OnInputAsync" @bind:event="oninput"` (plus the separate `@onchange` for the `Change` event), matching `RadzenTextArea`. `OnInputAsync` takes the bound `string?` directly instead of `ChangeEventArgs`.

enchev:
Thank you!. After reviewing it again we have new comments:

1. The toolbar commands are not reliable for real-world selections. The formatter handles only clean, exact selections. A few examples we reproduced:

- Selecting text with a trailing space (common with mouse drags) and clicking Bold produces **hello **world — CommonMark's flanking rules reject the closing ** after a space, so it renders as literal asterisks.
- Caret inside a word with no selection inserts **** mid-word instead of expanding to the word (what GitHub does).
- A partial selection inside existing **bold** produces broken output like hello **w**orl**d**.
- Star emphasis (*text*) isn't recognized by the underscore-based italic toggle, so toggling pasted text nests instead of unwraps.
- There is no first-class undo/redo support — RadzenHtmlEditor ships dedicated Undo/Redo toolbar tools, while here undo relies on the deprecated execCommand keeping the native history alive, and the fallback path sets the textarea value directly, which destroys the browser's undo stack entirely.

Making this robust requires GitHub-style selection normalization — trimming edge whitespace, expanding the caret to word boundaries, and detecting enclosing/overlapping tokens for both * and _ — plus an undo/redo implementation that doesn't depend on execCommand.

2. Also the direction we want for a markdown editor is a design-time (WYSIWYG) editing experience — similar to the Discourse composer on our forum, where you edit the rendered content directly and it round-trips to markdown, with a source-mode toggle and proper undo/redo. A textarea with syntax buttons can at best be the "source mode" half of that component.

artnim:
Thanks for the detailed review @enchev! Both points are addressed — the branch now implements the design-time (WYSIWYG) direction, and the source-mode commands got the full normalization treatment. In detail:

**1. Selection normalization — all five repros are fixed and pinned as unit tests (verbatim) in `MarkdownFormatterTests`:**

- *Trailing-space selection → `**hello **world`*: edge whitespace is trimmed off the selection before wrapping, so delimiters always land flanking-valid.
- *Caret in a word → `****`*: a collapsed caret now expands to the surrounding word, GitHub-style (caret in whitespace still inserts empty delimiters with the caret between them).
- *Partial selection inside `**bold**`*: enclosing tokens are detected and unwrapped as a whole; overlapping tokens inside a larger selection are stripped before wrapping once.
- *`*text*` vs `_text_`*: both delimiters are recognized when unwrapping. Emission switched to `**`/`*` (CommonMark forbids intraword `_`, which word-expansion requires — and it matches what the design-mode serializer emits).
- *Undo/redo*: the editor now owns its history — a JS-side snapshot stack shared across both modes, with dedicated `RadzenMarkdownEditorUndo`/`Redo` toolbar tools (enabled state reported live), Ctrl+Z / Ctrl+Shift+Z / Ctrl+Y interception, and typing-burst coalescing for word-level granularity. `execCommand` is gone from the component entirely.

Rather than regex heuristics, the normalization uses the real parser: `InlineParser` now records inline token source spans during its delimiter-stack pass, so flanking rules come from the same code that renders the markdown.

**2. Design-time editing** is now the primary mode (`MarkdownEditorMode.Design`, the default; Write/Preview/Split are gone). The architecture, briefly:

- `Value` stays markdown and remains the single source of truth. Markdown → HTML goes through a new `HtmlVisitor` over the existing AST; the contenteditable DOM serializes back to canonical markdown on every (debounced) edit, so mode switching has no separate sync step.
- Editing commands use the Selection/Range API; the serializer normalizes whatever contenteditable produces (`<div>`/`<b>`/styled spans). Pasted rich content is laundered through markdown → canonical HTML, parsed inertly so clipboard HTML can't execute.
- Toolbar buttons reflect the caret's active formats; `:emoji:` shortcodes replace live while typing; the mode switch is an icon toggle (with a `ShowModeSwitch` parameter).
- Since design-mode HTML is always AST-built, the `AllowHtml`/`AllowedHtmlTags`/`AllowedHtmlAttributes` XSS surface no longer exists and those parameters are removed; raw HTML in markdown renders as escaped text, and `javascript:` URLs are blanked at render.
- The parser gained GFM strikethrough and task lists to cover the GitHub-comment subset — note this adds `VisitStrikethrough` to the public `INodeVisitor` interface.

Round-tripping is contract-tested (`serialize(render(md)) === md` for canonical markdown, byte-exact over a corpus covering the whole subset) plus ~100 new unit/bunit tests; the demo page is reworked design-first.

One heads-up for merging: the premium theme CSS bundles predate `_markdown-editor.scss`, so Material 3/Fluent will show the design surface unstyled until they're recompiled on your side — the free themes pick the styles up from source.

akorchev:
Thanks for the updates - this will be a great addition!

 A few things before we can merge:

**Selection lost after removing bold/italic (Design mode).** The unwrap branch of `toggleInline` moves the children out of the `strong`/`em` and returns without restoring the selection, so the word ends up unselected. Re-select the moved span before returning:

```js
var first = node.firstChild, last = node.lastChild;
while (node.firstChild) parent.insertBefore(node.firstChild, node);
parent.removeChild(node);
if (first && last) { range.setStartBefore(first); range.setEndAfter(last); setRange(range); }
```

Also, after unbolding the Bold tool stays highlighted, so the tool state is not refreshed after the command.

Related: the unwrap fires when the selection is anywhere inside the element, so selecting one word in a bold sentence and pressing Ctrl+B unbolds the whole sentence. Please decide what the behaviour should be for a partial selection and a collapsed caret, and apply the same rule in `MarkdownFormatter` for Source mode.

**Other issues**

- Select All + Heading turns every top-level block into `<h1>`, including `<ul>`, `<table>` and `<pre>`, which destroys lists, tables and code blocks. Heading should only touch paragraphs and headings.
- Select All + Bold wraps all blocks in one `<strong>` and `extractContents` leaves empty shells of the partially selected blocks behind. The serialized value becomes `# ` followed by the whole document on one bold line and an empty code fence. Inline toggles should be limited to a single block, or applied per block.
- `reportState` is bound to `document` `selectionchange` and calls `OnToolStateAsync` (which calls `StateHasChanged`) for every selection change anywhere on the page. Skip the interop call when the selection is outside the editor or the state has not changed.
- Ctrl+Z/Ctrl+Y are intercepted on the textarea too, replacing native undo in Source mode.
- Adding `VisitStrikethrough` to `INodeVisitor` breaks external implementers. Consider a default interface implementation.
- Design-mode edits reserialize the whole value: `snake_case` -> `snake\_case`, `array[0]` -> `array\[0\]`, `__bold__` -> `**bold**`. Worth documenting.

**Housekeeping**

- The description is out of date (Edit/Preview/Split, `execCommand`). Please update it to match the Design/Source implementation.
- Please remove the narrative `//` comments (about 90 across the C#, JS and SCSS). We only keep XML docs on public API.
- es/fr/it/ja localizations have 4 of the 21 new strings; either complete them or leave them out.
- `MarkdownEditorButton.razor` is a near copy of `EditorButton.razor`.

akorchev:
Thanks, the selection and Select All fixes check out in the browser!

Two bigger things before merging. They touch the core of the implementation, so if you would rather not go another round I am happy to take the branch from here and finish it. Just say so, and I will merge it with your commits intact and do the rest on top.

**Timers.** The design surface relies on three of them and we cannot ship that:

- the 250 ms `inputTimeout` debounce on value sync (plus the `dirty` flag and blur flush that exist to cover it)
- the 100 ms `stateTimeout` debounce on tool state
- the `Date.now()` / `lastInputTime` coalescing in `snapshot`, which makes undo granularity depend on typing speed

**Global selectionchange.** `reportState` is registered on `document` for the lifetime of every editor instance, so every selection change anywhere on the page runs the handler for each editor. The unchanged-state check only saves the interop call, not the handler.

Please make all of it event driven and scoped to the editor, matching what `createEditor` does for the HtmlEditor:

- Value: call `notifyValue()` directly from the `input` handler and drop the timer, the `dirty` flag and the blur flush. Blur then only raises `Change`, like `RadzenTextArea`. Per-keystroke interop is what the HtmlEditor already does; `lastSurfaceValue` is what keeps the echo out of the surface, not the debounce.
- Tool state: report directly from the `selectionchange` handler, attach the listener on focus and remove it on blur, keep the unchanged-state check, and call `reportState()` at the end of `execute`, `undo` and `redo`.
- History: coalesce on `beforeinput` boundaries instead of elapsed time: a switch between inserting and deleting, `insertParagraph`, `insertFromPaste`, `insertFromDrop`, whitespace in `e.data` (one word per undo step, like a native textarea), and always before a toolbar command.

And two smaller ones: the four remaining `//` comments in the JS, and the fact that `apply` assigning `textarea.value` is what discards the native undo stack in Source mode (`execCommand('insertText')` would keep it, if we ever want to stop intercepting Ctrl+Z there).

artnim:
Thanks for the second look. All of it is on the branch now (head b1fd42829).

Your latest points:

- Timers: none left. The value is reported from the input handler on every keystroke; the debounce, the dirty flag and the blur flush are gone. Blur only raises Change, and only when the content differs from the content at focus (what a textarea does).
- Tool state: reported straight from selectionchange, whose listener is attached when the surface or the textarea gets focus and removed on blur, so it never runs for other editors or the rest of the page. The unchanged-state check stays, and execute, undo, redo and apply report at their end.
- History: undo steps are cut on beforeinput boundaries only (insert/delete switch, insertParagraph, insertFromPaste, insertFromDrop, typed whitespace, and always before a command). No Date.now() anywhere.
- The remaining // comments are removed.
- apply still assigns textarea.value since Ctrl+Z stays intercepted in Source mode (as enchev asked for a history that doesn't depend on execCommand). If you'd rather keep the native stack intact as well, switching apply to execCommand('insertText') is a one-line change.

Two things that surfaced while testing the selection fixes, also on the branch:

- The cycling Heading button is replaced by RadzenMarkdownEditorFormatBlock, a level picker (Normal, Heading 1 to 6) modelled on RadzenHtmlEditorFormatBlock and reusing its strings. Cycling per block was ambiguous on mixed selections; the picker is explicit and closer to the HtmlEditor. It follows the caret in both modes and is disabled while the selection contains no paragraph or heading. To make that work, EditorDropDown got a Disabled parameter (so it opens without an HtmlEditor cascading in) and now drops its selected item when the value no longer matches an item; the second part also applies to the HtmlEditor dropdowns, where a stale label could previously linger.
- The design surface is normalized before every command; without it, repeated inline toggles on a whole-document selection left empty text nodes and shells behind and corrupted the blocks.

Heading in Source mode now also only touches paragraph and heading lines, matching Design.

akorchev:
Thank you for the work on this, and for the quick turnarounds on every round. The parser additions, the formatter, the tools and the tests are solid, and the editor is something we want in the library.

We have decided to merge the PR into the `v12` branch rather than `master`, and to change the Design mode implementation before it ships. Two things drove that decision:

- The design surface converts DOM to Markdown and back, which is lossy by nature, and every edit reserializes the whole document. We want untouched Markdown to survive byte for byte.
- Source mode tool state adds a second, regex-based Markdown recognizer next to the parser, and the two will drift.

The plan is to keep the Markdown text as the single source of truth in both modes: the parser records source positions, the rendered view maps DOM positions back to source offsets, and every edit, including typing, toolbar commands and undo, becomes a text edit computed once in .NET through `MarkdownFormatter`. That removes the serializer, the DOM command code and the second parser. Your parser changes, `MarkdownFormatter`, `HtmlVisitor`, the tool components, the public API, the localization and the tests carry over as they are.

Thanks again, this gave us a much better starting point than we would have had otherwise.

We expect the v12 branch to ship by the end of the month.

PR #2687, opened 2026-08-31
Selection membership: one cache for CheckBoxList, ChipList, SelectBar, DropDown and Tree

pianomanjh:
## Summary

- Cache compiled item-property getters by runtime type and property name.
- Reuse selection membership during a render in CheckBoxList, ChipList, SelectBar, and DropDown.
- Share checked-value membership across Tree items and compute parent tri-state in one subtree walk.
- Invalidate render-scoped state at component boundaries so in-place collection mutations remain visible.
- Hold type metadata weakly and bound item-property entries to avoid process-lifetime growth.

## Behavior and compatibility

The fast path is used only when it preserves the collection's existing equality semantics. Custom collection contracts are still asked directly, wrappers retain their forwarding behavior, and null handling remains consistent with the original binding. If a closed delegate cannot be created under trimming or AOT, reflective invocation preserves the same answer.

Custom list-like `ICollection<T>` implementations may remain linear when their equality behavior cannot be reproduced safely. This is the pre-existing behavior and avoids trading correctness for caching.

The caches do not retain components or collection instances. Runtime types are weakly keyed, and each runtime item type stores at most 64 property decisions. Collectible-type tests cover both successful and failed property lookups as well as collection-policy metadata.

## Performance

Measured with a temporary BenchmarkDotNet/bUnit harness outside this branch. Each operation creates a test render context and performs a full first render. List controls use 1,000 items and 200 selected values. Tree uses 100 expanded roots with 10 children each and 200 checked descendants. RadioButtonList is included because it uses the shared property getter changed by this PR.

| Component | `master` mean | PR mean | Time | `master` allocated | PR allocated | Allocation |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| CheckBoxList | 14.178 ms | 6.778 ms | **-52.2%** | 25.76 MB | 8.90 MB | **-65.5%** |
| ChipList | 26.969 ms | 22.701 ms | **-15.8%** | 30.49 MB | 26.29 MB | **-13.8%** |
| SelectBar | 15.616 ms | 10.970 ms | **-29.8%** | 31.45 MB | 23.06 MB | **-26.7%** |
| DropDown | 6.856 ms | 6.991 ms | +2.0% | 11.36 MB | 11.36 MB | 0.0% |
| RadioButtonList | 6.586 ms | 6.437 ms | -2.3% | 8.81 MB | 8.75 MB | -0.7% |
| Tree (expanded hierarchy) | 41.182 ms | 37.115 ms | **-9.9%** | 35.10 MB | 33.82 MB | **-3.6%** |

DropDown and RadioButtonList are effectively neutral within the observed variance. DropDown's change primarily fixes render-scoped invalidation; its existing fast path already handled the measured first-render case.

Environment: Apple M4, .NET 10.0.11, BenchmarkDotNet 0.14.0, 3 warmups and 10 measured iterations. Benchmark infrastructure is intentionally not included in this PR.

## Validation

- 94 focused membership and DropDown tests pass.
- Full suite: 5,054 passed, 0 failed.
- Library builds for .NET 8, .NET 9, and .NET 10 with 0 warnings and 0 errors.
- `git diff --check` passes.

<details>
  <summary>AI Toolchain</summary>

  VSCode, Claude Code CLI - [Opus 5 - high], reviews from Codex - [5.6 Sol - High]
</details>


enchev:
Thanks for the pull request Here is what I've found:

- A type that overrides Equals without a consistent GetHashCode now loses its selection highlight in CheckBoxList/SelectBar/ChipList/Tree (master scanned with Equals only; the HashSet memo requires hashing). Please memoize only when the element type's hash contract is safe and fall back to the linear scan otherwise.
- Collections with custom comparers (e.g. HashSet<string>(StringComparer.OrdinalIgnoreCase)) are now asked through their typed ICollection<T> contract, so items get selected that weren't before. Please keep master's dispatch (ICollection<object> only, default equality otherwise) so released behavior is unchanged.

PR #2689, opened 2026-09-01
RadzenDataGrid: async IQueryable execution for EF Core

### LINKED ISSUE #2688: Support asynchronous execution of Entity Framework queries

When an `IQueryable<T>` backed by an asynchronous remote provider such as Entity Framework Core is bound directly to a Radzen data component, Radzen composes the query correctly but materializes it with synchronous operations such as `Count()`, `ToList()`, and `FirstOrDefault()`.

For a database-backed query, those calls perform blocking I/O on the Blazor renderer or request thread. This can be difficult to notice in development but becomes costly on production sites with concurrent users or slow queries: blocked threads reduce throughput, increase tail latency, and can contribute to thread-pool starvation. Overlapping reloads, filtering, virtualization, and keyboard lookups also need careful cancellation and serialization when a scoped `DbContext` is involved.

Applications can avoid the synchronous calls by implementing `LoadData`, but that requires each application to duplicate filtering, sorting, paging, counting, cancellation, and stale-result handling that the components otherwise provide through direct `IQueryable` binding.

Support true asynchronous execution for directly bound provider-backed queries while retaining the existing component APIs and query semantics.

The support should:

- allow an optional provider adapter to execute count and materialization operations asynchronously;
- keep provider-specific dependencies, such as Entity Framework Core, out of the core `Radzen.Blazor` package;
- safely coordinate cancellation, superseded loads, virtualization, and overlapping operations for each component;
- preserve the current synchronous behavior for unsupported providers and applications that do not register an adapter; and
- avoid adding work or allocations to ordinary synchronous component paths.

This would preserve the convenience and correctness of direct `IQueryable` binding while allowing database I/O to remain asynchronous end to end.

- Implementing `LoadData` in every application works, but repeats component query orchestration and gives up much of the value of direct query binding.
- Pre-materializing the data prevents provider-side filtering and paging and does not address virtualization.
- Wrapping synchronous queries in `Task.Run` does not make the underlying database operation asynchronous and can worsen thread-pool pressure.
- Referencing Entity Framework Core directly from `Radzen.Blazor` would introduce an unnecessary provider dependency; a small optional adapter is preferable.

**Additional context**

Issues #76 and #1943 show related async and EF-backed component execution concerns, but neither provides a general asynchronous path for directly bound `IQueryable` sources.

A proposed implementation, including an Entity Framework Core adapter, component demos, cancellation/concurrency tests, and allocation benchmarks, is available in #2689.

pianomanjh:
Binding an Entity Framework Core `IQueryable` directly to a Radzen data component lets the component compose filtering, sorting, and paging that EF can translate to SQL, but the existing path materializes those queries with synchronous `Count()` and `ToList()` calls. That blocks the calling thread on database I/O unless the application replaces normal data binding with a custom `LoadData` handler.

Closes #2688.

This PR adds an opt-in, provider-agnostic async query seam, modeled on Blazor QuickGrid, and wires it through Radzen's data-bound components:

```csharp
// Program.cs
builder.Services.AddRadzenQueryableEntityFrameworkAdapter();
```

With the adapter registered, supported queryables are counted and materialized with awaited provider operations. With no executor registered, an unsupported queryable, or an excluded component path, existing synchronous behavior is unchanged.

### Provider seam and adapter

- `IAsyncQueryExecutor` provides support detection, async count/materialization, and provider-specific fallback classification without introducing an Entity Framework dependency in `Radzen.Blazor`.
- `Radzen.Blazor.EntityFrameworkAdapter` supplies the EF Core implementation and `AddRadzenQueryableEntityFrameworkAdapter()` registration for net8.0, net9.0, and net10.0.
- EF-specific busy-`DbContext` handling stays within the adapter.

### Per-component lifecycle coordination

Each component lazily creates one internal, sealed `AsyncQueryCoordinator` after an executor accepts a query. The coordinator owns load generations, cancellation sources, virtual requests, FIFO provider serialization, typed count/materialization operations, and owner-only render notification. Synchronous and unsupported paths do not create it.

Readonly `LoadLease` and disposable `VirtualRequestLease` values scope ownership, cancellation, and deferred source disposal. The provider queue has an uncontended fast path, preserves FIFO ordering under contention, and converts synchronous executor throws into faulted tasks.

`RadzenComponent` caches executor resolution and forwards lifecycle operations to its nullable coordinator. `PagedDataBoundComponent` supplies the shared count/page/materialization policy.

### Component coverage

| Component | Async path |
| --- | --- |
| `RadzenDataGrid` | paged, unpaged, virtualized, virtualized with paging, grouped, and supported `object`/dynamic views |
| `RadzenDataList` | reload and virtualized windows |
| `RadzenPivotDataGrid` | its overridden reload path |
| `RadzenDropDown`, `RadzenListBox` | `DropDownBase` virtualization and out-of-window keyboard lookup |
| `RadzenDropDownDataGrid` | search/count/page loading |

For an `object`-typed grid, a one-row async query supplies the runtime element type needed for provider-side ordering.

The DataGrid IQueryable demo registers the EF adapter and presents paged and virtualized async-query examples followed by the opt-in setup.

The adapter's packaged README is release-facing documentation for installation, registration, supported components, fallback behavior, and the per-component `DbContext` concurrency boundary.

Async execution is not selected for `LoadData` callbacks, `QueryOnlyVisibleColumns`, self-referencing hierarchies, `object`-typed grids with an in-memory `SortComparer`, unpaged virtualized `object`-typed grids, or `RadzenGantt`'s hierarchy-wide internal loader. Gantt data that requires asynchronous I/O should be materialized before binding.

## Behavior and safety

- Superseding a load cancels its predecessor and waits for the whole load to exit before a replacement reads the same provider. Cancellation is advisory; serialization prevents a replacement from becoming the second concurrent operation on one `DbContext` when the old provider call ignores or delays cancellation.
- Pending-load suppression keeps render-time getters from synchronously reading a source already owned by an awaited query. Simultaneous superseders retain suppression until every wait exits.
- Reloads cancel an active virtual request before waiting for the provider. A replaced or disposed virtual request retains a valid token until its executor call has returned, then clears and disposes its linked source.
- Only cancellations requested by the component or Virtualize are swallowed. Provider-originated cancellation and other faults still propagate.
- If a provider supports asynchronous execution but rejects a composed query shape, its executor decides whether the existing synchronous fallback is safe.
- Unpaged materialization derives the total from returned rows; paged loads retain the count needed to clamp the page.

No existing parameters, rendered output, query composition semantics, fallback rules, or synchronous behavior are intentionally changed.

## Performance

Paired BenchmarkDotNet runs against the fixed baseline found lower construction and async-path allocations with no steady-state timing regressions. Direct tests also confirm that synchronous and unsupported paths do not create coordinator state.

<details>
  <summary>AI Toolchain</summary>

  VSCode, Claude Code CLI - [Opus 5 - high], reviews from Codex - [5.6 Sol - High]
</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,5 +25,7 @@ jobs:
     
     - name: Build
       run: dotnet build Radzen.Blazor/Radzen.Blazor.csproj
+    - name: Build Entity Framework adapter
+      run: dotnet build Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj
     - name: Test
       run: dotnet test Radzen.Blazor.Tests/Radzen.Blazor.Tests.csproj
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -81,6 +81,9 @@ jobs:
       - name: Test
         run: dotnet test Radzen.Blazor.Tests/Radzen.Blazor.Tests.csproj -c Release
 
+      - name: Pack Entity Framework adapter
+        run: dotnet pack Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj -c Release
+
       - name: Upload package artifact
         if: ${{ !inputs.publish }}
         uses: actions/upload-artifact@v4
@@ -89,7 +92,13 @@ jobs:
           path: |
             Radzen.Blazor/bin/Release/*.nupkg
             Radzen.Blazor/bin/Release/*.snupkg
+            Radzen.Blazor.EntityFrameworkAdapter/bin/Release/*.nupkg
+            Radzen.Blazor.EntityFrameworkAdapter/bin/Release/*.snupkg
 
       - name: Push to NuGet
         if: ${{ inputs.publish }}
         run: dotnet nuget push "Radzen.Blazor/bin/Release/*.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json --skip-duplicate
+
+      - name: Push Entity Framework adapter to NuGet
+        if: ${{ inputs.publish }}
+        run: dotnet nuget push "Radzen.Blazor.EntityFrameworkAdapter/bin/Release/*.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json --skip-duplicate
--- a/Radzen.Blazor.EntityFrameworkAdapter/README.md
+++ b/Radzen.Blazor.EntityFrameworkAdapter/README.md
@@ -1,0 +1,62 @@
+# Radzen.Blazor.EntityFrameworkAdapter
+
+Execute Entity Framework Core queries from Radzen data-bound components asynchronously.
+
+Radzen components compose filtering, sorting, paging, and virtualization over a bound `IQueryable`. This adapter lets supported Entity Framework queries use `CountAsync` and `ToListAsync` without requiring a custom `LoadData` handler. `Radzen.Blazor` remains provider-agnostic; this package supplies the EF Core implementation of `IAsyncQueryExecutor`.
+
+## Install
+
+Install the adapter version that matches your `Radzen.Blazor` version:
+
+```shell
+dotnet add package Radzen.Blazor.EntityFrameworkAdapter
+```
+
+The package targets .NET 8, .NET 9, and .NET 10.
+
+## Configure
+
+Register the adapter once in `Program.cs`:
+
+```csharp
+builder.Services.AddRadzenQueryableEntityFrameworkAdapter();
+```
+
+## Use
+
+Bind an EF Core `IQueryable` or `DbSet` normally:
+
+```razor
+<RadzenDataGrid Data="@db.Orders"
+                TItem="Order"
+                AllowPaging="true"
+                AllowSorting="true"
+                AllowFiltering="true">
+    <Columns>
+        <RadzenDataGridColumn TItem="Order" Property="Number" Title="Number" />
+        <RadzenDataGridColumn TItem="Order" Property="Customer.Name" Title="Customer" />
+    </Columns>
+</RadzenDataGrid>
+```
+
+The component continues to compose the query. Count and materialization operations are awaited through EF Core when the query provider supports asynchronous execution.
+
+## Supported components
+
+- `RadzenDataGrid`, including paging, grouping, and virtualization
+- `RadzenDataList`, including virtualization
+- `RadzenPivotDataGrid`
+- `RadzenDropDown` and `RadzenListBox` virtualization
+- `RadzenDropDownDataGrid` search and paging
+
+Existing synchronous behavior is preserved when the adapter is not registered or a query provider does not support asynchronous execution. Application-provided `LoadData` handlers continue to own their query execution.
+
+`RadzenGantt` retains its existing synchronous internal loader because it performs hierarchy-wide traversal in addition to paging. Materialize its data asynchronously before binding when the source requires asynchronous database access.
+
+## Cancellation and concurrency
+
+Each component serializes its provider operations. A newer load requests cancellation of a superseded load and waits for that load to exit before using the same provider again. Virtualized requests link the component lifetime with the request cancellation token.
+
+Coordination is per component. If multiple components or application services share one `DbContext`, they can still issue concurrent operations against that context. Follow EF Core lifetime guidance: avoid concurrent use of a `DbContext`, scope it to the appropriate unit of work, or create contexts with `IDbContextFactory<TContext>`.
+
+Provider-originated failures and cancellations continue to propagate. The adapter permits synchronous fallback only for query shapes EF Core cannot execute asynchronously; it does not retry against a busy `DbContext`.
--- a/Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj
+++ b/Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj
@@ -1,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
+    <Nullable>enable</Nullable>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <IsPackable>true</IsPackable>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageId>Radzen.Blazor.EntityFrameworkAdapter</PackageId>
+    <Description>Entity Framework Core adapter that lets data-bound Radzen.Blazor components execute their count and paging queries asynchronously.</Description>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PackageTags>blazor;radzen;entityframework;efcore;async</PackageTags>
+    <Version>11.2.8</Version>
+    <Copyright>Radzen Ltd.</Copyright>
+    <Authors>Radzen Ltd.</Authors>
+    <PackageProjectUrl>https://www.radzen.com</PackageProjectUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <RepositoryUrl>https://github.com/radzenhq/radzen-blazor</RepositoryUrl>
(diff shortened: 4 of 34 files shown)


pianomanjh:
## Async-query coordinator refactor

This PR adds a provider-agnostic async-query seam, a lazy per-component coordinator, and a production Entity Framework Core adapter. `LoadLease` binds typed operations and stale-owner checks to a whole load; `VirtualRequestLease` owns linked cancellation and deferred disposal; the provider queue serializes count, materialization, and count-plus-page operations without capturing delegates; and `PagedDataBoundComponent` supplies the common count/page/materialize policy used by normal reloads and Pivot.

The DataGrid IQueryable demo now registers the EF adapter and shows paged and virtualized async-query examples before the opt-in setup instructions. The packaged adapter README now provides release-ready installation, registration, coverage, fallback, and `DbContext` concurrency guidance.

`RadzenGantt` remains outside this seam because its internal loader traverses the full hierarchy rather than executing a count/page operation. Data requiring asynchronous I/O can be materialized before binding to Gantt.

### Verification

Commands run on the submitted tree:

```text
dotnet build Radzen.Blazor/Radzen.Blazor.csproj -c Release --no-restore --disable-build-servers -m:1
dotnet build Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj -c Release --no-restore --disable-build-servers -m:1
dotnet pack Radzen.Blazor.EntityFrameworkAdapter/Radzen.Blazor.EntityFrameworkAdapter.csproj -c Release --no-restore --disable-build-servers -m:1
dotnet test Radzen.Blazor.Tests/Radzen.Blazor.Tests.csproj --no-restore --disable-build-servers -m:1
dotnet test Radzen.Blazor.Tests/Radzen.Blazor.Tests.csproj -c Release --no-restore --disable-build-servers -m:1
dotnet test Radzen.Blazor.Tests/Radzen.Blazor.Tests.csproj -c Release --no-restore --disable-build-servers -m:1 --filter 'FullyQualifiedName~AsyncQueryCoordinatorTests|FullyQualifiedName~AsyncQueryExecutorTests|FullyQualifiedName~AsyncSeamRenderThreadTests|FullyQualifiedName~VirtualWindowTests|FullyQualifiedName~AsyncSeamProviderConcurrencyTests|FullyQualifiedName~EntityFrameworkAdapterTests'
```

Results:

- `Radzen.Blazor`: net8.0/net9.0/net10.0, 0 warnings, 0 errors
- `Radzen.Blazor.EntityFrameworkAdapter`: net8.0/net9.0/net10.0, 0 warnings, 0 errors
- adapter NuGet package: created successfully with the release README and all three target frameworks
- complete Release suite: **5,082 passed, 0 failed, 0 skipped**
- focused coordinator/component/EF-SQLite/render-thread/virtual-window set: **71 passed**
- direct coordinator suite: **14 passed**
- server-hosted demo: **clean Debug build**, page and CSS/JS assets smoke-tested over localhost
- GitHub Actions build: **passed in 3m36s** ([run 33462819385](https://github.com/pianomanjh/radzen-blazor/actions/runs/33462819385))

The test project still reports its pre-existing AngleSharp NU1902 advisory and nullable-context warnings; the production projects build cleanly.

### Benchmarks

Baseline: `90ef872bd70275d28f5a32f3d0f3283fe525d559`  
Submitted head: `a6116959beabc57b36fa5501f030fc11731c22af`

I reused PR #9 head `8d7da58e65555afb6f89ea4e269fb65448e0a8a5`'s BenchmarkDotNet `BenchmarkRenderer`/no-op JS pattern in temporary, untracked harnesses. No benchmark plumbing is committed.

The host became strongly multimodal during the adaptive job, so I stopped that run and used identical bounded `ShortRun` jobs for the fixed baseline and submitted tree. The two renderer-heavy cases were then repeated together. Commands (Apple M4, .NET 10.0.11):

```text
git worktree add --detach <baseline-worktree> 90ef872bd70275d28f5a32f3d0f3283fe525d559
dotnet restore gridbench-coordinator/CoordinatorBench.csproj --disable-build-servers
dotnet build gridbench-coordinator/CoordinatorBench.csproj -c Release --no-restore --disable-build-servers -m:1 -p:RunAnalyzers=false
dotnet run --project gridbench-coordinator/CoordinatorBench.csproj -c Release --no-build --no-restore -- --filter '*CoordinatorBench*' --job short --artifacts <results>/paired
dotnet run --project gridbench-coordinator/CoordinatorBench.csproj -c Release --no-build --no-restore -- --filter '*SyncRenderNoExecutor*' '*FirstAsyncLoad*' --job short --artifacts <results>/focused
```

Paired bounded results (the focused repeat supplies the two renderer-heavy rows):

| Case | Baseline mean | Head mean | Mean delta | Baseline alloc | Head alloc | Alloc delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `RadzenComponent` construction | 40.12 ns | 37.15 ns | -7.40% | 400 B | 352 B | -48 B |
| `RadzenDataGrid` construction | 176.01 ns | 153.86 ns | -12.58% | 3,064 B | 3,016 B | -48 B |
| sync render, no executor | 712.71 us | 732.29 us | +2.75% | 474,193 B | 474,761 B* | +568 B* |
| unsupported reload | 430.92 us | 348.57 us | -19.11% | 189,954 B | 189,948 B | -6 B |
| first async load | 733.11 us | 727.45 us | -0.77% | 491,278 B | 490,861 B | -417 B |
| warmed async reload | 392.22 us | 370.61 us | -5.51% | 199,040 B | 198,680 B | -360 B |
| virtualized async load | 258.82 us | 245.04 us | -5.32% | 25,819 B | 25,492 B | -327 B |

The paired means show no timing regressions beyond the 3% gate; improvements larger than 3% are retained. Construction, first-use, warmed, virtual, and unsupported allocations all improve.

`*` The renderer-sized no-executor allocation counter varied by about 0.1% between launches. The paired default job showed **474,245 B baseline / 474,114 B head** (-131 B); the bounded pair showed +568 B. Direct regression tests assert that the coordinator remains null. Component eligibility and executor presence are checked before query composition and provider probing. A later single-case run was thermally unstable (1.120 ms) and repeated the 474,761 B counter, so it is disclosed rather than substituted for the paired timing row. This is treated as process-wide renderer variance, not an async-path allocation.

enchev:
Thanks for the pull request however we believe that there is a better way to implement the async support - I'll push it later today.

enchev:
Thank you for this contribution! The async query seam, coordinator, and test suite from this PR landed in master with 75004e882, reworked so that no separate NuGet package or service registration is needed: Radzen.Blazor now detects queryables that implement IAsyncEnumerable<T> (which covers Entity Framework Core) and executes count and materialization asynchronously out of the box. Counting goes through a GroupBy-composed sequence query so the EF dependency could be dropped entirely, operations are serialized per query provider instance to keep components sharing a DbContext safe, and the Radzen.Blazor.DisableAsyncQueryExecution AppContext switch opts out. Also fixed on the way in: a dispatcher violation where filtering a virtualized DropDown/ListBox to zero results crashed the Blazor Server circuit (the debounce timer thread reached the new StateHasChanged in LoadItems through an unmarshalled RefreshDataAsync).

pianomanjh:
> Thank you for this contribution! The async query seam, coordinator, and test suite from this PR landed in master with [75004e8](https://github.com/radzenhq/radzen-blazor/commit/75004e882ecb8290a527083af661b1346c41f52e), reworked so that no separate NuGet package or service registration is needed: Radzen.Blazor now detects queryables that implement IAsyncEnumerable (which covers Entity Framework Core) and executes count and materialization asynchronously out of the box. Counting goes through a GroupBy-composed sequence query so the EF dependency could be dropped entirely, operations are serialized per query provider instance to keep components sharing a DbContext safe, and the Radzen.Blazor.DisableAsyncQueryExecution AppContext switch opts out. Also fixed on the way in: a dispatcher violation where filtering a virtualized DropDown/ListBox to zero results crashed the Blazor Server circuit (the debounce timer thread reached the new StateHasChanged in LoadItems through an unmarshalled RefreshDataAsync).

Good to see a version of this make it in!  Agree the extra package would have been a burden, and enabling this by default is a better end result for the user.   That's a clever way to try and sniff out the provider async-ness.

enchev:
We needed to revert this feature - reasons are listed in the revert comit: https://github.com/radzenhq/radzen-blazor/commit/b04a2c423143fc0663af8b50e6da4d429af2d8ed

pianomanjh:
> We needed to revert this feature - reasons are listed in the revert comit: [b04a2c4](https://github.com/radzenhq/radzen-blazor/commit/b04a2c423143fc0663af8b50e6da4d429af2d8ed)

Shame this one had to come back out, keeping to Blazor Server guidance with DbContext and component lifetimes is genuinely hard, and I understand it's sharp to have on by default. (I've certainly hit that error a few times!)

PR #2697, opened 2026-09-02
RadzenTimeline: Add process-tracker support

pianomanjh:
The timeline could not express a progress tracker: its line was a single pseudo-element spanning each item's whole box, so coloring one item painted half of the connector on either side of its point rather than a connector between two points.

Split the line into a leading and a trailing half, and add `RadzenTimelineItem.LineStyle` to style the connector running from an item to the next. A connector spans two items, so `LineStyle` marks only the item that owns it and the sibling combinator reaches the other half — which keeps DOM order the single source of truth for both halves and for the first/last caps, and makes a connector span a hidden item for free.

The halves stop at the edge of the point rather than meeting under its centre, so a point does not have to be opaque to hide the line. The severity styles are opaque, but a tinted fill a caller supplies need not be, and the line showed through one.

The split also replaces the first/last-child offset rules that used to stop the line short of the outer points: those fall out of the first item having no leading half and the last no trailing half, leaving one `--rz-timeline-point-offset` where there were eight align-items-specific rules. The line-position overrides collapse likewise, since the offsets never depended on Reverse.

Alongside it:

- Current marks the step in progress: `aria-current="step"`, and a ring drawn from the point's own color. Muted base points ring in a mid-scale grey, since a tint of those is invisible against the page. It is an outline rather than a box-shadow because every point carries an `rz-shadow-*` class, and those declare `box-shadow` with `!important`.
- `PointClas`s reaches the point element, for the genuinely bespoke.
- The point glyph size moves to `--rz-timeline-point-font-size` so a custom `--rz-timeline-point-size` can carry a matching glyph.

Rendering the nine existing orientation, `line-position`, `reverse` and `align-items` permutations before and after the line change is pixel-identical.

image

### DIFF
--- a/Radzen.Blazor.Tests/TimelineItemTests.cs
+++ b/Radzen.Blazor.Tests/TimelineItemTests.cs
@@ -192,5 +192,80 @@ namespace Radzen.Blazor.Tests
 
             Assert.Contains("check", component.Markup);
         }
+
+        [Fact]
+        public void TimelineItem_Appends_PointClass_ToPointElement()
+        {
+            using var ctx = new TestContext();
+            var component = ctx.RenderComponent<RadzenTimelineItem>(parameters =>
+            {
+                parameters.Add(p => p.PointClass, "my-point");
+            });
+
+            Assert.Contains("rz-timeline-point-base my-point", component.Markup);
+        }
+
+        [Fact]
+        public void TimelineItem_Renders_Current_AsAriaCurrentStep()
+        {
+            using var ctx = new TestContext();
+            var component = ctx.RenderComponent<RadzenTimelineItem>(parameters =>
+            {
+                parameters.Add(p => p.Current, true);
+            });
+
+            Assert.Contains(@"aria-current=""step""", component.Markup);
+        }
+
+        [Fact]
+        public void TimelineItem_Omits_AriaCurrent_WhenNotCurrent()
+        {
+            using var ctx = new TestContext();
+            var component = ctx.RenderComponent<RadzenTimelineItem>();
+
+            Assert.DoesNotContain("aria-current", component.Markup);
+        }
+
+        [Fact]
+        public void TimelineItem_Renders_ListItemRole()
+        {
+            using var ctx = new TestContext();
+            var component = ctx.RenderComponent<RadzenTimelineItem>();
+
+            Assert.Contains(@"role=""listitem""", component.Markup);
+        }
+
+        [Fact]
+        public void TimelineItem_Renders_LineStyle()
+        {
+            using var ctx = new TestContext();
+            var component = ctx.RenderComponent<RadzenTimelineItem>(parameters =>
+            {
+                parameters.Add(p => p.LineStyle, PointStyle.Success);
+            });
+
+            Assert.Contains("rz-timeline-line-success", component.Markup);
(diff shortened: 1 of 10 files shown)


enchev:
Thanks for the pull requst as always!

There is a problem with Styling demo (TimelineStyling.razor) - it uses a 36px line as a track behind small points. Master draws it continuous. The PR chops it into segments with gaps around every point, because the line halves now stop at the point's edge instead of running under it. Any user with a thick line or translucent points gets a visibly different timeline.

@yordanov might add more.

pianomanjh:
> Thanks for the pull requst as always!
> 
> There is a problem with Styling demo (TimelineStyling.razor) - it uses a 36px line as a track behind small points. Master draws it continuous. The PR chops it into segments with gaps around every point, because the line halves now stop at the point's edge instead of running under it. Any user with a thick line or translucent points gets a visibly different timeline.
> 
> @yordanov might add more.

Apologies, I did have a change queued to keep this backward compat that I did not push.  Now in place.

pianomanjh:
> Thanks for the pull requst as always!
> 
> There is a problem with Styling demo (TimelineStyling.razor) - it uses a 36px line as a track behind small points. Master draws it continuous. The PR chops it into segments with gaps around every point, because the line halves now stop at the point's edge instead of running under it. Any user with a thick line or translucent points gets a visibly different timeline.
> 
> @yordanov might add more.

Let me know if you'd like anything further on this, I believe it to be good to go

yordanov:
Thanks for the update! This looks good to me overall.

There is just one issue in RTL mode: the timeline line positioning is broken because it uses the physical `left` / `right` properties. Could you switch those to the logical `inset-inline-start` / `inset-inline-end` properties instead? That should preserve the current LTR behavior while positioning the line correctly in RTL as well.

image


Other than that, the PR looks good to go.

pianomanjh:
> Thanks for the update! This looks good to me overall.
> 
> There is just one issue in RTL mode: the timeline line positioning is broken because it uses the physical `left` / `right` properties. Could you switch those to the logical `inset-inline-start` / `inset-inline-end` properties instead? That should preserve the current LTR behavior while positioning the line correctly in RTL as well.
> 
> image
> Other than that, the PR looks good to go.

Thanks for the feedback, this was pre-existing in master, but happy to patch it up here.  Incoming.

pianomanjh:
> Thanks for the update! This looks good to me overall.
> 
> There is just one issue in RTL mode: the timeline line positioning is broken because it uses the physical `left` / `right` properties. Could you switch those to the logical `inset-inline-start` / `inset-inline-end` properties instead? That should preserve the current LTR behavior while positioning the line correctly in RTL as well.
> 
> image
> Other than that, the PR looks good to go.

Fix added.  Though I noticed something else:

In RTL, `LinePosition.Left`/`Right` still put the line and the content on the same side, so the labels squish against the line.  The item's flex row follows the writing direction, so the point moves to the far side while a pinned line doesn't:

| RTL, `LinePosition.Left` | |
|---|---|
| point | x 681–745 (flipped right by `dir`) |
| line | pinned left, ≈ x 353 |
| label box | x 321–681, `text-align: left` |

image

Master has this also, so I left it, but let me know if you want me to touch that up also.

PR #2704, opened 2026-09-06
HtmlEditor: resize images by dragging handles on the selected image

### LINKED ISSUE #2703: HtmlEditor: resize images with the mouse
There is no way to resize an image in the HtmlEditor with the mouse. The only
option is typing the width and height into the insert-image dialog.

I have a working implementation for the closed #2192, including the two things
left unresolved there: handles ending up in the editor value after cut and
paste, and the insert-image dialog not picking up the new dimensions.

PR follows.

MaurycyBartczak:
Closes #2703.

image resizing

A selected image gets eight handles. The corners keep the aspect ratio and
release it while <kbd>Shift</kbd> is held, the side handles change one axis.
<kbd>Esc</kbd> cancels a drag in progress and <kbd>Ctrl</kbd>+<kbd>Z</kbd> undoes
a finished one. No new parameter, it is always on like the table column resize.

The handles sit outside the contenteditable element, which is what keeps them
out of the editor value and out of the HTML after a cut and paste - the problem
#2192 could not get past. Size goes into the width and height attributes, so the
insert-image dialog picks up what a drag produced, which was the other one. The
commit messages cover the details.

The premium themes ship as prebuilt CSS and need regenerating on your side for
the handles to be styled there. The SCSS-built themes are covered here.

The Disabled guard has bUnit tests. The drag itself is browser behaviour, so I
checked it with Playwright against the demo app, including both problems from
#2192. The demo page gained a short "Image resizing" section.

### DIFF
--- a/Radzen.Blazor.Tests/HtmlEditorTests.cs
+++ b/Radzen.Blazor.Tests/HtmlEditorTests.cs
@@ -78,6 +78,41 @@ namespace Radzen.Blazor.Tests
         }
 
         [Fact]
+        public void HtmlEditor_Disabled_Ignores_OnChange()
+        {
+            using var ctx = CreateContext();
+            var raised = 0;
+            var component = ctx.RenderComponent<RadzenHtmlEditor>(parameters =>
+            {
+                parameters.Add(p => p.Value, "<p>initial</p>");
+                parameters.Add(p => p.Disabled, true);
+                parameters.Add(p => p.Input, args => raised++);
+            });
+
+            component.Instance.OnChange("<p>edited</p>");
+
+            Assert.Equal(0, raised);
+        }
+
+        [Fact]
+        public void HtmlEditor_Enabled_Raises_Input_OnChange()
+        {
+            using var ctx = CreateContext();
+            var raised = 0;
+            var html = string.Empty;
+            var component = ctx.RenderComponent<RadzenHtmlEditor>(parameters =>
+            {
+                parameters.Add(p => p.Value, "<p>initial</p>");
+                parameters.Add(p => p.Input, args => { raised++; html = args; });
+            });
+
+            component.Instance.OnChange("<p>edited</p>");
+
+            Assert.Equal(1, raised);
+            Assert.Equal("<p>edited</p>", html);
+        }
+
+        [Fact]
         public void HtmlEditor_Renders_Disabled_Attribute()
         {
             using var ctx = CreateContext();
--- a/Radzen.Blazor/RadzenHtmlEditor.razor.cs
+++ b/Radzen.Blazor/RadzenHtmlEditor.razor.cs
@@ -619,6 +619,11 @@ namespace Radzen.Blazor
         [JSInvokable]
         public void OnChange(string html)
         {
+            if (Disabled)
+            {
+                return;
+            }
+
             if (Html != html)
             {
                 Html = html;
--- a/Radzen.Blazor/themes/_css-variables.scss
+++ b/Radzen.Blazor/themes/_css-variables.scss
@@ -495,6 +495,12 @@
   --rz-editor-content-background-color: #{$editor-content-background-color};
   --rz-editor-focus-outline: #{$editor-focus-outline};
   --rz-editor-focus-outline-offset: #{$editor-focus-outline-offset};
+  --rz-editor-image-handle-size: #{$editor-image-handle-size};
+  --rz-editor-image-handle-background-color: #{$editor-image-handle-background-color};
+  --rz-editor-image-handle-border: #{$editor-image-handle-border};
+  --rz-editor-image-handles-border: #{$editor-image-handles-border};
+  --rz-editor-image-size-background-color: #{$editor-image-size-background-color};
+  --rz-editor-image-size-color: #{$editor-image-size-color};
 
   /* Fieldset */
   --rz-fieldset-border: #{$fieldset-border};
--- a/Radzen.Blazor/themes/components/blazor/_editor.scss
+++ b/Radzen.Blazor/themes/components/blazor/_editor.scss
@@ -4,11 +4,18 @@ $editor-toolbar-background-color: var(--rz-base-background-color) !default;
 $editor-content-background-color: var(--rz-base-background-color) !default;
 $editor-focus-outline: var(--rz-outline-focus) !default;
 $editor-focus-outline-offset: calc(-1 * var(--rz-outline-width)) !default;
+$editor-image-handle-size: 0.5rem !default;
+$editor-image-handle-background-color: var(--rz-base-background-color) !default;
+$editor-image-handle-border: 1px solid var(--rz-primary) !default;
+$editor-image-handles-border: 1px solid var(--rz-primary) !default;
+$editor-image-size-background-color: var(--rz-primary) !default;
+$editor-image-size-color: var(--rz-on-primary) !default;
 
 .rz-html-editor {
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
+  position: relative;
   border-radius: var(--rz-editor-border-radius);
   border: var(--rz-editor-border);
   overflow: hidden;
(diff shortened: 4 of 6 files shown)


akorchev:
Thanks for the PR! A few things to address before merging:

**Bugs**
- The Escape handler (Radzen.Blazor.js:5438) is a document-level keydown with no focus check, so Esc pressed anywhere (dismissing a dropdown or dialog) reverts the last resize. Move the Escape branch into the existing element-level `ref.keydownListener` (:5126). Same for the window `resize` listener (:5439) - register it in createImageHandles/removeImageHandles like the mousemove/mouseup pair.
- The drag only writes the `width`/`height` attributes (:5092). Images pasted from Word/Google Docs/web pages carry inline `style="width:..;height:.."` which wins over attributes, so they don't resize visibly while OnChange still commits contradictory attributes. Record and clear `img.style.width/height` at drag start (restore on Esc).
- The overlay is appended to the outer `.rz-html-editor` without clipping (:4950), so an image scrolled under the toolbar paints handles over the toolbar buttons and intercepts their clicks. Also subtract `clientLeft/clientTop` - every handle is currently 1px off because of the editor border.
- The MutationObserver has no `subtree` (:5440), so replacing/deleting an image inside a `<p>`, typing above it, or Align Center leaves stale handles at the old rectangle. Cheapest fix: call `ref.positionImageHandles()` from `ref.inputListener`.
- A Disabled editor still gets handles and fires Change on mouseup (:5240). Table column resize has the same hole, so a single `if (Disabled) return;` at the top of `OnChange(string)` in RadzenHtmlEditor.razor.cs:620 covers both.

**Simplification**
- imageResizeStartListener/stopImageResize/documentImageResizeEndListener (:4981-5108) repeat the table column resize lifecycle (mousedownResizeListener/mouseupResizeListener, :5293-5309). Extract a shared begin/end drag helper and end the drag via `ref.inputListener()` so the existing dispose-safe wrapper is kept - the two bare `invokeMethodAsync('OnChange')` calls (:5058, :5108) throw if mouseup lands after dispose.
- imageHandlesScrollListener, imageHandlesResizeListener and the observer callback (:5111-5121) are identical wrappers around `ref.positionImageHandles` - pass it directly. The dispose block (:5452-5456) inlines stopImageResize - call it.
- Cache the size label instead of `querySelector` on every move (:4957, :4965). `|| ref.parentElement` in getEditorContainer is unreachable.
- imageHandlePositions, the four indexOf booleans and the three-way if (:5076-5094), plus the two set-or-remove blocks in restoreImageSize (:5021-5032), collapse into one direction map `{nw:[-1,-1], n:[0,-1], ...}`.
- positionImageHandles and updateImageSizeLabel each read both rects with style writes in between (:4913-4919, :4973-4978) - two forced layouts per mousemove. Read once per move and pass them along.
- The Esc-after-drag one-entry undo (:5047-5058, :5105, :5232) doesn't integrate with Ctrl+Z or the Undo button. Either drop it (keep only the mid-drag cancel) or commit the final size through `execCommand('insertHTML')` once at mouseup so the native undo stack handles it.

**Comments**
- Please remove the comments at :4974-4975, :5037 and :5047-5048. The first also points at the hard-coded 28px (:4979), which breaks when a theme changes the label font - measure `label.offsetHeight` instead.

MaurycyBartczak:
All fixed.

- Escape and the window resize listener are scoped to the drag. The one-entry
  undo after a drag is gone: the drag is committed with
  execCommand('insertHTML'), so Ctrl+Z and the Undo button revert a resize.
- Inline width/height is recorded and cleared for the drag, restored on Escape.
- The handles moved into a clip layer over the editable area, so they no longer
  paint over the toolbar or take its clicks, and the overlay is offset by
  clientLeft/clientTop.
- positionImageHandles runs from inputListener. I kept a MutationObserver but
  narrowed it to the hidden attribute, since switching to source view raises no
  input and the handles stayed on screen without it.
- A disabled editor gets no handles, and OnChange returns early when Disabled,
  with two bUnit tests.
- trackDrag is shared with the table column drag, handle direction comes from a
  map, restoreImageSize loops over the recorded attributes instead of the two
  set-or-remove blocks, geometry is read once per mousemove, and the label is
  cached and measures its own height.
- The comments you pointed at are gone.

akorchev:
Thanks, almost there. A few leftovers:

- **New comments.** The three I flagged are gone, but three new ones appeared in Radzen.Blazor.js:
  > `// Absolutely positioned children start at the padding box, past the border.`
  > `// Pasted content often carries an inline width/height, which would outrank the attributes a drag writes.`
  > `// Rewound first so the undo entry records the size from before the drag.`

  Please remove them as well.
- **Stale demo text.** HtmlEditorPage.razor still says:
  > Press **Esc** during a drag to cancel it, or right after one to restore the previous size.

  The post-drag undo is gone, so end the sentence at "cancel it" (or mention Ctrl+Z).
- **Column resize still bypasses `inputListener`.** `mouseupResizeListener` keeps:
  > `try { instance.invokeMethodAsync('OnChange', ref.innerHTML); } catch { }`

  A rejected promise is not caught by try/catch. End the drag via `ref.inputListener()` like the image path now effectively does.
- **Reselect after `insertHTML`.** `ref.querySelector('img[width="' + width + '"][height="' + height + '"]')` picks the first image with those dimensions, not the one just inserted. Tag the clone with a temporary `data-` attribute, select it, then drop the attribute.
- **PR description.** It still says a resize does not enter the undo stack and that Esc undoes a finished drag. Both are now wrong.

MaurycyBartczak:
Done:

- The three comments are removed.
- The demo text ends at cancelling the drag and mentions Ctrl+Z.
- The column resize ends through ref.inputListener().
- The reselect uses the image's index in querySelectorAll('img'), taken before
  the replacement. A temporary data- attribute would end up in the value:
  insertHTML raises input synchronously, so inputListener sends the HTML while
  the attribute is still on the image.
- The PR description is updated.

While checking this I found a few more problems and fixed them:

- A linked image was pulled out of its link by insertHTML. The command now
  replaces the outermost inline element around the image.
- Undo brought rz-state-selected back into the value. The class is dropped
  before the command.
- Escape during a drag also closed a dialog the editor sits in. It is now caught
  in the capture phase.
- Disabling the editor or replacing its value left the handles on screen.
- container.clientTop was read after a style write, forcing a second layout.

akorchev:
Thanks, all the feedback so far is addressed. I found three more things while going through the last round that users would run into right away, so one more pass before merging:

- **Handles stay after keyboard deselect.** Only `clickListener` drops `rz-state-selected` and the handles. Select an image, press an arrow key or type elsewhere, and the handle box stays on the image until the next click. Before this PR the class was invisible so this went unnoticed. Please remove the class and handles from `selectionChangeListener` when the selection no longer contains the image.
- **Flash on mousedown for pasted images.** `imageResizeStartListener` clears `img.style.width`/`height` but does not write the `width`/`height` attributes until the first mousemove. An image sized only by `style="width:300px"` therefore snaps to its natural size on mousedown and jumps back once the mouse moves. When you clear the inline style, set the `width` and `height` attributes to the measured `offsetWidth`/`offsetHeight` in the same step so the image keeps its rendered size.
- **Empty `style=""` leaks into the value.** Clearing `width`/`height` on the clone leaves an empty `style` attribute when those were the only inline properties. Remove the attribute when the style list is empty.

MaurycyBartczak:
Done:

- selectionChangeListener drops the class and the handles when the selection no
  longer contains the image. It stays inside the existing activeElement check, so
  moving the focus out of the editor keeps the image selected, which the image
  dialog relies on: selectionAttributes('img') reads img.rz-state-selected even
  when the selection is outside the editor.
- Mousedown sets the width and height attributes to offsetWidth/offsetHeight when
  it clears the inline style. The no-op check at mouseup now compares against
  that size, otherwise clicking a handle without moving would commit and add an
  undo entry.
- The empty style attribute is removed.

Also fixed:

- Redo after a resize put rz-state-selected back into the value: undo detaches
  the reselected image with its class and redo restores the same node. The value
  from inputListener and queryCommands now leaves the class out, which also
  covers undo after deleting a selected image and toolbar commands on a selected
  image.
- clickListener left an empty class attribute on the image it deselected. It
  now goes through deselectImage, like selectionChangeListener and the mouseup
  commit.

PR #2706, opened 2026-09-08
Spreadsheet: two reader round-trip fixes, 47% less allocation when filling, and a bulk CellStore entrance

pianomanjh:
Five independent changes to `Radzen.Documents.Spreadsheet`, found while writing a data-grid export over
it. Each commit stands alone and can be taken or dropped separately.

### 1. Read a boolean cell back as a boolean

`XlsxWriter` writes `t="b"` correctly. `XlsxReader` ignored the attribute — every type but a shared
string fell through to `SetValueInvariant`, which re-infers from the text — so `<v>1</v>` came back as
the number 1. A boolean written by the library did not survive being read by it.

ECMA-376 types `<v>` as `ST_Xstring`, so a producer may write the word rather than the digit. The branch
accepts `1` and anything `bool.TryParse` takes, which is what the old path accepted before it.

### 2. Keep a quote-prefixed cell as text when reading

`quotePrefix` on a cell's format marks a value entered as literal text (ECMA-376 §18.8.45). `XlsxReader`
parsed the attribute into its style table, then set the value through the type-inferring path anyway —
so `007` came back as 7 and `1E3` as 1000. Same shape as (1): the file was correct and the reader
re-typed it.

Only a text cell is read back as text. The flag lives on the cell format and formats are shared by key,
so a number or a date can carry it without ever having been text; without that guard those load as
strings and a `SUM` over the range silently drops them.

The style is resolved once at the top of `ParseCell` and handed to `ApplyCellStyle`, which used to look
it up again for the same cell.

### 3. Do not walk the dependency graph for a cell nothing depends on

`OnCellValueChanged` runs on every write to `Cell.Value` and asks which cells depend on it.
`GetTopologicallySortedDependencies` allocates a `HashSet`, a `List` and a `Stack` before it can answer,
so a sheet with no formulas pays three allocations per cell for an empty result. `HasDependents` answers
off the dictionary already kept.

No API change.

### 4. Share one `CellData` for every empty cell

`Cell` initialised `Data` with `new CellData(null)`, so every cell allocated an object representing
nothing — and discarded it on first write. `CellData` has no setters, so one instance serves them all.

Adds `CellData.Empty`.

### 5. `CellStore.SetValues`, a bulk entrance

Filling a cell at a time grows the cell dictionary from empty, rehashing its way up to the final size.
`SetValues` knows how big the block is before it starts and sizes the dictionary once. It also does the
bounds check once for the block rather than once per cell.

**No update batch.** Each write notifies its own dependents, exactly as an assignment through the
indexer does — see below. A value replaces whatever the cell held, a formula included.

Adds `SetValues`. This is the only commit that adds meaningful public API.

## Measurements

50,000 rows × 11 columns of mixed strings, numbers, dates and booleans — 550,000 cells — three
interleaved passes, allocation via `GC.GetTotalAllocatedBytes(precise: true)`. **The values are
constructed before the measured region**, so the figures are the sheet's cost rather than the test
data's.

| | a cell at a time | via `SetValues` |
| --- | --- | --- |
| upstream | 197.6 MB | — |
| + no dependency walk (3) | 130.5 MB | — |
| + shared empty `CellData` (4) | 113.7 MB | — |
| + `SetValues` (5) | 113.7 MB | **94.0 MB** |

**52% less allocation**, and the drop to 113.7 MB arrives without any caller changing a line.

`SetValues`' share of that is **entirely the dictionary pre-sizing**: commenting out its
`EnsureCapacity` puts the arm back to 113.7 MB exactly, byte for byte with the indexer loop. The bounds
check it also saves per cell costs no allocation, and does not show above the noise in time.

Wall-clock is not quoted as a result anywhere here. On this machine the same benchmark arm has come back
at 1.36×, 1.40× and 1.79× of its baseline across three runs of one unchanged binary; `SetValues` itself
has measured both faster and slower than the indexer loop depending on the shape of the harness, while
its allocation was identical to the tenth of a megabyte every time.

*(An earlier version of this description gave 217 → 150 → 133 → 114 MB. Those came from a throwaway
harness that no longer existed; the rebuilt one reports about 20% lower at every rung because it builds
the test values outside the measured region rather than inside it. The code is the same either way — the
old numbers were simply not reproducible, so they have been replaced by ones that are.)*

### Why `SetValues` does not batch

`Worksheet.Batch` suspends formula evaluation, and `EndUpdate` then walks **every formula cell on the
sheet** with no dirty tracking. A 10×10 block written to a sheet carrying 5,000 unrelated formulas:

| | allocated |
| --- | --- |
| `SetValues`, no batch | **0.02 MB** |
| the same inside a batch | **1.94 MB** |

A hundredfold for a hundred cells, all of it re-evaluating formulas that never read the block. On a
sheet with no formulas the two are identical to the byte — 94.0 MB either way over 550,000 cells — which
is why no test can see the difference.

`Worksheet.Batch` itself is untouched and still used by `Sort`, `DeleteRow` and `InsertRow`; what
changed is that `SetValues` does not call it.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/BooleanRoundTripTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/BooleanRoundTripTests.cs
@@ -1,0 +1,129 @@
+using System.IO;
+using System.Linq;
+using System.IO.Compression;
+using System.Xml.Linq;
+using Radzen.Documents.Spreadsheet;
+using Xunit;
+
+namespace Radzen.Blazor.Spreadsheet.Tests;
+
+#nullable enable
+
+public class BooleanRoundTripTests
+{
+    private static Worksheet RoundTrip(Workbook workbook)
+    {
+        using var stream = new MemoryStream();
+
+        workbook.SaveToStream(stream);
+        stream.Position = 0;
+
+        return Workbook.LoadFromStream(stream).Sheets[0];
+    }
+
+    [Fact]
+    public void BooleanSurvivesTheRoundTrip()
+    {
+        var workbook = new Workbook();
+        var sheet = workbook.AddSheet("Sheet1", 3, 1);
+
+        sheet.Cells[0, 0].Value = true;
+        sheet.Cells[1, 0].Value = false;
+
+        Assert.Equal(CellDataType.Boolean, sheet.Cells[0, 0].ValueType);
+
+        var loaded = RoundTrip(workbook);
+
+        Assert.Equal(CellDataType.Boolean, loaded.Cells[0, 0].ValueType);
+        Assert.Equal(true, loaded.Cells[0, 0].Value);
+
+        Assert.Equal(CellDataType.Boolean, loaded.Cells[1, 0].ValueType);
+        Assert.Equal(false, loaded.Cells[1, 0].Value);
+    }
+
+    // ECMA-376 types <v> as ST_Xstring, so a producer may write the word. This writer writes 1 and 0,
+    // which is why the file has to be edited to produce the case at all.
+    [Theory]
+    [InlineData("true", true)]
+    [InlineData("TRUE", true)]
+    [InlineData("True", true)]
+    [InlineData("false", false)]
+    [InlineData("FALSE", false)]
+    [InlineData("0", false)]
+    public void ABooleanWrittenAsAWordIsRead(string written, bool expected)
+    {
+        var workbook = new Workbook();
+        var sheet = workbook.AddSheet("Sheet1", 2, 1);
+
+        sheet.Cells[0, 0].Value = true;
+
+        using var saved = new MemoryStream();
+
+        workbook.SaveToStream(saved);
+        saved.Position = 0;
+
+        using var edited = WithCellValue(saved, written);
+
+        var loaded = Workbook.LoadFromStream(edited).Sheets[0];
+
+        Assert.Equal(CellDataType.Boolean, loaded.Cells[0, 0].ValueType);
+        Assert.Equal(expected, loaded.Cells[0, 0].Value);
+    }
+
+    // Numbers keep the branch they always took: the boolean is chosen by t="b" and not by what the
+    // text happens to say, so a cell holding 1 is still the number 1.
+    [Fact]
+    public void NumbersAreUnaffected()
+    {
+        var workbook = new Workbook();
+        var sheet = workbook.AddSheet("Sheet1", 3, 1);
+
+        sheet.Cells[0, 0].Value = 42;
+        sheet.Cells[1, 0].Value = 1;
+        sheet.Cells[2, 0].Value = 0;
+
+        var loaded = RoundTrip(workbook);
+
+        Assert.Equal(CellDataType.Number, loaded.Cells[0, 0].ValueType);
+        Assert.Equal(42d, loaded.Cells[0, 0].Value);
+
+        Assert.Equal(CellDataType.Number, loaded.Cells[1, 0].ValueType);
+        Assert.Equal(1d, loaded.Cells[1, 0].Value);
+
+        Assert.Equal(CellDataType.Number, loaded.Cells[2, 0].ValueType);
+        Assert.Equal(0d, loaded.Cells[2, 0].Value);
+    }
+
+    /// <summary>Rewrites the single cell's &lt;v&gt;, so the file reads as another producer's.</summary>
+    private static MemoryStream WithCellValue(Stream source, string value)
+    {
+        var result = new MemoryStream();
+        source.CopyTo(result);
+        result.Position = 0;
+
+        using (var archive = new ZipArchive(result, ZipArchiveMode.Update, leaveOpen: true))
+        {
+            var entry = archive.GetEntry("xl/worksheets/sheet1.xml")!;
+            XDocument doc;
+            using (var stream = entry.Open())
+            {
+                doc = XDocument.Load(stream);
+            }
+
+            var ns = doc.Root!.Name.Namespace;
+            var cell = doc.Descendants(ns + "c").First();
+
+            Assert.Equal("b", cell.Attribute("t")!.Value);
+
+            cell.Element(ns + "v")!.Value = value;
+
+            entry.Delete();
+            var newEntry = archive.CreateEntry("xl/worksheets/sheet1.xml");
+            using var output = newEntry.Open();
+            doc.Save(output);
+        }
+
+        result.Position = 0;
+        return result;
+    }
+}
--- a/Radzen.Blazor.Tests/Spreadsheet/BooleanRoundTripTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/BooleanRoundTripTests.cs
@@ -41,8 +41,6 @@ public class BooleanRoundTripTests
         Assert.Equal(false, loaded.Cells[1, 0].Value);
     }
 
-    // ECMA-376 types <v> as ST_Xstring, so a producer may write the word. This writer writes 1 and 0,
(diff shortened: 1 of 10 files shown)


akorchev:
Hi @pianomanjh,

Thanks for the PR. 

> found while writing a data-grid export over it

We plan to release that ourselves. Please do not spend any time on this task.

Commits 3 and 4 look good and can go in as they are. I have findings on 1, 2 and 5, and a question about 6.

### 1. Boolean read

The `t="b"` branch compares `<v>` to `"1"` only, so `<v>true</v>` / `<v>TRUE</v>` from other producers now loads as `false`, where the old path (`bool.TryParse` via `TryConvertFromString`) gave `true`. ECMA-376 types `<v>` as `ST_Xstring` with no lexical restriction, and the reader already accepts `is "1" or "true"` for booleans in three other places in `XlsxReader.cs`. Please use the same pattern here.

The `"1"`-as-text case in `BooleanRoundTripTests.NumbersAndStringsAreUnaffected` is vacuous: `Cell.Value` type-infers `"1"` to a number before saving, so the writer never emits a shared string and the row is identical to the `Value = 1` row. Either build a genuinely string-typed cell or drop the case.

### 2. Quote prefix

The new branch is not limited to string cells, so any cell whose xf carries `quotePrefix` is now loaded as text. Our own writer emits exactly that combination: `CellStyleKey` includes `QuotePrefix` regardless of `ValueType`, so `cell.Value = 5; cell.QuotePrefix = true;` save/load now comes back as the string `"5"` (before: number 5 with `QuotePrefix` set), and `DateTime(2020,3,4)` comes back as the string `"43894"`. `SUM` over such a range silently drops the values. Guard the branch with `cellType is "s" or "str" or "inlineStr"`.

Also, `HasQuotePrefix` re-parses the `s` attribute and re-looks-up the style that `ApplyCellStyle` resolves for the same cell a few lines later, and then sets `QuotePrefix` through the public `SetValue("'" + value)` apostrophe protocol so it ends up written twice. Resolve the style once at the top of `ParseCell`, pass it to `ApplyCellStyle`, and expose the apostrophe branch of `Cell.SetValue` as an internal `SetText(string)` next to the existing internal `SetValueInvariant`.

### 5. SetValues

- It assigns through `Cell.Value`, which does not clear `Formula`, and its own `EndUpdate` then re-evaluates every formula, so a value written over a formula cell is silently discarded: `Cells[0,0].Formula = "=1+1"; Cells.SetValues(0, 0, [["x"]]);` leaves `Value == 2` and the formula intact. `Cell.SetValue(string)` clears the formula and is what the existing bulk path in `Worksheet` uses. Either clear `Formula` or document that formula cells are preserved, and add a test with a formula inside the block.
- Wrapping the fill in `BeginUpdate`/`EndUpdate` makes `SetValues` recalculate every formula on the sheet, since `Worksheet.EndUpdate` walks all formula cells with no dirty tracking. A sheet with thousands of unrelated formulas pays a full topological sort plus evaluation for a 10x10 block, where the indexer loop with your new `HasDependents` guard pays 100 O(1) lookups. Skip the batch when the graph has no formula cells, or evaluate only the dependents of the written cells.
- The get-or-create against `data` duplicates the indexer getter, and the hand-rolled `BeginUpdate`/`try`/`finally`/`EndUpdate` duplicates the public `Worksheet.Batch(Action)` that `DeleteRow`/`InsertRow`/`Sort` already use. Please extract a private `GetOrAdd(row, column)` shared by both and use `Batch`.

### 6. downloadFile

The current function is a faithful copy of the pattern in the official Blazor guidance (https://learn.microsoft.com/en-us/aspnet/core/blazor/file-downloads), which also revokes the object URL immediately after the click and does not attach the anchor. Both things the commit fixes were real browser bugs, but as far as I can tell they were fixed years ago:

- Firefox needing the anchor in the document for `click()` was https://bugzilla.mozilla.org/show_bug.cgi?id=1218456, fixed in Firefox 70 (backported to ESR 68.3).
- Revoking immediately after the click cancelling the download was https://bugzilla.mozilla.org/show_bug.cgi?id=1282407 in Firefox (fixed in 50) and https://bugs.webkit.org/show_bug.cgi?id=212279 in WebKit (fixed May 2020, covers the iOS case in 211234). The spec's intent, discussed in the Firefox bug, is that URL parsing is synchronous and copies the blob reference at parse time, so a later revoke cannot cancel a navigation that already started.

The 0.1 ms gap between click and revoke is always going to be tiny, so on its own it does not show a lost download. Did you actually see a download fail, and if so in which browser and version? If there is a reproducible failure in a supported browser I am happy to take a deferred revoke, but I would prefer `setTimeout(() => URL.revokeObjectURL(url))` with no delay and no `appendChild`, since leaving the click's tick is the whole point and a 30 second timer pins a multi-megabyte export blob in memory for no reason.

### Comments

Please strip the narration, rationale and history comments from the diff (test class headers, "used to", "pre-fix", the `HasDependents first: the walk allocates...` note in `Worksheet.cs`, `The far corner is enough` in `CellStore.cs`, the two in `Radzen.Blazor.js`, and the sentences after the ECMA-376 citations in `XlsxReader.cs`). The spec citations themselves are fine. Also `CellDependencyGraph` is internal, so it should not get `///` docs.

pianomanjh:
Thanks. Force-pushed; the branch is five commits now and each one builds and passes the suite on its own (5,123 → 5,144).

Understood on the data-grid export.  We have built our own integration for this with ClosedXml in production for a couple years, so are experimenting transitioning to your API, thus the PR.

### 6. downloadFile — withdrawn

### 1. Boolean read

Widened, and you were right that it was a regression. One deviation from the literal pattern you named: `is "1" or "true"` is case-sensitive, so it would still lose the `TRUE` you listed. It reads `value is "1" || bool.TryParse(value, out var flag) && flag`, which is exactly what the old path accepted plus the digit. Happy to change it if you'd rather the three-way literal.

The vacuous case is gone. `Cell.Value = "1"` type-infers to a number before saving, as you say, so that row was the `Value = 1` row twice. In its place is a theory over `true`/`TRUE`/`True`/`false`/`FALSE`/`0` that edits `xl/worksheets/sheet1.xml` in the saved file to produce the values this writer never emits, and a `NumbersAreUnaffected` that pins the branch to `t="b"` rather than to what the text says.

### 2. Quote prefix

Guarded with `cellType is "s" or "str" or "inlineStr"`. `ATypedCellCarryingTheFlagKeepsItsType` covers your exact case — a number, a date and a boolean each written with `QuotePrefix = true` — and fails without the guard. Worth noting what the date does: it comes back as the serial `43894` as a **number**, not a date, because `QuotePrefix` is part of `CellStyleKey` so the cell does not share the format that makes a serial read as a date. That predates this PR and I have left it alone; the test asserts the number, since a number still sums and the string you found did not.

The style is resolved once at the top of `ParseCell` and handed to `ApplyCellStyle`, which no longer re-parses `s`. The tuple got a file-scoped `using CellXf = (...)` alias so the signature stays readable.  `Cell.SetValue`'s apostrophe branch is now `internal void SetText(string)` beside `SetValueInvariant`, and `SetValue` calls it, so the reader sets the text directly instead of composing a string for `SetValue` to take apart.

### 5. SetValues

- The batch is **gone**, not gated, each write notifies its own dependents exactly as an indexer assignment does. That also removes a behavioral difference between the two paths, which is what the method claims to be. `FormulasOverTheBlockAreEvaluatedAfterIt` still passes without it.
- `Formula` is cleared before the value, with a test that writes over a formula inside the block and one that checks a formula *outside* it re-evaluates against the written value.
- `GetOrAdd(row, column)` is private and shared with the indexer getter. It still routes the insert through the (virtual) indexer setter rather than touching `data` directly, so an override still sees it.

Comments cleaned up.

PR #2708, opened 2026-09-08
Reduce XlsxWriter allocation: 442 MB to 12 MB over 550,000 cells, zero-GC

pianomanjh:
Reduces what `Workbook.SaveToStream` allocates, such no GC is necessary. Standalone commits measured over a
50,000 x 11 sheet (550,000 cells):

| | allocated |
| --- | --- |
| `master` | 442 MB |
| read a cell's format without creating one | 337 |
| format a cell reference into a span | 270 |
| sort the cells once instead of a tree per row | 215 |
| write the cells to the stream instead of building them as a tree | 81 |
| write the shared string table from the table | 64 |
| write references and numbers to the stream as characters | 25.7 |
| rent the sort array, size the string table | **17.7** |
| *review:* drop the string table hint | 21.6 MB |
| *review:* keep the shared string table in rented arrays | **12.4 MB** |

Measured into a `MemoryStream`; with the sink outside the window it is 9.6 MB. Each commit message
carries its own mechanism and figure.

## Against ClosedXML 0.104.2

BenchmarkDotNet, `[MemoryDiagnoser]`, same sheet. Both save arms build a workbook per invocation.

| save only | Mean | Allocated | Gen0 | Gen1 | Gen2 |
| --- | --- | --- | --- | --- | --- |
| `master` | 1,094.2 ms | 442.01 MB | 53000 | 28000 | 1000 |
| this branch | **261.0 ms** | **12.35 MB** | **-** | **-** | **-** |
| ClosedXML, in the `master` run *(control)* | 522.4 ms | 278.70 MB | 21000 | 1000 | 1000 |

The two builds are separate runs, so the control is there to say whether they compare: its allocation is
identical and its time moved 4.7%, which puts the save at **4.0x faster** normalised against it and **36x**
less allocated. It causes no collection of any generation.

Collections in a whole export are the fill's, not the writer's:

| | Allocated | Gen0 | Gen1 | Gen2 |
| --- | --- | --- | --- | --- |
| fill only (`CellStore.SetValues`) | 93.97 MB | 10667 | 5667 | 1000 |
| fill and save | 111.64 MB | 10000 | 5000 | 1000 |

Whole export, `master` against this branch. One process cannot hold two builds, so these are two runs
with ClosedXML as a control:

| run | | Mean | Allocated | Gen0 | Gen1 | Gen2 |
| --- | --- | --- | --- | --- | --- | --- |
| on `master` | this package | 1,396.0 ms | 535.99 MB | 65000 | 33000 | 3000 |
| on `master` | ClosedXML *(control)* | 630.8 ms | 353.81 MB | 29000 | 6000 | 3000 |
| on this branch | this package | 556.2 ms | 106.32 MB | 10000 | 5000 | 1000 |

Against the control measured beside it, the
export goes from **2.21x** ClosedXML's time on `master` to **0.71x** on this branch.

## Produced file

`Sort the cells once` numbers the shared string table and `cellXfs` in document order rather than in the order `CellStore` enumerates. `Give a merge anchor its style where the merge is written` (review) extends that to a formatted merge anchor, whose style is now numbered where the merge appears instead of ahead of every cell. Every index still selects what it selected - tests pin both tables and the merge case, and ClosedXML reads the file back with the same values and styles. Two workbooks with the same contents now serialise the same way whatever order they were filled in.

The other commits change no byte of a saved workbook. Checked against a workbook carrying a drawing, a table, hyperlinks, merges, conditional formatting, data validations, an autofilter, sheet protection, frozen panes, shared formulas with a master and a follower, error cells, a text-result formula, a quote prefix, hidden and custom-sized rows and columns, escaped text with a surrogate pair, and a sheet with no cells: sixteen parts, byte for byte.

## Points to check

- `WriteRaw` does not escape. Only a cell reference and an invariant-culture number reach it; text goes
  through `WriteString`.
- The shared string table is sized by a sheet's string cells, not the distinct strings among them.
- `CellRef.ToString` and `ColumnRef.ToString(int)` are public and now format through a span. Output is
  unchanged, including the row at `int.MaxValue`.

5,169 tests green, and green at each commit on its own.

Figures are allocation rather than wall-clock, from interleaved A/B runs of two builds, with each
mechanism confirmed by removing the line. The benchmark project and harnesses are left out of this PR;
happy to open them separately.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/ColumnRefTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/ColumnRefTests.cs
@@ -1,0 +1,71 @@
+using System;
+using Radzen.Documents.Spreadsheet;
+using Xunit;
+
+namespace Radzen.Blazor.Spreadsheet.Tests;
+
+#nullable enable
+
+public class ColumnRefTests
+{
+    [Theory]
+    [InlineData(0, "A")]
+    [InlineData(25, "Z")]
+    [InlineData(26, "AA")]
+    [InlineData(27, "AB")]
+    [InlineData(51, "AZ")]
+    [InlineData(52, "BA")]
+    [InlineData(701, "ZZ")]
+    [InlineData(702, "AAA")]
+    [InlineData(16383, "XFD")]
+    public void ToString_WritesTheColumnInA1Notation(int column, string expected)
+    {
+        Assert.Equal(expected, ColumnRef.ToString(column));
+    }
+
+    [Fact]
+    public void ToString_RejectsANegativeColumn()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => ColumnRef.ToString(-1));
+    }
+
+    [Theory]
+    [InlineData(0, 0, "A1")]
+    [InlineData(1, 26, "AA2")]
+    [InlineData(99, 702, "AAA100")]
+    public void CellRef_ToString_WritesTheColumnThenTheRow(int row, int column, string expected)
+    {
+        Assert.Equal(expected, new CellRef(row, column).ToString());
+    }
+
+    [Theory]
+    [InlineData(1048575, 16383, "XFD1048576")]
+    [InlineData(int.MaxValue, 0, "A-2147483648")]
+    public void CellRef_ToString_WritesTheWidestReferences(int row, int column, string expected)
+    {
+        Assert.Equal(expected, new CellRef(row, column).ToString());
+    }
+
+    [Fact]
+    public void CellRef_ToString_WritesTheWidestReferenceOfAll()
+    {
+        var reference = new CellRef(int.MaxValue, 2147483646)
+        {
+            IsRowAbsolute = true,
+            IsColumnAbsolute = true,
+        };
+
+        var text = reference.ToString();
+
+        Assert.Equal("$FXSHRXW$-2147483648", text);
+        Assert.Equal(CellRef.MaxLength, text.Length);
+    }
+
+    [Fact]
+    public void CellRef_ToString_MarksAnAbsoluteReference()
+    {
+        var reference = new CellRef(1, 27) { IsRowAbsolute = true, IsColumnAbsolute = true };
+
+        Assert.Equal("$AB$2", reference.ToString());
+    }
+}
--- a/Radzen.Blazor.Tests/Spreadsheet/SharedStringTableTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/SharedStringTableTests.cs
@@ -1,0 +1,98 @@
+using System.Collections.Generic;
+using System.Globalization;
+using Radzen.Documents.Spreadsheet;
+using Xunit;
+
+namespace Radzen.Blazor.Spreadsheet.Tests;
+
+#nullable enable
+
+public class SharedStringTableTests
+{
+    [Fact]
+    public void GetOrAdd_NumbersStringsInTheOrderTheyArrive()
+    {
+        using var table = new SharedStringTable();
+
+        Assert.Equal(0, table.GetOrAdd("b"));
+        Assert.Equal(1, table.GetOrAdd("a"));
+        Assert.Equal(2, table.GetOrAdd(string.Empty));
+
+        Assert.Equal(new[] { "b", "a", string.Empty }, table.Strings.ToArray());
+        Assert.Equal(3, table.Count);
+    }
+
+    [Fact]
+    public void GetOrAdd_ReturnsTheIndexAnEqualStringAlreadyHas()
+    {
+        using var table = new SharedStringTable();
+
+        table.GetOrAdd("Yes");
+        table.GetOrAdd("No");
+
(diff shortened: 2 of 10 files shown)


akorchev:
Hi @pianomanjh,

Thank you for the PR!

A few things before merging:

**Bugs**
- `CreateMergePlaceholders` (XlsxWriter.cs:1712) now uses the bounds-checked indexer. A populated cell outside RowCount/ColumnCount covered by a merge range makes `SaveToStream` throw (fill row 10, merge A1:A10, set Rows.Count = 5). Use a non-bounds-checked lookup. Same for the anchor read at :1697, which currently inserts an empty cell into the store as a side effect of saving.
- `sharedStrings.EnsureCapacity(Count + interned)` (:1374) sizes by string cells, not distinct strings. A 200k-row Yes/No column pre-allocates ~6 MB for a 2-entry table, and it compounds per sheet. Drop or cap the hint.
- Placeholders are created before any cell is written (:1378), so formatted merge anchors get the lowest style ids regardless of position. Contradicts the "document order" claim and the new ordering test. Compute anchor styles lazily or document the order.

**Comments**
- ColumnRef.cs:66 and XlsxWriter.cs:1481 add narration comments. Please remove.

**Simplifications**
- `FormatOrNull` pushed `format?.X == true` into eight style helpers. Substitute a single static default `Format` at the entry point instead. Also `Format.IsDefault` is a LINQ closure that allocates on every formatted cell.
- `TryFormatNumber` (:1669): CellData forces every number to double, so four of five switch arms are dead. A single `ISpanFormattable.TryFormat` produces identical text on net8/net10.
- `WriteRows` (:1407-1428): both arrays are already sorted, so the duplicated min/max scans reduce to one end-of-run scan per source.
- Placeholder `OrderBy().ThenBy().ToList()` (:1722) can be an in-place `List.Sort`. Note overlapping merges rely on stable order today, so include an insertion index.

akorchev:
A few comments crept back in. Comments are only allowed as `///` docs on public API, spec citations, or named workarounds for external bugs. Please remove:

- SharedStringTable.cs:66 and :101-102
- SharedStringTableTests.cs:67
- XlsxWriterSheetOrderTests.cs:235-236
- the `///` docs on `SharedStringTable` and its members - the class is internal

Thanks!

pianomanjh:
> Hi @pianomanjh,
> 
> Thank you for the PR!
> 
> A few things before merging:
> 
> **Bugs**
> 
> * `CreateMergePlaceholders` (XlsxWriter.cs:1712) now uses the bounds-checked indexer. A populated cell outside RowCount/ColumnCount covered by a merge range makes `SaveToStream` throw (fill row 10, merge A1:A10, set Rows.Count = 5). Use a non-bounds-checked lookup. Same for the anchor read at :1697, which currently inserts an empty cell into the store as a side effect of saving.
> * `sharedStrings.EnsureCapacity(Count + interned)` (:1374) sizes by string cells, not distinct strings. A 200k-row Yes/No column pre-allocates ~6 MB for a 2-entry table, and it compounds per sheet. Drop or cap the hint.
> * Placeholders are created before any cell is written (:1378), so formatted merge anchors get the lowest style ids regardless of position. Contradicts the "document order" claim and the new ordering test. Compute anchor styles lazily or document the order.
> 
> **Comments**
> 
> * ColumnRef.cs:66 and XlsxWriter.cs:1481 add narration comments. Please remove.
> 
> **Simplifications**
> 
> * `FormatOrNull` pushed `format?.X == true` into eight style helpers. Substitute a single static default `Format` at the entry point instead. Also `Format.IsDefault` is a LINQ closure that allocates on every formatted cell.
> * `TryFormatNumber` (:1669): CellData forces every number to double, so four of five switch arms are dead. A single `ISpanFormattable.TryFormat` produces identical text on net8/net10.
> * `WriteRows` (:1407-1428): both arrays are already sorted, so the duplicated min/max scans reduce to one end-of-run scan per source.
> * Placeholder `OrderBy().ThenBy().ToList()` (:1722) can be an in-place `List.Sort`. Note overlapping merges rely on stable order today, so include an insertion index.

Thank you for the review!  I addressed each commit by commit.  Also coming around with another comment cleanup.
One thing changed beyond what you asked: dropping the string table hint moved the all-distinct fixture from 17.7 to 21.6 MB (the dictionary's doubling series), which felt like the wrong direction for the headline, so the last commit puts the table in rented arrays instead.  (I've used this approach in past, stolen largely from [Collections.Pooled](https://github.com/jtmueller/Collections.Pooled#pooleddictionarytkey-tvalue)) That reads 12.4 MB for the same fixture and 4.3 MB into `Stream.Null` regardless of how repetitive the strings are.  Hopefully that's ok to keep in.
The bench on that:
| Save only, into `Stream.Null` | with the hint | hint dropped | rented arrays |
|---|---|---|---|
| every string distinct | 9.6 MB | 13.5 MB | 4.3 MB |
| one distinct column, one of five statuses, one constant | 17.9 MB | 13.5 MB | 4.3 MB |
| every string the same | 19.3 MB | 4.3 MB | 4.3 MB |

Let me know if you need anything else!

akorchev (approved):

PR #2710, opened 2026-09-09
Spreadsheet: read a cell that stores its own string (t="inlineStr")

pianomanjh:
`XlsxReader` reads no text from a cell that stores its own string, and returns it as an empty cell.

## The problem

ECMA-376 lets a cell carry its text inline instead of pointing into the shared string table: `t="inlineStr"` with the text under `<is>` and no `<v>` element at all. `ParseCell` looks only for `<v>` and `<f>`, so such a cell falls out at the first guard and is never written to the sheet.

```xml
<c r="A1" t="inlineStr"><is><t>Hello</t></is></c>
```

Loading a sheet of those gives a sheet of blanks, with no error. Excel writes inline strings in several situations, and other writers use them in place of the shared string table entirely, so a file can lose all of its text on load.

## The fix

Read `<is>` when there is no `<v>`, taking the value as the concatenation of its `<t>` descendants so that rich text reads as its whole string rather than as its first run. Everything after that is the existing string path, including the `quotePrefix` handling, which already listed `inlineStr` among the types it applies to.

## Tests

`InlineStringXlsxTests`, four cases: a plain inline string, one with `xml:space="preserve"`, rich text split across runs, and a sheet mixing an inline string with a number and a boolean. All four fail on `master` and pass with the change.

The full suite passes: 5,196 tests.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/InlineStringXlsxTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/InlineStringXlsxTests.cs
@@ -48,6 +48,29 @@ public class InlineStringXlsxTests
         return ms;
     }
 
+    private static MemoryStream BuildXlsxWithSharedStrings(string sheetDataXml, string sharedStringsXml)
+    {
+        var ms = BuildXlsxWithSheetData(sheetDataXml);
+
+        using (var zip = new ZipArchive(ms, ZipArchiveMode.Update, leaveOpen: true))
+        {
+            zip.GetEntry("xl/sharedStrings.xml")?.Delete();
+
+            var sst = XElement.Parse(sharedStringsXml);
+            foreach (var element in sst.DescendantsAndSelf())
+            {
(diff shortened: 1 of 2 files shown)


akorchev:
Hi @pianomanjh,

Thank you for the PR! One small thing before merging:

`inlineElem.Descendants(sNs + "t")` also picks up phonetic runs (`<is><rPh><t>`, ECMA-376 18.4.6), so a cell with furigana would get the phonetic text appended to its value. Please take only the direct `<t>` and the `<r><t>` runs.

pianomanjh:
> Hi @pianomanjh,
> 
> Thank you for the PR! One small thing before merging:
> 
> `inlineElem.Descendants(sNs + "t")` also picks up phonetic runs (`<is><rPh><t>`, ECMA-376 18.4.6), so a cell with furigana would get the phonetic text appended to its value. Please take only the direct `<t>` and the `<r><t>` runs.

Interesting!  I see the same issue above so added another commit for it

PR #2711, opened 2026-09-09
Spreadsheet: reduce fill and save allocation (zero-GC)

pianomanjh:
Follows #2708. The written Xlsx is unchanged; allocations reduced.

A `Cell` carried ten fields, six of which most cells never use. It now holds its value and type inline and the rest behind one reference, allocated only when a cell needs one.

The store keeps a value per address rather than a cell. A bulk fill writes values and builds nothing; the first caller to ask for a cell at an address gets one and it is kept, so `Cells[r, c]` still returns the same instance every time.

`CellView` lets the writer read a cell it does not have to build.

## Measurement

550,000 cells, eleven columns of mixed strings, numbers, dates and booleans, values built outside the measured region.

| | allocated | B / cell | Gen0 | Gen1 | Gen2 |
| --- | --- | --- | --- | --- | --- |
| `SetValues` | 94.0 → **18.4 MB** | 179 → 35 | 10 → **0** | 5 → **0** | 1 → **0** |
| a cell at a time | 113.7 → **75.9 MB** | 217 → 145 | 10 → 6 | 5 → 3 | 1 → 1 |
| fill and save, into `Stream.Null` | 103.6 → **23.9 MB** | 198 → 46 | 10 → **0** | 5 → **0** | 1 → **0** |

**A bulk fill and the save it feeds cause no garbage collection of any generation.** An address costs what it did before when a cell is built for it, and a third of that when none is. The same sheet filled one row in ten over 500,000 rows reports the same figures, so a sparse sheet is not paying for a dense one.

Times are not quoted: allocation repeated to a tenth of a megabyte across every run and time did not.

## Behavior

`Cell.Data` is built when asked rather than held, so two reads of an unchanged cell are two instances. `CellData` compares by type and value, matching the hash it already had, and four tests cover it. `IsEqualTo` is untouched and remains the lenient comparison a sheet uses, where an empty string equals a blank.

`HyperlinkCommand_UndoRestoresExactCellData` asserted the identity of `Cell.Data`. It now asserts the value and the type it was written to protect, and still fails if undo rebuilds the value from its text.

A cell the indexer builds is assigned through the indexer, so a derived store overriding the setter still observes it, and a derived store takes a per-cell path through that setter for a bulk fill as well. `CellStore.data` is private rather than protected, and nothing in the library derives from `CellStore`.

## Tests

`CellDataTests`, `CellStoreRegressionTests`, `AutoFitExportTests` and additions to `SpreadsheetCultureTests` cover the value-versus-cell storage, the derived-store dispatch, the type and quote-prefix inference, and auto-fit over a shrunk row count. The full suite passes: 5,206 tests.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/AutoFitExportTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/AutoFitExportTests.cs
@@ -1,0 +1,22 @@
+using System.IO;
+using Radzen.Documents.Spreadsheet;
+using Xunit;
+
+namespace Radzen.Blazor.Spreadsheet.Tests;
+
+public class AutoFitExportTests
+{
+    [Fact]
+    public void Save_AutoFitColumnWithCellBelowShrunkRowCount_DoesNotThrow()
+    {
+        var workbook = new Workbook();
+        var sheet = workbook.AddSheet("Sheet1", 20, 4);
+
+        sheet.Cells["A10"].Value = "a wide piece of text";
+        sheet.Columns.SetAutoFit(0);
+
+        sheet.Rows.Count = 5;
+
+        workbook.SaveToStream(Stream.Null);
+    }
+}
--- a/Radzen.Blazor.Tests/Spreadsheet/CellDataTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/CellDataTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using Xunit;
 using Radzen.Documents.Spreadsheet;
 namespace Radzen.Blazor.Spreadsheet.Tests;
@@ -34,5 +35,55 @@ public class CellDataTests
         var a = new CellData(1.0);
         var b = new CellData("hello");
         Assert.False(a.IsEqualTo(b));
+    }
+
+    private static Worksheet CreateSheet()
+    {
+        var workbook = new Workbook();
+
+        return workbook.AddSheet("Sheet1", 8, 4);
+    }
+
+    [Fact]
+    public void Data_ReadTwiceFromOneCell_IsEqualAndHashesAlike()
+    {
+        var sheet = CreateSheet();
+        sheet.Cells["A1"].Value = 42.0;
+
+        var first = sheet.Cells["A1"].Data;
+        var second = sheet.Cells["A1"].Data;
+
+        Assert.NotSame(first, second);
+        Assert.True(first.Equals(second));
+        Assert.Equal(first, second);
+        Assert.Equal(first.GetHashCode(), second.GetHashCode());
+    }
+
+    [Fact]
+    public void Data_ReadTwiceFromOneCell_IsFoundInAHashSet()
+    {
+        var sheet = CreateSheet();
+        sheet.Cells["A1"].SetValueInvariant("0123");
+
+        var set = new HashSet<CellData> { sheet.Cells["A1"].Data };
+
+        Assert.Contains(sheet.Cells["A1"].Data, set);
+    }
+
+    [Fact]
+    public void Equals_SameValueAndType_IsTrue()
+    {
+        Assert.True(CellData.FromNumber(42).Equals(CellData.FromNumber(42)));
+        Assert.True(CellData.FromString("a").Equals(CellData.FromString("a")));
+        Assert.True(CellData.Empty.Equals(new CellData(null)));
+    }
+
+    [Fact]
+    public void Equals_DifferentValueOrType_IsFalse()
+    {
+        Assert.False(CellData.FromNumber(42).Equals(CellData.FromNumber(43)));
+        Assert.False(CellData.FromNumber(1).Equals(CellData.FromString("one")));
+        Assert.False(CellData.FromNumber(42).Equals(null));
+        Assert.False(CellData.FromNumber(42).Equals("not a cell value"));
     }
 }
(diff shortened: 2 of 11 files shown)


akorchev:
Thank you for the PR, but I don't think this is the right direction and I'm going to decline it in its current shape.

**Why**

This replaces a central piece of the model to optimize one scenario - a bulk fill of hundreds of thousands of cells - and the cost lands on everything else. Formulas, formats and change notification are core features of a spreadsheet, not extras, and hiding them behind a lazily created bag plus a two-state store makes every lookup path ask "value or cell?" and materialise on demand. That produced the `GetType()` dispatch for derived stores, the key copy in `GetPopulatedCells`, a parallel `CellView`, and a second type-inference table to keep in sync with `CellData.Infer`. It also changes public behaviour: `Cell.Data` returning a new instance on every read is unexpected, and `CellStore.data` going from protected to private is a source break for anyone deriving from `CellStore`.

The measurements at realistic sizes also don't support the trade. At 10k-50k rows the per-cell fill (how most users fill a sheet) gains about a third, while reading the data back pays much of it away: the first read of a bulk-filled cell allocates a `Cell`, every `.Data` read allocates a `CellData`, and `=SUM` over a bulk-filled column allocates ~30% more than master.

**Concrete problems found**

- `Adopt` uses `GetValueRefOrAddDefault`, so enumerating `GetPopulatedCells` across `DeleteRow` or `Compact` resurrects removed keys as empty cells (CellStore.cs:362).
- Saving mutates the store: `ComputeAutoFitWidths` reads through `sheet.Cells[address]`, materialising every raw cell in auto-fit columns and throwing an `ArgumentOutOfRangeException` per out-of-range row as control flow (XlsxWriter.cs:1207).
- `Find`/`TryGet` now do two probes and permanently materialise raw entries, so one formula, filter, table or Ctrl+End pass over a bulk-filled block discards the saving for the sheet's lifetime (CellStore.cs:168, :134).
- The new `CellData.Equals` is ordinal while `CompareTo` is case-insensitive, so `CompareTo == 0` no longer implies `Equals` (CellData.cs:617).
- The `SetValues` remarks still say every write notifies dependents through `Cell.Value`, which is false for the raw path (CellStore.cs:263).
- `Cell.Changed`'s `add` allocates the `Extras` bag, and the grid subscribes on every rendered cell, so rendered cells are larger than on master. `StoredValue`, `StoredType` and `CellView.FormatDisplayText` are unused.

**What we would take**

If you want to pursue the allocation reduction, a much smaller PR could keep `CellData` as the stored field (so `Data` identity and zero-allocation reads stay), and the `CellView` path for the writers. No two-state store, no `GetType()` dispatch, no public API changes. Slimming `Cell` might also be an option but I'd keep `Formula`, `Format` and `Changed` as plain fields; they aren't rare.

pianomanjh:
> Thank you for the PR, but I don't think this is the right direction and I'm going to decline it in its current shape.
> 
> **Why**
> 
> This replaces a central piece of the model to optimize one scenario - a bulk fill of hundreds of thousands of cells - and the cost lands on everything else. Formulas, formats and change notification are core features of a spreadsheet, not extras, and hiding them behind a lazily created bag plus a two-state store makes every lookup path ask "value or cell?" and materialise on demand. That produced the `GetType()` dispatch for derived stores, the key copy in `GetPopulatedCells`, a parallel `CellView`, and a second type-inference table to keep in sync with `CellData.Infer`. It also changes public behaviour: `Cell.Data` returning a new instance on every read is unexpected, and `CellStore.data` going from protected to private is a source break for anyone deriving from `CellStore`.
> 
> The measurements at realistic sizes also don't support the trade. At 10k-50k rows the per-cell fill (how most users fill a sheet) gains about a third, while reading the data back pays much of it away: the first read of a bulk-filled cell allocates a `Cell`, every `.Data` read allocates a `CellData`, and `=SUM` over a bulk-filled column allocates ~30% more than master.
> 
> **Concrete problems found**
> 
> * `Adopt` uses `GetValueRefOrAddDefault`, so enumerating `GetPopulatedCells` across `DeleteRow` or `Compact` resurrects removed keys as empty cells (CellStore.cs:362).
> * Saving mutates the store: `ComputeAutoFitWidths` reads through `sheet.Cells[address]`, materialising every raw cell in auto-fit columns and throwing an `ArgumentOutOfRangeException` per out-of-range row as control flow (XlsxWriter.cs:1207).
> * `Find`/`TryGet` now do two probes and permanently materialise raw entries, so one formula, filter, table or Ctrl+End pass over a bulk-filled block discards the saving for the sheet's lifetime (CellStore.cs:168, :134).
> * The new `CellData.Equals` is ordinal while `CompareTo` is case-insensitive, so `CompareTo == 0` no longer implies `Equals` (CellData.cs:617).
> * The `SetValues` remarks still say every write notifies dependents through `Cell.Value`, which is false for the raw path (CellStore.cs:263).
> * `Cell.Changed`'s `add` allocates the `Extras` bag, and the grid subscribes on every rendered cell, so rendered cells are larger than on master. `StoredValue`, `StoredType` and `CellView.FormatDisplayText` are unused.
> 
> **What we would take**
> 
> If you want to pursue the allocation reduction, a much smaller PR could keep `CellData` as the stored field (so `Data` identity and zero-allocation reads stay), and the `CellView` path for the writers. No two-state store, no `GetType()` dispatch, no public API changes. Slimming `Cell` might also be an option but I'd keep `Formula`, `Format` and `Changed` as plain fields; they aren't rare.

Understood, I can see your reasoning.  I'll get this one closed.  I took a brief look at the more minimal approach and don't think it will move the needle much.

Our application has several high density grids that need to dump their filtered contents to excel, thus the recent push to get this path optimized.  The good news is that I think it's still possible, with the prior save path optimization, to get a high-performance export without these changes, while still honoring the API that you've got here.

I'll post what we're looking for later today.  Thanks for the consideration.

PR #2712, opened 2026-09-09
Fix: RangeNavigator keeps stale scales when its series data changes

### LINKED ISSUE #2713: RadzenRangeNavigator keeps stale scales when its series data changes
**Describe the bug**

`RadzenRangeNavigator` keeps the scales it built from the data it first saw. When the bound series' `Data` changes, the axis, the tick labels and the handle labels continue to show the previous domain. Resizing the browser window corrects them.

`UpdateScales()` is reached from only two places: `OnAfterRenderAsync(firstRender: true)`, after the element is measured, and the `OnResize` JS callback when the measured size changes. The series own the domain — `TransformCategoryScale`/`TransformValueScale` derive it from their items — but changing their `Data` changes no parameter of the navigator itself, so nothing recomputes. A resize is the only other route to `UpdateScales()`, which is why it appears to fix things.

The value scale goes stale the same way when the data changes shape without changing its span.

**To Reproduce**

1. Put a `RadzenRangeNavigator` with a `RadzenRangeNavigatorLineSeries` on a page, bound to a field.
2. Render it and let it measure.
3. Replace the series data with a set covering a different range and re-render.
4. The navigator still draws the first range's axis and handle labels.
5. Resize the browser window — the navigator snaps to the correct range.

A failing unit test is included in the linked PR: rendering with values 10–20, then re-rendering with 100–300, leaves `ValueScale.Input.Start` at 10.

**Expected behavior**

The navigator rebuilds its scales when the series data changes, without needing a resize.

**Desktop**
 - OS: macOS 15
 - Browser: Chrome
 - Version: Radzen.Blazor 11.3.2, .NET 10

**Additional context**

Same area as #2693. A consumer can work around it by giving the navigator a `@key` that changes with the data, forcing a remount and a fresh measurement.

pianomanjh:
Fixes #2713.

Changing a RangeNavigator series' data leaves its axis, handle labels, and SVG path stale until resize. This also affects changes that preserve the scale domain: renaming A/B/C to X/Y/Z or changing values from 10/20/15 to 10/20/12.

The line series now detects a changed Data reference in SetParametersAsync and asks the navigator to refresh after its items have been updated. Refresh recomputes the scales and renders the navigator, which owns the series SVG. Repassing the same Data reference does not request another refresh. This replaces the post-render domain comparison and removes the duplicate SameDomain helper.

Validation: all 16 RangeNavigator tests pass, including changed bounds, renamed categories, changed values within the same bounds, and a same-reference render check.

### DIFF
--- a/Radzen.Blazor.Tests/RangeNavigatorTests.cs
+++ b/Radzen.Blazor.Tests/RangeNavigatorTests.cs
@@ -184,6 +184,36 @@ namespace Radzen.Blazor.Tests
             Assert.DoesNotContain("NaN", component.Markup);
         }
 
+        [Fact]
+        public void RangeNavigator_WithLineSeries_UpdatesScales_WhenSeriesDataChanges()
+        {
+            // The scales used to keep the domain of the data first seen until a resize (#2713).
+            using var ctx = CreateChartContext();
+
+            var component = RenderNavigatorWithLineSeries(ctx);
+
+            Assert.Equal(10, component.Instance.ValueScale.Input.Start);
+            Assert.Equal(20, component.Instance.ValueScale.Input.End);
+
+            component.SetParametersAndRender(parameters =>
+                parameters.AddChildContent<RadzenRangeNavigatorLineSeries<DataItem>>(series =>
+                {
+                    series.Add(p => p.Data, WiderData);
+                    series.Add(p => p.CategoryProperty, nameof(DataItem.Category));
+                    series.Add(p => p.ValueProperty, nameof(DataItem.Value));
+                }));
+
+            Assert.Equal(100, component.Instance.ValueScale.Input.Start);
+            Assert.Equal(300, component.Instance.ValueScale.Input.End);
+        }
+
+        private static DataItem[] WiderData => new[]
+        {
+            new DataItem { Category = "A", Value = 100 },
+            new DataItem { Category = "B", Value = 300 },
+            new DataItem { Category = "C", Value = 200 },
+        };
+
         private static IRenderedComponent<RadzenRangeNavigator> RenderNavigatorWithLineSeries(TestContext ctx)
         {
             return ctx.RenderComponent<RadzenRangeNavigator>(parameters =>
--- a/Radzen.Blazor.Tests/RangeNavigatorTests.cs
+++ b/Radzen.Blazor.Tests/RangeNavigatorTests.cs
@@ -207,6 +207,46 @@ namespace Radzen.Blazor.Tests
             Assert.Equal(300, component.Instance.ValueScale.Input.End);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void RangeNavigator_Rerenders_WhenDataChangesWithinTheSameDomain(bool renameCategories)
+        {
+            using var ctx = CreateChartContext();
+            var component = RenderNavigatorWithLineSeries(ctx);
+            component.SetParametersAndRender(p => p.Add(x => x.ShowAxis, true));
+            var data = new[]
+            {
+                new DataItem { Category = renameCategories ? "X" : "A", Value = 10 },
+                new DataItem { Category = renameCategories ? "Y" : "B", Value = 20 },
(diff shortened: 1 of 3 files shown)


enchev:
Thanks! I don't think the post-render domain comparison is enough on its own.

The navigator renders the series SVG inside its own render tree, so when the parent re-renders with new data the navigator builds its markup before the child series has received the new Data. The first pass still draws the old items with the old scales, and the second pass only happens when UpdateScales() reports that the domain moved. Any data change that keeps the same domain is left stale on screen. I reproduced two such cases with bUnit on your branch:

```
        [Fact]
        public void RangeNavigator_WithLineSeries_RerendersAxis_WhenCategoriesRenamed()
        {
            using var ctx = CreateChartContext();

            var component = ctx.RenderComponent<RadzenRangeNavigator>(parameters =>
            {
                parameters.Add(p => p.ShowAxis, true);
                parameters.AddChildContent<RadzenRangeNavigatorLineSeries<DataItem>>(series =>
                {
                    series.Add(p => p.Data, SampleData);
                    series.Add(p => p.CategoryProperty, nameof(DataItem.Category));
                    series.Add(p => p.ValueProperty, nameof(DataItem.Value));
                });
            });

            Assert.Contains(">A<", component.Markup);

            component.SetParametersAndRender(parameters =>
                parameters.AddChildContent<RadzenRangeNavigatorLineSeries<DataItem>>(series =>
                {
                    series.Add(p => p.Data, RenamedData);
                    series.Add(p => p.CategoryProperty, nameof(DataItem.Category));
                    series.Add(p => p.ValueProperty, nameof(DataItem.Value));
                }));

            Assert.DoesNotContain(">A<", component.Markup);
            Assert.Contains(">X<", component.Markup);
        }

        [Fact]
        public void RangeNavigator_WithLineSeries_RerendersPath_WhenValuesChangeWithinSpan()
        {
            using var ctx = CreateChartContext();

            var component = RenderNavigatorWithLineSeries(ctx);

            Assert.Contains("166.66666666666669 100 ", component.Markup);

            component.SetParametersAndRender(parameters =>
                parameters.AddChildContent<RadzenRangeNavigatorLineSeries<DataItem>>(series =>
                {
                    series.Add(p => p.Data, SameSpanData);
                    series.Add(p => p.CategoryProperty, nameof(DataItem.Category));
                    series.Add(p => p.ValueProperty, nameof(DataItem.Value));
                }));

            Assert.Equal(10, component.Instance.ValueScale.Input.Start);
            Assert.Equal(20, component.Instance.ValueScale.Input.End);
            Assert.Contains("166.66666666666669 160 ", component.Markup);
            Assert.DoesNotContain("166.66666666666669 100 ", component.Markup);
        }

        private static DataItem[] RenamedData => new[]
        {
            new DataItem { Category = "X", Value = 10 },
            new DataItem { Category = "Y", Value = 20 },
            new DataItem { Category = "Z", Value = 15 },
        };

        private static DataItem[] SameSpanData => new[]
        {
            new DataItem { Category = "A", Value = 10 },
            new DataItem { Category = "B", Value = 20 },
            new DataItem { Category = "C", Value = 12 },
        };
```

- Renaming the categories A/B/C to X/Y/Z with the same count and values. CategoryScale.Data holds X/Y/Z afterwards, but the rendered axis still shows A, B, C. OrdinalScale.Data isn't part of the comparison and the fitted Input is identical for any three categories.
- Changing values within the same span, 10/20/15 to 10/20/12. Min and max don't move, so no re-render happens and the old path stays drawn. That's the "changes shape without changing its span" case from the issue text.

Neither is a regression, master behaves the same, but the PR claims to fix them.

I'd suggest making the refresh series-driven instead, the way CartesianSeries does it for RadzenChart (CartesianSeries.cs, SetParametersAsync). Have RadzenRangeNavigatorLineSeries override SetParametersAsync, check parameters.DidParameterChange(nameof(Data), Data), and call a Navigator.Refresh() that runs UpdateScales() and StateHasChanged(). The reference check covers the spin concern: the navigator's own re-render re-passes the same Data reference, so the series won't ask for another refresh. The chart can get away with a conditional refresh because its series render themselves; the navigator can't, because its series render through it.

Also ScaleBase already has IsEqualTo, which SameDomain partly duplicates.

PR #2716, opened 2026-09-10
Spreadsheet: Stream to a sheet from a row source

pianomanjh:
This introduces a path for performant streaming to a sheet, for bulk export scenarios.  (i.e. large grids or tables)

`Workbook.SaveToStreamAsync` takes a `StreamedSheet` instead of a workbook: a sheet name, a column list, and an `IAsyncEnumerable<T>` of rows. The rows are read once, in order, and never retained, so what the write holds does not grow with them.

The XML goes into the archive synchronously and only the row source is awaited, so the streamed path calls the same `WriteCell`, `WriteTypedValue` and `WriteRowStart` per usual, and types values through the same `CellData` inference. A streamed file and a built one agree cell for cell, which is what the equivalence test asserts.

The shared string table and the table part are written after the sheet, so the table range gets the final row count for free. `cols` and `dimension` precede the rows and cannot:

- `ColumnWidthMode.Declared` writes what each column declares and reads nothing. `Sampled(n)` buffers the first n rows (200 by default), measures the auto-fit columns over them, replays the buffer and streams the rest — bounded by n, not by the data.
- `InlineStrings` writes each string into the cell that holds it, trading the shared string table for a larger file.
- `dimension` is exact when `RowCount` is set, and exact without being told when the sample held the whole source; otherwise it is omitted, which the schema allows.

## Scope

Additive and opt-in. `SaveToStream(workbook)` is untouched, and the existing path stays the one for formatted reports.

A streamed sheet writes column widths (declared or sampled auto-fit), frozen panes, one table, per-column and header formats, and shared or inline strings. At this time it does not do merged cells, conditional formatting, data validation, protection, hyperlinks, images, autofilter, formulas, per-cell formats, multiple sheets or row heights.

Most of those sit after `sheetData` in `CT_Worksheet`, where this already emits `tableParts`, so they are additions a caller could declare up front rather than things streaming rules out. None is needed for what this serves: a flat extract too large for the built path to hold.

## Limits

The XLSX format limits are enforced as they are reached, rather than left for whatever opens the file: at most 16,384 columns (refused before a row is read), 1,048,576 rows, and 32,767 characters in a cell. `RowCount` is a promise, and a source that yields a different number faults the write.

## Measurement

Tested w 11 columns written to `Stream.Null`.

| rows | streamed, string columns | streamed, mixed types | built, then saved |
| --- | ---: | ---: | ---: |
| 10,000 | 173 KB | 2.7 MB | 26.8 MB |
| 50,000 | 186 KB | 12.8 MB | 131.0 MB |
| 200,000 | 173 KB | 50.5 MB | 533.8 MB |

**Read the two streamed arms as different questions.** The string-column arm is the writer alone, and it is flat: 171–173 KB whatever the row count, causing no collection of any generation. The mixed-type arm is what a caller with `int`, `DateTime` and `bool` columns pays.  A column is a `Func<T, object?>`, so every non-string cell boxes. Those boxes die in gen0 unwritten, though at 200,000 rows they are the whole of the 50.5 MB.

Against the built path the streamed one is **10.6x** cheaper on mixed types and **3,000x** on strings. The built arm is the control: if a sweep does not move it, the instrument is not resolving anything.

As a note: the ordinary save is a bit cheaper: 550.6 MB → 533.8 MB over 200,000 rows by eleven columns, and a save with no auto-fit column now copies no cells at all.

## Demo

`spreadsheet-streamed-export` writes an XLSX straight from an EF query of Northwind order lines joined across five tables, handed to the writer through `AsAsyncEnumerable`.

**Drop that commit if you would rather this PR stayed on the writer.** It is the last one and touches only `RadzenBlazorDemos`.

### DIFF
--- a/Radzen.Blazor.Tests/Spreadsheet/AutoFitXlsxRoundTripTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/AutoFitXlsxRoundTripTests.cs
@@ -138,6 +138,51 @@ public class AutoFitXlsxRoundTripTests
         Assert.Equal(78, sheet.Columns[0]);
     }
 
+    // A Top 10 rule reads every address in its range through the indexer, which adds the cells it does
+    // not find and throws past the sheet's bounds. Measuring an auto fitted column asks that rule for a
+    // format, so the width pass has to survive both.
+    [Fact]
+    public void Write_AutoFitColumn_SurvivesATop10RuleOverUnpopulatedCells()
+    {
+        var wb = Build();
+        var sheet = wb.Sheets[0];
+
+        for (var row = 0; row < 4; row++)
+        {
+            sheet.Cells[row, 0].Value = row * 1.5;
+        }
+
+        sheet.Columns.SetAutoFit(0);
+        sheet.ConditionalFormats.Add(
+            new RangeRef(new CellRef(0, 0), new CellRef(9, 0)),
+            new Top10Rule { Count = 2, Format = new Format { Bold = true } });
+
+        using var ms = Save(wb);
+
+        Assert.Single(ReadSheetXml(ms).Descendants(Ns + "col"));
+    }
+
+    [Fact]
+    public void Write_AutoFitColumn_SurvivesATop10RulePastTheSheetBounds()
+    {
+        var wb = Build();
+        var sheet = wb.Sheets[0];
+
+        for (var row = 0; row < 4; row++)
+        {
+            sheet.Cells[row, 0].Value = row * 1.5;
+        }
+
+        sheet.Columns.SetAutoFit(0);
+        sheet.ConditionalFormats.Add(
+            new RangeRef(new CellRef(0, 0), new CellRef(40, 0)),
+            new Top10Rule { Count = 2, Format = new Format { Bold = true } });
+
+        using var ms = Save(wb);
+
+        Assert.Single(ReadSheetXml(ms).Descendants(Ns + "col"));
+    }
+
     [Fact]
     public void Write_AutoFitColumn_KeepsWiderModelWidth()
     {
--- a/Radzen.Blazor.Tests/Spreadsheet/StreamedSheetPipelineTests.cs
+++ b/Radzen.Blazor.Tests/Spreadsheet/StreamedSheetPipelineTests.cs
@@ -1,0 +1,159 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Radzen.Documents.Spreadsheet;
+using Xunit;
+
+namespace Radzen.Blazor.Spreadsheet.Tests;
+
+#nullable enable
+
+/// <summary>
+/// §66: that the rows are pulled one at a time and written as they arrive, rather than drained first.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Every allocation figure this feature quotes rests on the source and the writer alternating: a path
+/// that read the rows into anything before writing them would report the same bytes per row on a small
+/// sheet and fall over on a large one. Allocation cannot see the difference - a list that is filled and
+/// dropped inside one measured region looks like garbage either way - so this asserts the ordering
+/// itself, by watching what the source has yielded at the moment bytes reach the destination.
+/// </para>
+/// <para>
+/// It is not a claim that nothing anywhere buffers. The zip's deflate stream holds tens of kilobytes
+/// before it passes any on, which is why the bound below is generous rather than one row; a real
+/// provider's reader holds a packet of rows for the same kind of reason. What is asserted is that
+/// <em>this</em> code holds nothing: writing begins while the source is near its start, and the source
+/// is still being read when the last of the file is written.
+/// </para>
+/// </remarks>
+public class StreamedSheetPipelineTests
+{
+    private const int Rows = 20_000;
+
+    private sealed class Source
+    {
+        public int Yielded;
+
+        public async IAsyncEnumerable<int> Read()
+        {
+            for (var i = 0; i < Rows; i++)
+            {
+                Yielded++;
+
+                yield return i;
+            }
+
+            await Task.CompletedTask;
+        }
+    }
+
+    private sealed class Watcher(Source source) : Stream
+    {
+        public int FirstWriteAtRow = -1;
+        public int LastWriteAtRow = -1;
+        public long Length_;
+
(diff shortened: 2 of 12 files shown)


akorchev:
Hi @pianomanjh,

Thank you for the PR. The idea is good and the measurements make the case: building a `Cell` for every value is the cost, and a bulk export should not pay it.

Some context on where this sits for us. The v12 branch already has a DataGrid export (`RadzenDataGrid.Export.cs`), with `ExportToExcelAsync` and `ExportToCsvAsync`. It reads the grid's rows through an `IAsyncEnumerable<TItem>`, paging through `LoadData` when the grid is bound that way, and builds a `Workbook` with the column titles, number formats derived from each column's `FormatString`, and the grid's theme colors and borders per cell. That export is the main consumer of large XLSX writes in the library, and it is exactly the case that suffers from materializing every cell. So whatever we add for streaming has to be something that export can call from the workbook it already builds, rather than a separate API with its own column types that it cannot use.

That is why we would rather not take the PR in its current shape. It adds a second writer next to the existing one, with its own archive assembly, sheet assembly and row loop, plus a parallel set of public types (`StreamedSheet`, `StreamedColumn`, `ColumnWidthMode`). Every future writer feature would then be done twice or excluded, and the exclusion list in the description is already long. The grid export could not use it either: it needs per-cell formats for alternating rows and borders, and it needs strings kept as text, while the streamed column supports one format per column and parses strings.

The design we want keeps the existing writer as the only writer. The `Workbook` stays the frame: the caller builds the header row, column widths, frozen panes and a table as usual, and passes a row source to save. The writer emits the built cells first, appends rows from the source through the same cell-writing helpers, then closes the table range with the final count. Concretely:

- Keep your first two commits. Having `WriteTypedValue`, `CellTypeAttribute` and the style helpers take a value, type and format instead of a `Cell` is a good refactor on its own.
- `WriteSheetData` takes an optional row source and loops it after the last built row, with a style cached per column and type.
- The writer chain becomes async once, awaiting only the source. The sync `SaveToStream` calls the same code with no source.
- Tables are written after the sheet XML so the range can use the written row count. `dimension` is omitted when a source is present, which the schema allows, so there is no `RowCount` promise to fault on.
- Auto-fit needs no sampling machinery. A caller who wants data-driven widths builds the first couple of hundred rows as real cells and streams the rest, and the existing measurement works unchanged.
- Rows yield `CellData`, optionally with a `Format`, so text stays text and per-row formatting such as alternating backgrounds is possible. That is what lets the v12 grid export switch to it later without losing anything.

The public surface would be a single method:

```csharp
Task SaveToStreamAsync(Stream stream, IAsyncEnumerable<CellData?[]> rows, CancellationToken cancellationToken = default)
```

appending to the workbook's only sheet, possibly with a small generic convenience over `IAsyncEnumerable<T>` and a selector.

If you would like to rework the PR along these lines we would be glad to review it. If you would rather not, no problem at all, we will implement it ourselves and credit the idea and the refactor to you.

pianomanjh:
> Hi @pianomanjh,
> 
> Thank you for the PR. The idea is good and the measurements make the case: building a `Cell` for every value is the cost, and a bulk export should not pay it.
> 
> Some context on where this sits for us. The v12 branch already has a DataGrid export (`RadzenDataGrid.Export.cs`), with `ExportToExcelAsync` and `ExportToCsvAsync`. It reads the grid's rows through an `IAsyncEnumerable<TItem>`, paging through `LoadData` when the grid is bound that way, and builds a `Workbook` with the column titles, number formats derived from each column's `FormatString`, and the grid's theme colors and borders per cell. That export is the main consumer of large XLSX writes in the library, and it is exactly the case that suffers from materializing every cell. So whatever we add for streaming has to be something that export can call from the workbook it already builds, rather than a separate API with its own column types that it cannot use.
> 
> That is why we would rather not take the PR in its current shape. It adds a second writer next to the existing one, with its own archive assembly, sheet assembly and row loop, plus a parallel set of public types (`StreamedSheet`, `StreamedColumn`, `ColumnWidthMode`). Every future writer feature would then be done twice or excluded, and the exclusion list in the description is already long. The grid export could not use it either: it needs per-cell formats for alternating rows and borders, and it needs strings kept as text, while the streamed column supports one format per column and parses strings.
> 
> The design we want keeps the existing writer as the only writer. The `Workbook` stays the frame: the caller builds the header row, column widths, frozen panes and a table as usual, and passes a row source to save. The writer emits the built cells first, appends rows from the source through the same cell-writing helpers, then closes the table range with the final count. Concretely:
> 
> * Keep your first two commits. Having `WriteTypedValue`, `CellTypeAttribute` and the style helpers take a value, type and format instead of a `Cell` is a good refactor on its own.
> * `WriteSheetData` takes an optional row source and loops it after the last built row, with a style cached per column and type.
> * The writer chain becomes async once, awaiting only the source. The sync `SaveToStream` calls the same code with no source.
> * Tables are written after the sheet XML so the range can use the written row count. `dimension` is omitted when a source is present, which the schema allows, so there is no `RowCount` promise to fault on.
> * Auto-fit needs no sampling machinery. A caller who wants data-driven widths builds the first couple of hundred rows as real cells and streams the rest, and the existing measurement works unchanged.
> * Rows yield `CellData`, optionally with a `Format`, so text stays text and per-row formatting such as alternating backgrounds is possible. That is what lets the v12 grid export switch to it later without losing anything.
> 
> The public surface would be a single method:
> 
> ```cs
> Task SaveToStreamAsync(Stream stream, IAsyncEnumerable<CellData?[]> rows, CancellationToken cancellationToken = default)
> ```
> 
> appending to the workbook's only sheet, possibly with a small generic convenience over `IAsyncEnumerable<T>` and a selector.
> 
> If you would like to rework the PR along these lines we would be glad to review it. If you would rather not, no problem at all, we will implement it ourselves and credit the idea and the refactor to you.

Appreciate the context about the v12 goals, that helps frame the work.  I was trying to keep clear of impacting the existing Worksheet code, but fitting into the existing API does make sense if this is to be used more broadly.  You've provided a very clear prescription, so I'll see if I can work it up to match.

pianomanjh:
This PR superceded by #2723.

@akorchev I've implemented the design per your request.
In order to facilitate `CellData` and `Format` I did add a thin `CellContent` struct wrapper for that.
This also allowed for typed factories to avoid the boxing for even lower allocation, for your consideration.  (as an optional commit)
Finally, there's an optional commit that refactors the styling handling, as it was getting tough to keep those code paths in alignment.
Happy to hear any further feedback.

PR #2718, opened 2026-09-11
RangeNavigator: opt-in hover tooltip

### LINKED ISSUE #2717: RadzenRangeNavigator: show a tooltip with the hovered point's value

`RadzenRangeNavigator` draws a series but gives no way to read it. Hovering the track shows nothing, so to learn what a peak or a gap represents a user has to drag a handle onto it and read the handle label, which moves the selection they were trying to inspect. `RadzenChart` series answer the same question with a tooltip; the navigator's series have no equivalent.

A hover tooltip on the navigator, opt-in:

- `RadzenRangeNavigator.ShowTooltip` (default `false`), plus `TooltipFormatString` for the value.
- While the pointer is over the track and not dragging, show the nearest point's category and value in a label that reuses the handle-label styling (`rz-range-nav-label`), so no theme changes are needed.
- The series supply their points through a new default-implemented member on `IRangeNavigatorSeries`, so existing custom series keep compiling and simply show no tooltip.
- The points go to the client once per scale update, next to the existing `Radzen.updateRangeNavigatorLabels` call, so hovering costs no round trips.

- Native SVG `<title>` elements on the series: they never fire, because the selection window and shades sit above the SVG and take the pointer for dragging.
- Invoking .NET on every mouse move to look the point up: a round trip per pixel of movement, which is noticeable on Blazor Server.
- Putting a `RadzenChart` with tooltips beside the navigator: duplicates the series and the scales, and the two drift apart.

**Additional context**

Found while building a date-range activity timeline on the navigator. Happy to send the PR; opening this first because it adds public API.

pianomanjh:
Fixes #2717.

Adds an opt-in hover tooltip to `RadzenRangeNavigator`. Hovering the track highlights the nearest point on the line and shows its category and value, the way chart series do. That makes it possible to read a peak or gap without dragging a handle onto it, which would move the selection being inspected.

- `ShowTooltip` turns it on and `TooltipFormatString` formats the value, defaulting to `{0:N}`. Changes to the format or culture refresh the tooltip at runtime.
- Disabling the tooltip clears its points so hover events skip DOM lookup.
- It works on the chart-attached `RadzenChartRangeNavigator` too, and the range navigator demo now shows it.
- It reuses the chart's tooltip and active-point styles, so themes need no changes.
- Custom `IRangeNavigatorSeries` implementations keep compiling; they show no tooltip until they implement `GetDataPoints`.

With several series, the tooltip shows only the single nearest point and doesn't name its series.

Validation: all 26 RangeNavigator tests pass, including runtime format/culture changes and disabling/re-enabling tooltips. JavaScript syntax and the disabled-hover early exit were also checked.

### DIFF
--- a/Radzen.Blazor.Tests/RangeNavigatorTests.cs
+++ b/Radzen.Blazor.Tests/RangeNavigatorTests.cs
@@ -1,6 +1,11 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
 using Bunit;
+using Microsoft.AspNetCore.Components;
+using Radzen.Blazor.Rendering;
 using Xunit;
 using static Radzen.Blazor.Tests.ChartTestHelper;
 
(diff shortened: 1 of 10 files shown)


enchev:
Thanks! Some findings:

- Changing TooltipFormatString or Culture at runtime without also flipping ShowTooltip doesn't re-send the points until the next resize, because only ShowTooltip sets the change flag.
- The default value format is raw {0}, so unformatted doubles show full precision. Chart tooltips look nicer by default, so a {0:N}-style fallback might be worth considering.
- The mousemove handler runs a querySelector on every move even when the tooltip is off. Cheap, but it could early-exit on a flag.

PR #2723, opened 2026-09-14
Spreadsheet: Stream rows from a source to a workbook

pianomanjh:
Bulk exports currently build a `Cell` for every value before saving. This adds `Workbook.SaveToStreamAsync(Stream, IAsyncEnumerable<CellData?[]>, CancellationToken)` to append source rows to the workbook's only sheet. Callers build the header, widths, frozen panes and tables in the workbook, then supply the remaining rows as an asynchronous source.

- Built cells are written first, then source rows are read once through the existing writer. Only source enumeration is asynchronous; ordinary synchronous saves retain multi-sheet support.
- Tables ending at the frame's last row, without a totals row, extend to the final written row without mutating the workbook. Streamed saves omit `dimension`; autofit measures built cells.
- Shared strings are the default. The overload with `useInlineStrings: true` avoids retaining distinct text from appended rows. A source or destination that buffers data, such as a `MemoryStream`, still contributes to memory use.
- Text stays text, including numeric-looking strings, and both string encodings preserve surrounding whitespace. Streamed values use default numeric and date formats; custom formats cannot be supplied for appended cells.
- Cancellation and failures release save resources and attempt to restore seekable output to its starting position.

A Northwind EF demo exports order lines using `CellData[]` rows.

Validation: all 2,046 spreadsheet tests pass. Release library builds pass for .NET 8, 9 and 10 with no warnings or errors. The Razor export demo compiles against the local library with its real EF context and models in an isolated harness. Tests cover both string modes, whitespace, asynchronous enumeration, cancellation, output failures, destination restoration, non-seekable output, tables and multi-sheet synchronous saves.


akorchev:
Thanks for the rework. This is the design we asked for: one writer, the workbook as the frame, `CellData?[]` rows appended to the only sheet, tables written after the sheet, `dimension` omitted, auto-fit over built cells, text kept as text. 

A few changes before we merge.

**Trim to the core.** We would take the first five commits, up to "Append rows from a source to a workbook's only sheet as it saves". Please drop the rest:

- **`CellContent`.** We asked for per-cell formats for the v12 grid export, but that export will not move to this path, so there is no consumer. Without one, `CellContent` and its factories are public surface we would have to maintain. Drop both commits and rewrite the demo on the `CellData?[]` overload; the frame already carries header formatting, widths and per-column number formats. If per-cell formats are wanted later, an overload over `(CellData? Data, Format? Format)[]` rows needs no new type.
- **The stylesheet move.** It is the largest commit and the feature does not depend on it. The streamed style cache can be a dictionary on the writer keyed by column and type. If you want the move, send it as its own PR.

**No sync over async.** `Write` now does `WriteAsync(stream, null, ...).GetAwaiter().GetResult()`, and `SaveSheets`, `SaveSheet`, `WriteSheetXml` and `WriteSheetData` became async to reach the single await on the source. It works today only because nothing awaits without a source. The first real await added to that path would deadlock `SaveToStream` under Blazor Server, and the sync path pays for state machines it never uses.

The writer does not need to be async. With one sheet, the streamed rows always land at the same place: after the frame's rows inside `sheetData`. Cut the existing sync writer there:

```csharp
SaveSession Begin(Stream stream);            // everything up to and including the frame's rows
void Append(SaveSession s, CellData?[] row); // one <row>
void End(SaveSession s);                     // the rest of the sheet, tables, styles, shared strings, package parts
```

```csharp
public void Write(Stream stream)
{
    var session = Begin(stream);
    End(session);
}

public async Task WriteAsync(Stream stream, IAsyncEnumerable<CellData?[]> rows, CancellationToken ct)
{
    var session = Begin(stream);
    await foreach (var row in rows.WithCancellation(ct))
    {
        Append(session, row);
    }
    End(session);
}
```

The sync path is the master code called in two halves, and the only await in the library is on the caller's source.

**Shared strings by default.** Streamed strings are always `inlineStr` while the frame uses the shared table, so one file mixes both, and nothing in the API shows it. We see the point: the table grows with distinct text. But Excel never writes inline strings, readers are less uniform about them, and for the common export where text repeats the file just gets bigger. Default to shared strings. If you want the flat-memory case, make inline strings an opt-in on the save.

**Small points.**

- `Discard` swallows every exception. Narrow it to `IOException` and `NotSupportedException`.
- If the style cache stays keyed by `Format` reference, document that a `Format` passed to the save is treated as immutable, since `Format` has no value equality. Keying by column and type avoids this.

With those, we are glad to merge.

pianomanjh:
Thanks, I’ve trimmed the optional additions, rewritten the demo on `CellData[]`, rebased onto master, and made the save-flow, shared-string and cleanup changes. 

On formatting: our application’s grid export currently uses ClosedXML to keep values numeric/date-typed while applying column formats such as four-decimal currency and explicit date patterns. We were hoping to migrate that export to this streaming writer.

In checking our migration path, I found that the current implementation doesn’t provide a way to apply column number formats to streamed values. The existing column API doesn’t expose them, and `CellData[]` carries values only. Default short dates work, but we’d need custom number/date formats to preserve our current exports.

Would you accept a follow-up adding formatting support for streamed rows? Your suggested `(CellData? Data, Format? Format)[]` overload could meet our needs without introducing a new public type, or we could follow a column-format approach if you’d prefer.

akorchev:
Merged, thank you. We removed the leftover `partial` modifier and a stray comment on master ourselves; nothing else needed changing.

On formats for streamed rows: yes, please send a follow-up. The `(CellData? Data, Format? Format)[]` overload is the shape we want. Build one `Format` per column and reuse it across rows, and the style cache should key on the format instance, type and quoting so a reused instance costs nothing per cell. We would rather not add a column-format API: columns carry no format in the model today and adding one would touch the built path as well.

